PR #8, opened 2015-04-01
[WIP] Add WebRTC support through OpenWebRTC

stefanalund:
Do not land, this is work in progress.


stefanalund:
Adding on top of develop instead.

PR #9, opened 2015-04-02
[WIP] Add WebRTC support through OpenWebRTC

stefanalund:
Do not land, this is work in progress.

Update: on top of develop instead of master.

### DIFF
--- a/samples/matrixConsole/Podfile
+++ b/samples/matrixConsole/Podfile
@@ -10,9 +10,9 @@ pod 'MatrixSDK', :path => '../../MatrixSDK.podspec'
 pod 'HPGrowingTextView', '~> 1.1'
 pod 'libPhoneNumber-iOS', '~> 0.7.6'
 pod 'GBDeviceInfo', '~> 2.2.9'
+pod 'OpenWebRTC-SDK', :path => '../../../openwebrtc-ios-sdk/OpenWebRTC-SDK.podspec'
 end
 
 target "matrixConsole" do
 
 end
-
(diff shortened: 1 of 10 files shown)


stefanalund:
@dbkr @ara4n @manuroe how do you want to proceed with this PR? From my end it is almost done and the basic calling works between iOS and your web app running in Chrome.

For this to be really useful these fixes comes to mind:
* Receive calls regardless of where you are in the app, not only for a specific room
* Add UI for the WebRTC enabled setting
* Ringtone
* Possibly fix ugly Call button in RoomViewController
* Polish the UI in CallViewController

I think some (all?) of the above bullets are better handled by you guys as you are more at home in the app (and possibly switching to MatrixKit soon?). 

Cheers

ara4n:
This sounds sensible to me - the main thing missing on your side (i think?) is an optimised cocoapod of OWR to link against.  We are all out until Tuesday for Easter, when I'm finally back from holiday and will check with the rest of the gang.

I suspect we will start by porting this over to the MatrixKit based build of Console; not sure it's worth the faff of maintaining against the pre-matrixkit codebase too and will give people a good reason to upgrade to MatrixKit :)

Stefan: thanks again for this...

stefanalund:
@ara4n no problem, I think this serves both our projects well! Once I had our SDK "done" it didn't take that long to integrate in to Console (which is probably a good testament for both of us :)).

You are correct about our CocoaPod, what I have been using now is only for experimentation. Rob is hoping to have our first (optimised) release out already end of next week though. What's the ETA on MatrixKit and an updated Console?

manuroe:
OpenWebRTC was integrated in Matrix Console a while ago.
FTR, it is enabled only on the dev branch.

PR #12, opened 2015-10-05
MXHTTPClient: add block to handle unrecognized certificate during aut…

giomfo:
…hentication challenge from a server.

### DIFF
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -16,7 +16,6 @@
 
 #import "MXRestClient.h"
 
-#import "MXHTTPClient.h"
 #import "MXJSONModel.h"
 #import "MXTools.h"
 
@@ -92,7 +91,7 @@ MXAuthAction;
     _currentPreferredAPIVersion = inPreferredAPIVersion;
 }
 
--(id)initWithHomeServer:(NSString *)inHomeserver
+-(id)initWithHomeServer:(NSString *)inHomeserver andOnUnrecognizedCertificateBlock:(MXHTTPClientOnUnrecognizedCertificate)onUnrecognizedCertBlock
 {
     self = [super init];
     if (self)
@@ -100,7 +99,9 @@ MXAuthAction;
         homeserver = inHomeserver;
         preferredAPIVersion = _currentPreferredAPIVersion;
         
-        httpClient = [[MXHTTPClient alloc] initWithBaseURL:[NSString stringWithFormat:@"%@%@", homeserver, kMXAPIPrefixPath] andAccessToken:nil];
+        httpClient = [[MXHTTPClient alloc] initWithBaseURL:[NSString stringWithFormat:@"%@%@", homeserver, kMXAPIPrefixPath]
+                                               accessToken:nil
+                         andOnUnrecognizedCertificateBlock:onUnrecognizedCertBlock];
         
         // By default, use the same address for the identity server
         self.identityServer = homeserver;
@@ -108,7 +109,7 @@ MXAuthAction;
     return self;
 }
 
--(id)initWithCredentials:(MXCredentials*)inCredentials
+-(id)initWithCredentials:(MXCredentials*)inCredentials andOnUnrecognizedCertificateBlock:(MXHTTPClientOnUnrecognizedCertificate)onUnrecognizedCertBlock
 {
     self = [super init];
     if (self)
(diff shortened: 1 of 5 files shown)


MatrixSDK/Utils/MXHTTPClient.m
```diff
@@ -81,6 +86,17 @@ -(id)initWithBaseURL:(NSString *)baseURL andAccessToken:(NSString *)access_token
         
         httpManager = [[AFHTTPRequestOperationManager alloc] initWithBaseURL:[NSURL URLWithString:baseURL]];
         
+        // Check whether some certificates have been included in app bundle
+        AFSecurityPolicy *securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeCertificate];
+        if (securityPolicy.pinnedCertificates.count)
```
dbkr: Looks like this is AFNetworking looking for certs in the app bundle - do we need this? Or does it look in the Documents directory too?

MatrixSDK/Utils/MXHTTPClient.m
```diff
@@ -81,6 +86,17 @@ -(id)initWithBaseURL:(NSString *)baseURL andAccessToken:(NSString *)access_token
...
+        {
+            securityPolicy.allowInvalidCertificates = YES;
```
dbkr: Is this option needed for AFNetworking to go & defer the cert validation to us via the callback? It seems like if we're doing the cert validation ourselves (which it looks like we are below) this should be redundant.

MatrixSDK/Utils/MXHTTPClient.m
giomfo: This code looks only for certificate in app bundle (no in Documents dir).

I keep it here to support the original AFNetworking pinned certificates handling. This is an additional option for the developer to handle certificates

MatrixSDK/Utils/MXHTTPClient.m
giomfo: If the certificate is included in app bundle, AFNetworking module will allow it without prompting the user. Is that your question?

It is not redundant, it is another way to handle certificate, if the developer uses its own certificates

MatrixSDK/Utils/MXHTTPClient.m
dbkr: Oh, I see - that's a nice idea. Might be worth commenting to that effect.

MatrixSDK/Utils/MXHTTPClient.m
dbkr: Aha, yes - this makes sense now.

MatrixSDK/Utils/MXHTTPClient.m
```diff
@@ -348,6 +364,87 @@ - (void)tryRequest:(MXHTTPOperation*)mxHTTPOperation
...
+            else
+            {
+                NSLog(@"[MXHTTPClient] Shall we trust %@?", protectionSpace.host);
+                
+                if (onUnrecognizedCertificateBlock)
+                {
+                    SecTrustRef trust = [protectionSpace serverTrust];
+                    CFIndex index = SecTrustGetCertificateCount(trust);
+                    
+                    for (int i = 0; i < index; i++)
```
dbkr: I'm not sure we want to be looping through the certificates here, as I think this would mean we'd keep popping up dialogs for the other certs in the chain if the user chose not to trust it (although self signed certs will only have one). Android prompts for the leaf certificate, so I'd suggest just doing that (the one at index 0: https://developer.apple.com/library/mac/documentation/Security/Reference/certifkeytrustservices/#//apple_ref/c/func/SecTrustGetCertificateAtIndex).

MatrixSDK/MXRestClient.m
```diff
@@ -318,6 +321,9 @@ - (MXHTTPOperation*)registerOrLoginWithUser:(MXAuthAction)authAction user:(NSStr
                              // Workaround: HS does not return the right URL. Use the one we used to make the request
```
dbkr: I know this isn't really part of this PR, but is this something we need to fix in synapse, and if so is there a bug for it?

giomfo: This is not actually a bug, according to the spec (https://matrix.org/docs/spec/#login) the key 'home_server' provides the hostname of the Home Server.

I think this key was previously used to return the home server url. So matrixSDK defined the field 'homeServer' to store the server url and not the hostname. This was the origin of this workaround.

I don't think synapse is able to rename this key ('home_server') in server response, so
we should refactor this field in matrixSDK. I just created SYIOS-155.

dbkr:
LGTM

PR #16, opened 2015-10-15
Roomstate optim

ylecollen:


manuroe:
The roomState code looks good to me.
I am not so sure for the unrelated commit about the read receipts but I guess we could ignore push rules on read receipts for now.

PR #21, opened 2015-10-23
Server Sync V2: Increase supported Server-Client version API: v2

giomfo:
Only read receipts are enabled.
Other C-S v2 implementation is commented (uncomplete).


ylecollen:
LGTM

PR #24, opened 2015-10-28
MXSession: Added an @autoreleasepool on live events management.

manuroe:
Reformatted all existing blocks to be conformed with the code writing.

The added autoreleasepool block reduces this memory usage spike (which happens when retrieving thousands of events on app resuming):
screen shot 2015-10-27 at 16 59 20


ylecollen:
LGTM

PR #29, opened 2015-10-30
Core data store

manuroe:
Still WIP.
Read receipts storage is not implemented and I am not happy with the performance.
The performance issues:
 - eventWithEventId is relatively slow (around 1ms on a 5s) for the intensive usage made by the SDK. The issue is indeed more the usage by the SDK. Most of time, the SDK calls this method to check that an event already exists to avoid duplicated events: it does not need to get the full event data. We need a dedicated method for that in MXStore.

- there is a background queue for storing data to core data but the [MXStore commit] is currently implemented in a blocking way because the sdk needs to read data it just wrote. In the current MXCoreDataStore implementation, the core data saving must be completed before reading back data.


ylecollen:
Some useless code trigger some warnings

like  NSDate *startDate = [NSDate date]; in - (void)storeEventForRoom:(NSString*)roomId event:(MXEvent*)event direction:(MXEventDirection)direction

ylecollen:
LGTM

PR #31, opened 2015-10-30
BugFix SYIOS-170: Public Room: room history is wrong when user joins …

giomfo:
…for the second time.

This issue is due to a race condition between live event stream handling and initial sync of the joined room.

The boolean 'isSync' was updated too early. The server response handling must be complete before updating this flag.


manuroe:
LGTM

PR #32, opened 2015-11-02
MXRestClient: optimize server response parsing.

giomfo:
We define a new processing queue to process server response.
This queue is used to create models from JSON dictionary without blocking the main thread.


giomfo:
Related to SYIOS-169: Improve MXEvent conversion

ylecollen:
LGTM

PR #33, opened 2015-11-03
BugFix SYIOS-169: Improve MXEvent conversion.

giomfo:
We add some improvements in response handling during server sync v2.


ylecollen:
LGTM

PR #35, opened 2015-11-10
Performance profiling

manuroe:
Did some refactoring according to Xcode profiler.
Tested when doing a big app resuming.


giomfo:
LGTM

PR #36, opened 2015-11-17
Room state optimisation

manuroe:
This PR contains 2 improvements on [MXRoomState copyWithZone:]:
- limit the number of calls of it
- avoid useless deep copies in its implementation


ylecollen:
LGTM

PR #40, opened 2015-11-27
Room tag support

manuroe:
Added room tag support and some helper methods in MXSession to get ordered list of tagged rooms


ylecollen:
LGTM

PR #42, opened 2015-12-02
MXEvent: Added support of invite_room_state

manuroe:
It is part of SYIOS-174: Invited room: Show metadata and support rejection.

A user can now get a subset of the room state when it is invited.
As MatrixKit is magic, thanks to this change in the SDK, the room name where the user is invited displays now the room name in the *recents* list.


ylecollen:
LGTM

PR #44, opened 2015-12-08
Add the replacePassword method

ylecollen:


MatrixSDK/MXRestClient.m
```diff
@@ -222,6 +222,54 @@ - (NSString*)loginFallback;
 }
 
 
+#pragma mark - password update operation
+
+/**
```
manuroe: Public methods definitions should be only in .h not in .m

MatrixSDK/MXRestClient.m
```diff
@@ -222,6 +222,54 @@ - (NSString*)loginFallback;
...
+    }
+    
+    NSDictionary *authparameters = @{
+                                     @"type": kMXLoginFlowTypePassword,
+                                     @"user": self.credentials.userId,
+                                     @"password": oldPassword,
+                                     };
+    
+    NSDictionary *parameters = @{
+                                 @"auth": authparameters,
```
manuroe: FTR, authparameters  is not required here. You can use literals inside literals - ie, @{} inside a @{} and so on.

PR #45, opened 2015-12-09
read_receipts_for_incoming_messages

ylecollen:
The client assumes that an event sender reads the messages until its own messages.
Read receipts are created by now for each received messages.


giomfo:
Cancelled

PR #47, opened 2015-12-10
MXSDKOption:  Configure identicon use at SDK level.

giomfo:
By default Matrix SDK sets an identicon url when user's avatar is undefined
 (see [MXRestClient urlOfIdenticon:] use).

We define here a new property named 'disableIdenticonUseForUserAvatar' to disable identicon use at SDK level.


ylecollen:
LGTM

PR #51, opened 2015-12-15
Invite room state: BF: Do not use the invite_room_state field to dete…

manuroe:
…rmine if the user is invited or not. This is wrong.

So, come back to the check of the user membership in the room state.


ylecollen:
LGTM

PR #52, opened 2015-12-15
MXEvent Add readReceiptEventIds property

giomfo:
This property is relevant only for events with 'kMXEventTypeStringReceipt' type.
Contains the event IDs for which a read receipt is defined in this event.

### DIFF
--- a/MatrixSDK/JSONModels/MXEvent.h
+++ b/MatrixSDK/JSONModels/MXEvent.h
@@ -169,18 +169,6 @@ typedef enum : NSUInteger
 @property (nonatomic) NSString *sender;
 
 /**
- This property is relevant only for events with 'kMXEventTypeStringReceipt' type.
- Contains the event IDs for which a read receipt is defined in this event.
- */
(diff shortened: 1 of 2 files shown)


MatrixSDK/JSONModels/MXEvent.h
```diff
@@ -169,6 +169,12 @@ typedef enum : NSUInteger
 @property (nonatomic) NSString *sender;
 
 /**
+ This property is relevant only for events with 'kMXEventTypeStringReceipt' type.
+ Contains the event IDs for which a read receipt is defined in this event.
+ */
+@property (nonatomic) NSArray *readReceiptEventIds;
```
manuroe: I would prefer this thing (and `receiptSenders` indeed) to be a method rather than a property.
The reason is that some automatic computations for JSON parsing are done by listing properties of MXJSONModel classes.
As this information does not come from a JSON object received by an homeserver, it should be not a property.

manuroe:
LGTM

PR #53, opened 2015-12-16
Room invite race conditions

ylecollen:
The room accountData  for the new room is only stored if there is a tag update.
So, after an application restarts, this object was nil in the MXRoom class.

After, any new tag management was ignored because of this nil pointer.

By now, the accountData class is created if the dedicated file does not exist..


manuroe:
LGTM

PR #54, opened 2015-12-17
room_invite_race_condition_3

ylecollen:
Improve the invite management 

-> when the user joins a room on his own device, the room is removed from invited list as soon as the join request succeeds.

-> to delete a room from the invited ones list, we also check by roomId.
Indeed, during the room initial sync, the room object is replaced by a new one so the search by object is corrupted.


manuroe:
LGTM

PR #55, opened 2015-12-22
MXPublicRoom spec update: Added worldReadable, guestCanJoin & avatarURL

manuroe:
Update with the last spec changes.


giomfo:
LGTM

PR #56, opened 2015-12-22
Update syncv2 implementation

giomfo:
SPEC 254: V2 sync: return state corresponding to the *start* of the timeline.


manuroe:
LGTMSF

PR #60, opened 2016-01-11
invalid_read_receipts

ylecollen:
-> generate a read receipt for the sender of an incoming message
1 - Some clients don't send read receipt so it is a way to have read receipt
2 - The read receipts can be sent with delay so it avoids having a read receipt older than a message sent by an user.


manuroe:
LGTM
Note: on a typing event, the sdk does not create an implicit read receipt but this is how vector-web works too: https://github.com/vector-im/vector-web/issues/394

PR #61, opened 2016-01-18
Server sync v2 : Resume presence handling

giomfo:
We remove here the specific request used during catch-up sync v1 to handle separately the presence.
The presence are handled by the server sync v2 request when the event stream is resumed.


manuroe:
LGTM

PR #63, opened 2016-01-21
MXSession: Bug Fix - Push rules refresh

giomfo:
Load push rules from server before loading store data in order to highlight the bing events.


manuroe:
lgtm

PR #65, opened 2016-01-25
Disable core data

manuroe:
Disabled the code because the performance is not better than MXFileStore and it is boring to maintain when there is a change in MXStore.h


giomfo:
LGTM

PR #69, opened 2016-01-26
Remove mantle

manuroe:
In the continuation of https://matrix.org/jira/browse/SYIOS-169, Mantle has been totally removed.
The pro is that we have the best performance. 
The con is that we lost nice automatic conversion methods (dictionaryValue, NSCoding, NSCopying). These methods must be now coded by "hand" for each MXJSON derived class.

### DIFF
--- a/MatrixSDK/JSONModels/MXEvent.m
+++ b/MatrixSDK/JSONModels/MXEvent.m
@@ -196,42 +196,33 @@ uint64_t const kMXUndefinedTimestamp = (uint64_t)-1;
     return age;
 }
 
-- (NSDictionary *)originalDictionary
+- (NSDictionary *)JSONDictionary
 {
-    NSMutableDictionary *originalDictionary = [NSMutableDictionary dictionary];
-    if (originalDictionary)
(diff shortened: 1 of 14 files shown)


MatrixSDK/JSONModels/MXEvent.m
```diff
@@ -196,20 +196,30 @@ - (NSUInteger)age
     return age;
 }
 
-- (NSDictionary *)originalDictionary
+- (NSDictionary *)JSONDictionary
```
giomfo: Note returned dictionary is slightly different from the original JSON here because it does not consider 'unsigned' key introduced in v2.

giomfo:
LGTM - the handling of 'unsigned' key in event description must be clarified with server team.

MatrixSDK/JSONModels/MXEvent.m
manuroe: Agreed. The issue is that the hs fills the event differently according if it comes from /sync v2 or from /messages.

PR #71, opened 2016-02-01
Room page loading improvement

manuroe:
[MXRoom paginateBackMessages]: added an argument to load only available messages from the store


giomfo:
LGTM

PR #72, opened 2016-02-02
Bug Fix: App crashes on resume

giomfo:
Take into account pending Pause (if any) at the end of resume.

### DIFF
--- a/MatrixSDK/MXSession.h
+++ b/MatrixSDK/MXSession.h
@@ -254,7 +254,7 @@ FOUNDATION_EXPORT NSString *const kMXSessionNoRoomTag;
                    has received uptodate data/events. The live event listening
                    (long polling) is not launched yet.
                    CAUTION The session state is updated (to MXSessionStateRunning) after
-                   calling this block. It MUST not be modified by this block.
+                   calling this block. It SHOULD not be modified by this block.
  */
 - (void)resume:(void (^)())resumeDone;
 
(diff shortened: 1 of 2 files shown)


manuroe:
The pending pause is due to MatrixKit, not the SDK. So the comments about it seem out of scope.
It seems we fix something in the kit but at the sdk level. This is the issue of the 2 async responses for one request (one by callback, one by an event via NSNotificationCenter).

Instead of talking of pending pause, we could indicate in the spec of the methods how the 2 responses are sent (the session state is updated before the call of the callback)
Then the kit will simply conform to them and will no more crash.

giomfo:
First the comment is still relevant, it mentions the SDK user (not the SDK).

You're right about the 2 async responses for one request. This is the weak point. I think we should remove one of them.
The callback is only used to update the user presence (online). This operation may be performed on session state change when the session switches in running mode. I suggest removing the resume callback.

If you agree, I will prepare this change, and update the commit.

manuroe:
> First the comment is still relevant, it mentions the SDK user (not the SDK).
Hmm. The comment looks like the sdk must forecast and anticipate the SDK user behaviour which is strange.

The failure seems to be still required. It would look weird to have a failure callback without a success callback, no?

manuroe:
LGTM

giomfo:
Manu, solution has been change, please review again

manuroe:
LGTM

PR #74, opened 2016-02-11
Moving to r0 - Step #1

manuroe:
This PR replaces calls to v1 and v2_alpha apis by r0, which is configurable via MXRestClient.apiPathPrefix.
v1 global initialSync and related objects have been removed.

In step \#2, call of room initialSync will be removed as v2 /sync do the job. All comments relative to v1/v2 will also be removed


giomfo:
LGTM

PR #75, opened 2016-02-11
BugFix vector-ios #75

giomfo:
Email login broken?


manuroe:
I wonder if it is a good thing to let the registerOrLoginWithUser method guess what is the identity type passed by the sdk user.
But, as far as we support only matrix user id and email, it should not be not a problem.
So, LGTM.

PR #78, opened 2016-02-23
Event timeline

manuroe:
Created MXEventTimelime to manage a list of continuous events.
MXRoom has now a liveTimeline property that manages live events and state of the room. All messages handling and pagination code has moved from MXRoom to MXEventTimelime.

MXEventTimelime is generic enough to manage live events and events that will come from the context API. Only live events is implemented for now.


giomfo:
LGTM

PR #79, opened 2016-02-23
SYIOS-208: [MXSession startWithMessagesLimit]: if defined, the limit …

manuroe:
…argument is now passed to /sync requests


giomfo:
LGTM

PR #80, opened 2016-02-26
Remove MXEventDirectionSync

manuroe:
Fixed https://matrix.org/jira/browse/SYIOS-207.

MXEventDirectionSync was used to identify events coming from an initialSync. They were managed differently than MXEventDirectionForwards events which came from /events.
In sync v2, everything comes from a single API, /sync, MXEventDirectionForwards. It is no more possible to make such distinction.

MXEventDirectionSync has been removed. MXEventDirection* has been renamed to MXTimelineDirection*.
MXEventTimeline.m has been refactored to match sync v2 mechanism better (no more handleLiveEvent for example as live event has no more sense with /sync v2).

### DIFF
--- a/MatrixSDK/Data/MXEventListener.h
+++ b/MatrixSDK/Data/MXEventListener.h
@@ -17,6 +17,7 @@
 #import <Foundation/Foundation.h>
 
 #import "MXEvent.h"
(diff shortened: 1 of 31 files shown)


giomfo:
LGTM

PR #81, opened 2016-03-01
Create a timeline from a past event

manuroe:
This is then possible to paginate backwards and forwards.
The implementation makes use of /context and /messages in forwards mode.

### DIFF
--- a/MatrixSDK/Data/MXEventTimeline.m
+++ b/MatrixSDK/Data/MXEventTimeline.m
@@ -84,6 +84,30 @@ NSString *const kMXRoomInviteStateEventIdPrefix = @"invite-";
     }
 }
 
+- (MXHTTPOperation *)loadContextWithLimit:(NSUInteger)limit success:(void (^)())success failure:(void (^)(NSError *))failure
+{
+    NSAssert(_initialEventId, @"[MXEventTimeline] loadContextWithLimit cannot be called on live timeline");
+
+    // Get the context around the initial event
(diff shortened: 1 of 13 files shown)


MatrixSDK/Data/MXEventTimeline.m
```diff
@@ -125,16 +137,58 @@ - (void)resetPagination
...
+
+        for (MXEvent *event in eventContext.eventsAfter)
+        {
+            [self addEvent:event direction:MXTimelineDirectionForwards fromStore:NO];
+        }
+
+        [store storePaginationTokenOfRoom:room.roomId andToken:eventContext.start];
+        forwardsPaginationToken = eventContext.end;
+
+        success();
```
giomfo: I would prefer test success block before calling it

manuroe: I can add a NSParameterAssert to check its existence. It is mandatory as the sdk user can do nothing while the request is not comple.

MatrixSDK/Data/MXEventTimeline.m
```diff
@@ -167,73 +221,77 @@ - (MXHTTPOperation *)paginate:(NSUInteger)numItems direction:(MXTimelineDirectio
...
+        paginationToken = [store paginationTokenOfRoom:_state.roomId];
+        if (nil == paginationToken)
+        {
+            paginationToken = @"END";
+        }
     }
     else
     {
-        NSAssert(NO, @"Cannot paginate forwards on a live timeline");
+        paginationToken = forwardsPaginationToken;
```
giomfo: Why not check hasReachedHomeServerForwardsPaginationEnd flag before triggering forward pagination?

Issue #82, opened 2016-03-02
MXRoom.unreadEvents and [MXStore unreadEvents::] should be removed


manuroe:
because they can lie since v2 /sync and the potential gaps in received messages.

It should be replaced by 'unreadEventsCount' which is the most relevant information for an app.
This count should be provided by the homeserver but it is not available yet.

giomfo:
But it is not so easy for the server to compute the right unread count
because the server does not know which kind/type of events the client actually considers.
The server can only return the number of unread events whatever their type, whereas the client displays only a subset of these events (ignore some membership event for example)

We will limit this information to a boolean value: has unread or not.

PR #83, opened 2016-03-03
BugFix SYIOS-202: IOS should no longer reset badge count on launch.

giomfo:
+ fix: MXRoom.unreadEvents and [MXStore unreadEvents::] should be removed #82


manuroe:
LGTM

PR #85, opened 2016-03-24
Fixed https://github.com/vector-im/vector-ios/issues/176

manuroe:
User profile: user settings may be modified during pagination in past timeline


giomfo:
LGTM

PR #90, opened 2016-04-06
BugFix: Cancel correctly pending operation

giomfo:
It was not possible to cancel an operation for which the app was waiting for the connection.


manuroe:
Interesting.
LGTM

PR #100, opened 2016-04-25
MXFileStore: Improved storage and loading of read receipts

manuroe:
Loading is around 30% faster than before. 
Yeah, decoding dictionary consumes time.


giomfo:
lgtm

PR #105, opened 2016-05-13
Bug Fix: all blank after upgrade; no spinner

giomfo:
https://github.com/vector-im/vector-ios/issues/311

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -259,12 +259,18 @@ typedef void (^MXOnResumeDone)();
 
         // Set the store before going further
         __weak typeof(self) weakSelf = self;
+
         [self setStore:store success:^{
 
             // Then, start again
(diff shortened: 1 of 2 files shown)


MatrixSDK/MXSession.m
```diff
@@ -748,9 +765,6 @@ - (void)serverSyncWithServerTimeout:(NSUInteger)serverTimeout
         // Check whether the caller wants to handle error himself
         if (failure)
         {
-            // Inform the app there is a problem with the connection to the homeserver
```
manuroe: Was it removed to avoid double changes of the state? One here and one by the caller?

manuroe:
LGTM

MatrixSDK/MXSession.m
giomfo: Yes

PR #108, opened 2016-05-20
MXSession: Detect when the access token is no more valid

manuroe:
https://github.com/vector-im/vector-ios/issues/247


giomfo:
lgtm

PR #112, opened 2016-06-15
MXRoomPowerLevels: If the room contains no power_levels event, the st…

giomfo:
…ate_default is 0. The events_default is 0 in either of these cases.

But, state_default defaults to 50 if there is a power level event but no states_default key.

+ Support legacy CamelCase keys.
+ Force file store flush.


manuroe:
LGTM

Issue #113, opened 2016-07-01
Update AFNetworking to 3.0


manuroe:
That implies some refactoring:
https://github.com/AFNetworking/AFNetworking/wiki/AFNetworking-3.0-Migration-Guide

PR #114, opened 2016-07-07
Update AFNetworking to 3.1

manuroe:
https://github.com/matrix-org/matrix-ios-sdk/issues/113

I try to keep the same behaviour we had with the previous AFNetworking version. Specifically for media upload when the app goes in background.

I tested SSL in different situations but I failed to enter in this code https://github.com/matrix-org/matrix-ios-sdk/blob/1feaa3ddc4c1a49f9fb7438ac202721024dbc732/MatrixSDK/Utils/MXHTTPClient.m#L455-L471.
As I cannot understand how we can enter in this else block, I have removed it.

There is an API break in MXRestClient with the introduction of NSProgress.

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -27,6 +27,6 @@ Pod::Spec.new do |s|
 
   s.requires_arc  = true
 
-  s.dependency 'AFNetworking', '~> 2.6.0'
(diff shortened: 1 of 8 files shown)


giomfo:
sgtm

PR #122, opened 2016-08-23
MXCallStackCall: Do not force the main speaker on video call.

giomfo:
Let the Application handle the audio route. It is useful in case of headset use.

MXJingleCallStackCall: Run callbacks on the main thread.


giomfo:
Should fix: https://github.com/vector-im/vector-ios/issues/518

manuroe:
LGTM

PR #123, opened 2016-08-24
Add store events enumerator

manuroe:
The goal is to be able to have several stored messages iterators on a same room in parallel.
We are then able to do pagination AND, in parallel, enumerate stored messages to find the right last message.
This fixes https://github.com/vector-im/vector-ios/issues/520


giomfo:
SGTM

PR #124, opened 2016-08-24
MXFileStore: Add a background task during load and commit phases

manuroe:
This prevents these "long" operations from being suspended while the app in background


MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -476,6 +502,12 @@ - (void)commit
         dispatch_async(dispatchQueue, ^(void){
             [[NSFileManager defaultManager] removeItemAtPath:storeBackupPath error:nil];
             backupEventStreamToken = nil;
+
+            // Release the background task
+            dispatch_sync(dispatch_get_main_queue(), ^(void){
```
giomfo: Why not use dispatch_async here

manuroe: bad c+p. thanks

PR #125, opened 2016-08-24
Video Call: the remote and local video are not scaled to fill the vid…

giomfo:
…eo container

https://github.com/vector-im/vector-ios/issues/537


manuroe:
LGTM

PR #130, opened 2016-08-29
MXTools: Add methods to create permalinks to room or event

manuroe:
https://github.com/vector-im/vector-ios/issues/547


giomfo:
lgtm

PR #131, opened 2016-08-29
Bug Fix - Chat screen: New message(s) notification

giomfo:
https://github.com/vector-im/vector-ios/issues/532

MXRoom Changes:
- Define 'localUnreadEventCount' property: The number of unread events wrote in the store.
- Add API [acknowledgeEvent:]: If the event was not acknowledged yet, this method acknowlegdes it by sending a receipt event. This will indicate to the homeserver that the user has read up to this event.


manuroe:
LGTM

PR #133, opened 2016-09-03
Stop leaking background tasks, which kill the app after 180s of bg.

ara4n:
We were tracking the ID of "the current background task" in a member field on MXFileStore.
However, there's nothing to stop parallel background tasks from happening, at which point
we lose track of the background task in flight and so cannot terminate it; so it leaks.

Also adds a shedload of debug logging for commit operations to the filestore.


manuroe:
LGTM

PR #135, opened 2016-09-06
MXJingleCallStackCall: Fix: Show the front camera self preview as a mirror

manuroe:
It fixes: When screen is locked, rotating the screen landscape makes local video preview go upside down https://github.com/vector-im/vector-ios/issues/519.


giomfo:
OK

PR #136, opened 2016-09-06
MXSession: Expose the list of user ids for whom a 1:1 room exists.

giomfo:
https://github.com/vector-im/vector-ios/issues/529


manuroe:
Worth a test but it LGTM

PR #137, opened 2016-09-07
MXSession: Add a mechanism to keep the event streams running

manuroe:
even if the app goes in background

Make MXCall use it in order to fix "Locking phone whilst setting up a call interrupts the call setup" (https://github.com/vector-im/vector-ios/issues/552)


giomfo:
SGTM

PR #140, opened 2016-09-15
App crashes when we start a chat during initial sync

manuroe:
https://github.com/vector-im/vector-ios/issues/623

MXSession: In case of initialSync, mxsession.myUser.userId must be available before changing the state to MXSessionStateStoreDataReady.


giomfo:
LGTM

PR #144, opened 2016-09-26
Bug Fix - App got stuck and permanently spinning.

giomfo:
https://github.com/vector-im/vector-ios/issues/655

Cancel all the tasks currently run by the managed session when the background task expired (instead of invalidate the session).


giomfo:
When a task is cancelled, the MXHTTPOperation failed with NSURLErrorCancelled error code. So the MXHTTPOperation creator is notified.

PR #146, opened 2016-10-05
Bug Fix - Use contains_url filter for navigating files and showing a …

giomfo:
…file index for a room.

vector-im/vector-ios#652

MXRestClient: Support 3 filtering modes on attachments during the messages search (see `MXMessagesSearchMediaFilter`):
- Default value. No limitation.
- Only the messages which contain an URL are concerned by the search request.
- The messages which contain an URL are not concerned by the search request.

### DIFF
--- a/MatrixSDK/MXRestClient.h
+++ b/MatrixSDK/MXRestClient.h
@@ -87,28 +87,6 @@ typedef enum : NSUInteger
 } MXThumbnailingMethod;
 
 /**
- List the filtering options used during the messages search. This option concerns the attachments handling.
- */
-typedef enum : NSUInteger
-{
-    /**
-     Default value. No limitation.
-     */
-    MXMessagesSearchMediaFilterUndefined,
-    
-    /**
-     Only the messages which contain an URL are concerned by the search request.
-     */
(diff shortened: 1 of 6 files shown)


MatrixSDK/MXRestClient.h
```diff
@@ -71,20 +71,41 @@ FOUNDATION_EXPORT NSString *const kMXRestClientErrorDomain;
...
-     "crop" trys to return an image where the width and height are close to the requested size
+     "crop" tries to return an image where the width and height are close to the requested size
      and the aspect matches the requested size. The client should scale the image if it needs to
      fit within a given rectangle.
      */
     MXThumbnailingMethodCrop
 } MXThumbnailingMethod;
 
+/**
+ List the filtering options used during the messages search. This option concerns the attachments handling.
```
manuroe: I am not sure about the naming. 'Attachment' has no particular meaning for Matrix events. 'Media' is a bit too specific regarding what contains_url do, ie filtering events which have a 'content' with a 'url' field. Let's discuss it in 15min :)

manuroe: The MXMessagesSearchMediaFilter enum defintion can be now removed

PR #147, opened 2016-10-06
API break: MXRestClient: Remove registerWithUser and loginWithUser methods

manuroe:
which worked only with old CS auth API.

Add registerWithLoginType and loginWithLoginType which do the job with new CS auth api for dummy and password flows.

Tests have been updated.

### DIFF
--- a/MatrixSDK/MXRestClient.h
+++ b/MatrixSDK/MXRestClient.h
@@ -175,7 +175,7 @@ typedef enum : NSUInteger
 /**
  Generic registration action request.
 
- As described in http://matrix.org/docs/spec/#registration-and-login some registration flows require to
+ As described in http://matrix.org/docs/spec/client_server/r0.2.0.html#client-authentication some registration flows require to
  complete several stages in order to complete user registration.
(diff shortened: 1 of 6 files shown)


MatrixSDK/MXRestClient.h
```diff
@@ -195,21 +195,21 @@ typedef enum : NSUInteger
...
- It implements the password-based registration flow described at
- http://matrix.org/docs/spec/#password-based
+ Register a user.
+ 
+ This method manages the full flow for simple login types and returns the credentials of the newly created matrix user.
 
- @param user the user id (ex: "@bob:matrix.org") or the user id localpart (ex: "bob") of the user to register.
+ @param loginType the login type. Only kMXLoginFlowTypePassword and kMXLoginFlowTypeDummy (m.login.password and m.login.dummy) are supported.
+ @param username the user id (ex: "@bob:matrix.org") or the user id localpart (ex: "bob") of the user to register. Can be nil.
  @param password his password.
```
giomfo: is it nullable too?

PR #151, opened 2016-10-20
Vector 714

giomfo:


MatrixSDK/MXSession.h
```diff
@@ -427,6 +430,9 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
                     visibility:(MXRoomDirectoryVisibility)visibility
```
manuroe: Is is possible to create a new method rather than making the existing one complicate ?
It breaks a lot of things for no added value most of the time. We already have 2 "createRoom", we can have 3.

MatrixSDK/MXRestClient.h
```diff
@@ -895,6 +899,9 @@ typedef enum : NSUInteger
                     visibility:(MXRoomDirectoryVisibility)visibility
                      roomAlias:(NSString*)roomAlias
                          topic:(NSString*)topic
+                        invite:(NSArray<NSString*>*)inviteArray
```
manuroe: Same remark as below

Issue #149, opened 2016-10-13
Support `m.direct` event in account_data


giomfo:
see https://github.com/vector-im/vector-web/issues/2070:

We've finally defined the behaviour for tracking 1:1 rooms:

the idea is to use User account_data to track the list of DMs we see for a given remote user
which would have a m.direct event in it, whose content gives a bunch of mxid to list-of-roomid mappings (we could split these into separate m.direct events, but these feel ugly)
for bootstrapping: the first time Vector loads without any m.direct event, i'd expect it to synthesize one from the current heuristics of what counts as a 1:1 room.

PR #152, opened 2016-10-27
Feature - Support `m.direct` event in account_data

giomfo:
#149

When no direct chat is listed in account data, we synthesize them from the current heuristics of what counts as a 1:1 room.

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -770,6 +770,27 @@ NSString *const kMXRoomDidUpdateUnreadNotification = @"kMXRoomDidUpdateUnreadNot
     return ([self getDirectUserId] != nil);
 }
 
+- (BOOL)looksLikeDirect
+{
+    BOOL kicked;
+    if (self.state.membership == kMXMembershipStringLeave)
+    {
+        MXRoomMember *member = [self.state memberWithUserId:mxSession.myUser.userId];
+        kicked = ![member.originalEvent.sender isEqualToString:mxSession.myUser.userId];
+    }
+    
+    if (self.state.membership == MXMembershipJoin || self.state.membership == kMXMembershipStringBan || kicked)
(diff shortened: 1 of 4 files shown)


MatrixSDK/Data/MXRoom.m
```diff
@@ -779,21 +800,24 @@ - (MXHTTPOperation*)setIsDirect:(BOOL)isDirect
...
-            NSMutableArray *roomLists = [NSMutableArray arrayWithArray:directRooms[directUserId]];
-            
-            [roomLists removeObject:self.roomId];
-            
-            if (roomLists.count)
-            {
-                [directRooms setObject:roomLists forKey:directUserId];
-            }
-            else
+            @synchronized (mxSession.directRooms)
```
manuroe: In the sdk, we avoid to use @synchronized as the contract is to call the SDK API from the UI thread.
If we start doing thread locks, we will need to do it for every SDK methods

MatrixSDK/Data/MXRoom.m
```diff
@@ -824,14 +848,17 @@ - (MXHTTPOperation*)setIsDirect:(BOOL)isDirect
...
+        {
+            NSMutableArray *roomLists = (mxSession.directRooms[directUserId] ? [NSMutableArray arrayWithArray:mxSession.directRooms[directUserId]] : [NSMutableArray array]);
+            
+            [roomLists addObject:self.roomId];
+            
+            [mxSession.directRooms setObject:roomLists forKey:directUserId];
+        }
         
-        return [mxSession setDirectRooms:directRooms success:success failure:failure];
+        // Let the mxSession post the 'kMXSessionDirectRoomsDidChangeNotification' notification on account data update
```
manuroe: Other MXRoom methods does such check/wait for the caller, ie the caller does not need to call a method and then listen to a notification to get the result.

PR #156, opened 2016-11-02
Feature - Converting existing rooms to/from DMs

giomfo:
https://github.com/vector-im/vector-ios/issues/715

- Room creation: When the flag isDirect is turned on, only one user id is expected in the inviteArray. The room will be considered as direct only for the first mentioned user in case of several user ids.


manuroe:
SGTM

PR #164, opened 2016-11-14
Support for encrypting attachments

dbkr:
Adds class for taking a file, encrypting it, uploading it to the media repo and returning the object that goes in the message body.

Support for reading encrypted attachment coming soon.

### DIFF
--- a/MatrixSDK/Crypto/MXEncryptedAttachments.m
+++ b/MatrixSDK/Crypto/MXEncryptedAttachments.m
@@ -107,7 +107,8 @@
         if (plainBuf == nil || plainBuf.length == 0) break;
         
         if (buflen < plainBuf.length) {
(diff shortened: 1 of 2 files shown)


MatrixSDK/Crypto/MXEncryptedAttachments.m
```diff
@@ -0,0 +1,177 @@
...
+    } failure:^(NSError *error) {
+        failure(error);
+    }];
+}
+
++ (NSString *)base64ToUnpaddedBase64:(NSString *)base64 {
+    return [base64 stringByReplacingOccurrencesOfString:@"=" withString:@""];
+}
+
++ (NSString *)base64UrlToBase64:(NSString *)base64Url {
```
manuroe: this method and its reverse one are not used. Do we need to keep them?

PR #165, opened 2016-11-15
Don't crash on undecryptable events

dbkr:
Add format param to MXDecryptingErrorUnableToDecryptReason


dbkr:
Bah, didn't mean to include other changes.

PR #169, opened 2016-11-15
Crypto: message-replay mitigation

manuroe:
There is however an issue: it breaks permalinks or reloading events from a store.
The reason is that a permalink is opened in a separate window with its own timeline. And this timeline decodes events for a second time.


ylecollen:
LGTM

PR #187, opened 2016-11-29
E2E: Check devices to share keys with on each send

manuroe:
Instead of trying to maintain a list of devices we need to share with, just check all the devices for all the users on each send.

Port of https://github.com/matrix-org/matrix-js-sdk/pull/295


ylecollen:
LGTM

Issue #195, opened 2016-12-06
Pending new devices must be taken into account when sending a new message


manuroe:
This will avoid the patch at https://github.com/matrix-org/matrix-ios-sdk/commit/5e0e68938e5696305c19332e22171cfcbeefa8ea

PR #196, opened 2016-12-06
Crypto: Pending new devices must be taken into account when sending a…

manuroe:
… new message (#195)

and remove the patch in testAliceAndBobWithNewDevice


ylecollen:
LGTM

PR #197, opened 2016-12-06
Get user own devices keys

manuroe:
This fixes https://github.com/vector-im/riot-web/issues/2676 on iOS.
Little refactoring inside

### DIFF
--- a/MatrixSDK/Crypto/MXCrypto.m
+++ b/MatrixSDK/Crypto/MXCrypto.m
@@ -867,15 +867,20 @@
 {
     // This method is called when the initialSync was done or the session was resumed
 
-    // Catch up on any m.new_device events which arrived during the initial sync.
-    [self flushNewDeviceRequests];
-
     if (_store.deviceAnnounced)
     {
+        // Catch up on any m.new_device events which arrived during the initial sync.
+        [self flushNewDeviceRequests];
+
         NSLog(@"[MXCrypto] checkDeviceAnnounced: Already done");
         onComplete();
         return nil;
     }
+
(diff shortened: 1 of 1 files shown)


MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1051,16 +1053,11 @@ - (void)flushNewDeviceRequests
         // tight-loop.
         for (NSString *userId in usersDevicesInfoMap.userIds)
         {
-            [pendingNewDevices removeObjectsForUser:userId];
+            [pendingUsersWithNewDevices removeObject:userId];
```
richvdh: I think this is subject to a race:

* we start a device list download
* the other user starts a new device, and we get an m.new_device event for the new device
* we get the response to the device list download, but it doesn't yet include the new device
* we clear `pendingUsersWithNewDevices`, but we still don't know about the new device

The js-sdk clears the pending list before starting the device list download to avoid this.

ylecollen:
LGTM

MatrixSDK/Crypto/MXCrypto.m
manuroe: Agreed :/
The code changed to this in order to fix another race where the user sends a message whereas the code is downloading a new device key. (https://github.com/matrix-org/matrix-ios-sdk/pull/196)

richvdh: Ah. might need a separate "requests in progress" list :/

manuroe: Unfortunately, I do not see a nicer solution.

Issue #213, opened 2016-12-27
Content prefix path.


VictorLiubchenko:
When my home server something like this https://foo.org/something, and I want upload image, your method provide me path like "/_matrix/media/v1".

In MXHTTPClient try request method 
    NSString *URLString = [[NSURL URLWithString:path relativeToURL:httpManager.baseURL] absoluteString];

path start with /_matrix/media/v1.... and base url look like https://foo.org/something/ and we have result as https://foo.org/_matrix/media/v1.... (http://nshipster.com/nsurl/)

### Could you make kMXContentPrefixPath changeable or properties?

giomfo:
@VictorLiubchenko thanks for your remark.

I prepared a change in MXHTTPClient try request method in order to handle correctly the homeserver whose the url is a subdirectory (and not a root). My change is waiting for reviewing.

Why did you report this issue only for the image uploading? It should impact all your requests, no?

VictorLiubchenko:
Thanks for the quick answer)

MXRestClient has property "apiPathPrefix", it help in other case. For example 
set apiPathPrefix = /something/_matrix/client/r0, and

"https://foo.org/something/" after added path "/something/_matrix/client/r0" looks like "https://foo.org/something/_matrix/client/r0".

giomfo:
@VictorLiubchenko fixed in develop branch

PR #214, opened 2016-12-27
Bug Fix - MXHTTPClient: Handle correctly the case where the homeserve…

giomfo:
…r url is a subdirectory.

In MXHTTPClient try request method, remove the potential leading slash for the provided path, so that NSURL +URLWithString:relativeToURL: works as expected.

This should fix "Content prefix path" #213.

### DIFF
--- a/MatrixSDK/Utils/MXHTTPClient.m
+++ b/MatrixSDK/Utils/MXHTTPClient.m
@@ -204,11 +204,6 @@ NSString * const MXHTTPClientErrorResponseDataKey = @"com.matrixsdk.httpclient.e
         path = [path stringByAppendingString:[NSString stringWithFormat:@"%@access_token=%@", urlSeparator, accessToken]];
     }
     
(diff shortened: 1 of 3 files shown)


MatrixSDK/Utils/MXHTTPClient.m
```diff
@@ -204,6 +204,11 @@ - (void)tryRequest:(MXHTTPOperation*)mxHTTPOperation
         path = [path stringByAppendingString:[NSString stringWithFormat:@"%@access_token=%@", urlSeparator, accessToken]];
     }
     
+    // Remove the potential leading slash for the provided path, so that NSURL +URLWithString:relativeToURL: works as expected.
```
giomfo: Then I think we should just follow the Victor's suggestion: "Make kMXContentPrefixPath changeable or properties"

giomfo: I will do both

antoni12753 (commented):
Hs

Issue #216, opened 2017-01-14
macOS port


aapierce0:
I'm interested in building a true native macOS client. Is there any reason this framework couldn't (or shouldn't) include a macOS build target?

I'm currently looking into adding it myself. A lot of the work appears to be replacing the `UI*` classes with their `NS*` counterparts, like so:

    #if TARGET_OS_IPHONE
        UIImage* foo;
    #elif TARGET_OS_OSX
        NSImage* foo;
    #endif 

... and then `#import <Cocoa/Cocoa.h>` instead of `<UIKit/UIKit.h>` when the platform is macOS

I notice that the [OLMKit](https://matrix.org/git/olm/about/) pod currently doesn't support macOS, but after a very brief (10 min) skim through that project, it looks like that would be portable to macOS as well. However, since it's hosted on matrix.org instead of github, I'm not sure what the protocol is for contributing.

ara4n:
It would be absolutely awesome to have a true native macOS client.  matrix-ios-sdk itself is a relatively low level library that wraps and manages the Matrix HTTP API itself, and can probably be made to build for macOS very easily indeed.  However, all the UI stuff (e.g. UIImage and friends) lives in matrix-ios-kit (MatrixKit), which I think is more where the porting work would need to happen.

Many bits of MatrixKit probably can be made to support both targets, but there's going to be some higher level layers which will probably need entirely separate macOS variants rather than using `#ifdefs`.  For instance, I suspect you'd need to build a separate tree of XIBs for macOS in addition to the current iOS ones, and of course the storyboard of the app itself will need to be different.

My suggestion (beware: i've not actually hacked on matrix-ios-kit and matrix-ios-sdk that much and am a bit rusty on iOS dev) would be to fork matrix-ios-sdk and matrix-ios-kit on Github, take the MatrixKitSample app (https://github.com/matrix-org/matrix-ios-kit/tree/develop/Samples/MatrixKitSample), and try working your way through, adding in ifdefs, as well as 'forking' the XIBs & Storyboards to support macOS.  Perhaps you may not need to fork any actual controllers themselves.

In terms of OLMKit: libolm itself builds on OSX, although the Makefile requires minimal tweaking; i currently have:

```
diff --git a/Makefile b/Makefile
index 438fa68..db0e701 100644
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,8 @@ EMCC = emcc
 AFL_CC = afl-gcc
[...]

PR #218, opened 2017-01-16
macOS Compatibility

aapierce0:
Enables macOS compatibility via cocoapods. See #216 

**Note:** The current version of OLMKit currently won't work with a macOS build target over cocoapods. I forked the OLMKit repository at https://github.com/aapierce0/OLMKit which updates the podspec to include a macOS deployment target. (I also replaced a `UIKit` import with `Foundation`) That should be the only change that's needed for macOS compatibility.

Signed-off-by: Avery Pierce <aapierce0@gmail.com>


aapierce0:
That should do it.

PR #230, opened 2017-01-26
Fix a few examples in the README file.

aapierce0:


jfillingham-mt:
The first example currently has 

```
[mxRestClient publicRooms:^(NSArray *rooms) {

        // rooms is an array of MXPublicRoom objects containing information like room id
        NSLog(@"The public rooms are: %@", rooms);

    } failure:^(MXError *error) {
    }];
```

which does not appear to be a function (anymore?).

PR #232, opened 2017-01-30
Room summary part 1

manuroe:
This is still WIP but the concept is here.

I marked as TODO(summary) things that needs (or not) be changed.

The next step is make the app use first room summary and then load rooms states and messages asynchronously on demand.


MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -0,0 +1,120 @@
...
+        {
+            [summary resetLastMessage:nil failure:^(NSError *error) {
+                NSLog(@"[MXRoomSummaryUpdater] updateRoomSummary: Cannot reset last message after redaction. Room: %@", summary.roomId);
+            }];
+        }
+        return NO;
+    }
+
+    // Accept redacted event only if configured
+    if (_ignoreRedactedEvent && event.redactedBecause)
```
giomfo: You may used `isRedactedEvent` property from MXEvent class

PR #250, opened 2017-02-20
MXRestClient: Support phone number validation

giomfo:


MatrixSDK/MXRestClient.m
```diff
@@ -3966,21 +4018,41 @@ - (MXHTTPOperation*)requestEmailValidation:(NSString*)email
...
+- (MXHTTPOperation *)submit3PIDValidationToken:(NSString *)token
+                                        medium:(NSString *)medium
+                                  clientSecret:(NSString *)clientSecret
+                                           sid:(NSString *)sid
+                                       success:(void (^)())success
+                                       failure:(void (^)(NSError *))failure
 {
+    NSString *path;
+    
+    if ([medium isEqualToString:kMX3PIDMediumEmail])
```
manuroe: I am not sure we need to limit the validation process to email and msisdn. If a new medium is created, we will need to deliver a new sdk.
Testing medium.count should be enough as sanity check.

MatrixSDK/MXRestClient.h
```diff
@@ -1455,24 +1455,56 @@ typedef enum : NSUInteger
...
+ Request the validation of a phone number.
+ 
+ The identity server will send a validation token by sms. The end user
+ will have to send this token by using [MXRestClient submit3PIDValidationToken].
+ 
+ Use the returned sid to complete operations that require authenticated phone number
+ like [MXRestClient add3PID:].
+ 
+ @param phoneNumber the phone number (in international or national format).
+ @param countryCode the iso country code (required even if the phone number is internationalised).
```
manuroe: Is it possible to add a ref to the spec of the iso code expected ?

MatrixSDK/MXRestClient.m
giomfo: but medium is not an enum value!
Eventually, we should use medium string to directly complete the path. WDYT?

MatrixSDK/MXRestClient.h
giomfo: more precisely: ISO 3166-1 country code representation

MatrixSDK/MXRestClient.m
manuroe: Yes, medium is a string but the code in this PR treats it like an enum with a finite number of values with no possibility of extension.

MatrixSDK/MXRestClient.h
manuroe: Could you please add it to the code?

PR #297, opened 2017-05-29
Fix type of received value

samuel-gallet:


manuroe:
Hi Samuel,
Thanks for the PR.

How/where do you get a numeric value for ```sid```?
It must be a string.

samuel-gallet:
Hi,
If you are sure about string type of sid, it should be an error from our server.
I'm sorry

manuroe:
Yes, all ids are strings in matrix. 
The spec says for ```sid```:
```
"sid": {
    "description": "The session identifier given by the Identity Server.", 
    "type": "string"
}
```

PR #298, opened 2017-05-29
Remove depency to WebRTC and add parameter to use GA

samuel-gallet:
I'm ready to discuss about this fix.

The problem is the use of "use_frameworks!" parameter in the Podfile includ in swift project.

Indeed, with cocoa pods 1.x and the use_frameworks parameters, pod are built atomically and cannot refer to library that are not directly linked to the pod.

So you can have build issue like:

Undefined symbols '_GAI' not found in MXFileStore.o

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -30,5 +30,6 @@ Pod::Spec.new do |s|
 
   s.dependency 'AFNetworking', '~> 3.1.0'
   s.dependency 'GZIP', '~> 1.1.1'
+  s.dependency 'WebRTC', '~> 56.10'
 
 end
--- a/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
+++ b/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
@@ -30,11 +30,9 @@
 
 #import "MXEnumConstants.h"
 
-#ifdef MX_GA
 #import "GAI.h"
 #import "GAIFields.h"
 #import "GAIDictionaryBuilder.h"
-#endif
 
 NSUInteger const kMXFileVersion = 39;
 
--- a/MatrixSDK/MXSDKOptions.h
+++ b/MatrixSDK/MXSDKOptions.h
@@ -21,22 +21,6 @@
 #pragma mark - Build time options
 
 #pragma mark Automatic enabling
-/**
- These flags are automatically enabled if the application Xcode workspace contains
- their associated Cocoa pods.
- */
-
-/**
- VoIP with libjingle.
-
- Application can use the libjingle build pod provided at
- https://github.com/Anakros/WebRTC.git :
-
-     pod 'WebRTC'
- */
-#if __has_include(<WebRTC/RTCPeerConnection.h>)
-#define MX_CALL_STACK_JINGLE
-#endif
 
 /**
  Crypto.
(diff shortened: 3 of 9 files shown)


manuroe:
We want the application to be able to choose a VoIP stack. There are already several wrappers to other VoIP libs.
This PR sticks the SDK, and consequently apps,  to one of them. So, we cannot afford it.

samuel-gallet:
I understand your point.
However, your SDK cannot be used in a swift project.
I think it's a serious issue

manuroe:
Do you have some experience with Carthage?
I feel I spend 20% of my time hacking over Cocoapods. Carthage seems less intrusive but I have never played with it.

samuel-gallet:
If you want to use Carthage, you will lose a lot of your SDK's users.

In my opinion, if you want the user choose between some implementations, the best choice is to use protocols.

manuroe:
Protocols for VoIP are already there.
The real issue is that the wrapper that implements it with the ```WebRTC``` stack lib is in the SDK code. 
We used to put these wrappers in other GH repos with their own pods like https://github.com/matrix-org/matrix-ios-openwebrtc-wrapper/ but it added additional complexity to maintain all the things together.

I am slowly moving my mind to propose this wrapper as the default voip stack coming with the SDK. App will be then still free to use another wrapper.

manuroe:
> I am slowly moving my mind to propose this wrapper as the default voip stack coming with the SDK.

That will not work. ```WebRTC``` is iOS only (https://github.com/Anakros/WebRTC/blob/master/WebRTC.podspec#L14). We have MacOS developpers :/

samuel-gallet:
You should not provide a default VOIP stack.
But you can make an other pod project, which implement the WebRTC VOIP stack for exemple, and let the user linked the appropriated implementation by initializing the SDK in the app.

There is also the same problem by using GoogleAnalytics as an implicit depandency

manuroe:
Yeah. I am just checking if there are other options. 
Additional pod is what I did at first and I found myself spending a ridiculous amount of time syncing sdk and wrapper pods. ``` #if __has_include(<WebRTC/RTCPeerConnection.h>)``` was a nice trick to manage it in objc.
Plus that makes another thing to release.

The GA issue is handled by https://github.com/matrix-org/matrix-ios-sdk/pull/256 which I did not take time to handle. Shame on me.

samuel-gallet:
I'm afraid there are not other options :/.
Just to let you know that I already try adding this to my post_installer in my Podfile:
`

      #Add linker to GA in order to make it compiler
      if config.build_settings['PRODUCT_NAME'] == "MatrixSDK"
        config.build_settings['LIBRARY_SEARCH_PATHS'] = ["$(inherited)", "$(PODS_ROOT)/GoogleAnalytics/Libraries"]
        config.build_settings['FRAMEWORK_SEARCH_PATHS'] = ["$(inherited)", "$(PODS_ROOT)/WebRTC"]
        config.build_settings['OTHER_LDFLAGS'] = ["$(inherited)", %Q{-Wl -undefined dynamic_lookup}]
      end

`

It works when you are building your app. However, when you archive it, the app crasg

manuroe:
I wonder if putting the wrapper code into a separate CocoaPods subspec (like in http://www.dbotha.com/2014/12/04/optional-cocoapod-dependencies/) would make Swift happy.

samuel-gallet:
I already tried this and I could not make it build in Swift project.
However, AFNetworking is doing the same thing with UIKit https://github.com/AFNetworking/AFNetworking/tree/master/UIKit%2BAFNetworking and it seems to work.
So it should be a way to do this

PR #299, opened 2017-06-01
Crypto: Fix race in device list updates

manuroe:
Don't consider device lists up-to-date when we have another request for the
relevant user in the queue.

Port of https://github.com/matrix-org/matrix-js-sdk/pull/431


ylecollen:
SGTM

PR #302, opened 2017-06-07
Allow App to use custom certificate

samuel-gallet:
AFSecurityPolicy try to retrieve certificate in the AFNetworking Bundle
However, with Cocoapods 1.0 the pod bundle is not equal to the app
Bundle. So certificates are not retrieved


samuel-gallet:
Samuel Gallet - samuel.gallet@applidium.com

PR #306, opened 2017-06-14
CallKit support

morozkin:
`MXCallKitAdapter` is an entity through which we interact with CallKit API

The most interesting part is `MXJingleCallAudioSessionConfigurator`. There is no public API in WebRTC iOS SDK to configure `AVAudioSession`, it is only internal. CallKit has its own rules for where and when we need to configure and activate AVAudioSession. We need to configure AVAudioSession when `provider:performStartCallAction:` is called and in completion block of `reportNewIncomingCallWithUUID:update:completion`. So i decide to extract configuration values from WebRTC iOS SDK and put them into class called `MXJingleCallAudioSessionConfigurator`. So with it help we can configure `AVAudioSession` like it does WebRTC internally. Here is the WebRTC [sources](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/sdk/objc/Framework/Classes/Audio/RTCAudioSessionConfiguration.m).

Another important thing is `AVAudioSession` activation and deactivation. CallKit says us when it activated or deactivated `AVAudioSession` and we need to inform WebRTC somehow. But we also don't have any public API for that. They create new methods for better integration with CallKit and the will be available in v60. The funny fact is that they are located in internal class:)
Because of all these funny things i decided to use Obj-C runtime to have the same behaviour as methods they're going to release in v60. You can check additional comments in the code.

The next important thing i want to talk about is `CXProvider` configuration. For the moment i use a static method for it, but it must be fixed because we must provide chance for our SDK users to configure CallKit as they want. I see several options. In both situations i want to add a parameter for init. In the first case we can pass  `CXProviderConfiguration` instance, in the second we can create class like `MXCallKitConfiguration` which will mirror `CXProviderConfiguration` interface and also will have a static method `defaultConfiguration`. We also can use `init` as a convenience initializer, it will call `initWithConfiguration:` and pass a default configuration to it. Need your help with this.

### DIFF
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
@@ -1,0 +1,26 @@
+/*
+ Copyright 2017 Vector Creations Ltd
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+
+@import Foundation;
+
+@class MXCall;
+
+@interface MXCallKitAdapter : NSObject
+
+- (void)reportIncomingCall:(MXCall *)call;
+
+@end
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
@@ -14,13 +14,24 @@
  limitations under the License.
  */
 
-
 @import Foundation;
+
+NS_ASSUME_NONNULL_BEGIN
 
 @class MXCall;
 
 @interface MXCallKitAdapter : NSObject
 
+- (void)startCall:(MXCall *)call;
+- (void)endCall:(MXCall *)call;
+
 - (void)reportIncomingCall:(MXCall *)call;
 
+- (void)reportCall:(MXCall *)call startedConnectingAtDate:(nullable NSDate *)date;
+- (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date;
+
++ (BOOL)callKitAvailable;
+
 @end
+
+NS_ASSUME_NONNULL_END
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
@@ -42,9 +42,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date;
 
 /**
- Tells if CallKit is supported by the OS
+ Tell about support of CallKit by the OS
  
- @returns true iff iOS version >= 10.0
+ @return true if iOS version >= 10.0 otherwise false
  */
 + (BOOL)callKitAvailable;
 
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
@@ -19,8 +19,14 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class MXCall;
+@protocol MXCallAudioSessionConfigurator;
 
 @interface MXCallKitAdapter : NSObject
+
+/**
+ The AVAudioSession configurator.
+ */
+@property (nonatomic, nullable, strong) id<MXCallAudioSessionConfigurator> audioSessionConfigurator;
 
 - (void)startCall:(MXCall *)call;
 - (void)endCall:(MXCall *)call;
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
@@ -19,6 +19,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class MXCall;
+@class MXCallKitConfiguration;
 @protocol MXCallAudioSessionConfigurator;
 
 @interface MXCallKitAdapter : NSObject
@@ -28,6 +29,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, nullable, strong) id<MXCallAudioSessionConfigurator> audioSessionConfigurator;
 
+- (instancetype)initWithConfiguration:(MXCallKitConfiguration *)configuration NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)init NS_UNAVAILABLE;
++ (instancetype)new NS_UNAVAILABLE;
+
 - (void)startCall:(MXCall *)call;
 - (void)endCall:(MXCall *)call;
 
@@ -36,6 +42,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)reportCall:(MXCall *)call startedConnectingAtDate:(nullable NSDate *)date;
 - (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date;
 
+/**
+ Tells if CallKit is supported by the OS
+ 
+ @returns true iff iOS version >= 10.0
+ */
 + (BOOL)callKitAvailable;
 
 @end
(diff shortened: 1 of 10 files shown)


MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -0,0 +1,285 @@
...
+    }
+    
+    CXHandle *handle = [[CXHandle alloc] initWithType:CXHandleTypeGeneric value:handleValue];
+    CXStartCallAction *action = [[CXStartCallAction alloc] initWithCallUUID:callUUID handle:handle];
+    action.contactIdentifier = contactIdentifier;
+    
+    CXTransaction *transaction = [[CXTransaction alloc] initWithAction:action];
+    [self.callController requestTransaction:transaction completion:^(NSError * _Nullable error) {
+        if (error)
+            return;
```
morozkin: I suppose we need to hang up the call at that moment

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -0,0 +1,285 @@
...
+{
+    MXCall *call = self.calls[action.callUUID];
+    if (!call)
+    {
+        [action fail];
+        return;
+    }
+    
+    [call hangup];
+    self.calls[action.UUID] = nil;
```
manuroe: I would prefer to use removeObjectForKey. I always fear about exception thrown because of setting nil value in NSMutableDictionary.

MatrixSDK/VoIP/CallStack/Jingle/MXJingleCallAudioSessionConfigurator.h
```diff
@@ -0,0 +1,53 @@
...
+#import "MXSDKOptions.h"
+
+#ifdef MX_CALL_STACK_JINGLE
+
+@class AVAudioSession;
+
+/**
+ This class helps to provide better integration level with CallKit for WebRTC
+ */
+@interface MXJingleCallAudioSessionConfigurator : NSObject
```
manuroe: I think it worths to abstract that into a MXCallAudioSessionConfigurator protocol. MXJingleCallAudioSessionConfigurator would be the implementation for Jingle.

VoIP stacks have always their own specific things. I prefer to abstract things as much as possible.

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -0,0 +1,285 @@
...
+    [self.provider reportNewIncomingCallWithUUID:callUUID update:update completion:^(NSError * _Nullable error) {
+        if (error)
+        {
+            [call hangup];
+            return;
+        }
+        
+        self.calls[callUUID] = call;
+        
+#ifdef MX_CALL_STACK_JINGLE
```
manuroe: See my above comment about an abstract MXCallAudioSessionConfigurator.
Instead of this **#ifdef** depending on a VoIP stack, we could use something like ```[callManager.callStack.audioSessionConfigurator configureAudioSessionForVideoCall:]```

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
manuroe: at this time, is it still possible to back up and make the call within the app?

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -0,0 +1,285 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MXCallKitAdapter.h"
+
+@import AVFoundation;
+@import CallKit;
+@import UIKit;
```
manuroe: We will need to use ```#if TARGET_OS_IPHONE``` compilation flag as the sdk can be also used for macos dev.
You can quickly check mac os build thanks to the validation the sdk pod. Run ```pod lib lint MatrixSDK.podspec``` in the sdk folder.

manuroe:
Concerning CXProviderConfiguration, what about MXCallKitConfiguration with a CXProviderConfiguration object property?

It is a bit like your 2nd option but we avoid to mirror all CXProviderConfiguration stuff. I like the idea of an overall MXCallKitConfiguration class because if we need additional settings that are not in CXProviderConfiguration, we will be able to manage it in MXCallKitConfiguration.

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
morozkin: It doesn't matter for me what syntax to use, but using subscripting operator looks less verbos e than calling the method. I really want you to dispel your fear about this by having a look at Clang (documentation)[https://clang.llvm.org/docs/ObjectiveCLiterals.html#dictionary-style-subscripting]. For `NSMutableDictionary` using subscripting operator for setting values is equal to calling `setObject:forKeyedSubscript:` and it is specified in the documentation that **passing nil will cause any object corresponding to aKey to be removed from the dictionary**. So there is no reason to fear about it;)

But an exception is the case when we are using the new literal syntax for creating dictionaries. If the key or the value is `nil` then we will have a runtime error.

I just want to clarify the situation. It's for you to decide which way to use.

MatrixSDK/VoIP/CallStack/Jingle/MXJingleCallAudioSessionConfigurator.h
morozkin: I fully agree with you. I will create a protocol

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
morozkin: > at this time, is it still possible to back up and make the call within the app?
The call is already running by the time we are in this codepath.

I had a look at `CXErrorCodeRequestTransactionError`, this enum represent all possible errors and i don't thin we can encounter any of them, so we can ingnore an error here

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
manuroe: Thanks for updating me.
I have encountered so many crash reports for this stupid nil value. I guess I am traumatised for life.

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
manuroe: ok. Let's ignore it.

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.h
```diff
@@ -30,8 +41,15 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)reportCall:(MXCall *)call startedConnectingAtDate:(nullable NSDate *)date;
 - (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date;
 
+/**
+ Tells if CallKit is supported by the OS
+ 
+ @returns true iff iOS version >= 10.0
```
manuroe: s/iff/if

MatrixSDK/VoIP/CallKit/MXCallKitConfiguration.h
```diff
@@ -0,0 +1,54 @@
+//
```
manuroe: Please fix the file header.

MatrixSDK/VoIP/CallKit/MXCallKitConfiguration.m
```diff
@@ -0,0 +1,38 @@
+//
```
manuroe: Please fix the file header.

MatrixSDK/VoIP/MXCallManager.h
```diff
@@ -115,6 +116,13 @@ extern NSString *const kMXCallManagerConferenceFinished;
 @property (nonatomic) id<MXCallStack> callStack;
 
 /**
+
```
manuroe: Did you want to write a comment ;) ?

MatrixSDK/VoIP/CallStack/Jingle/MXJingleCallAudioSessionConfigurator.m
```diff
@@ -0,0 +1,181 @@
...
+
+#if defined MX_CALL_STACK_JINGLE && TARGET_OS_IPHONE
+
+@import AVFoundation;
+
+#import <objc/message.h>
+#import <objc/runtime.h>
+#import <sys/utsname.h>
+
+// Preferred hardware sample rate (unit is in Hertz). The client sample rate
```
manuroe: Could you please add a link to https://chromium.googlesource.com/external/webrtc/+/master/webrtc/sdk/objc/Framework/Classes/Audio/RTCAudioSessionConfiguration.m and explain why we temporary copy the code (basically what you wrote in the PR submission)?

The code is under BSP licence, we can copy part of it but we must give them credits.

PR #307, opened 2017-06-14
Calls improvements

morozkin:


MatrixSDK/VoIP/MXCallManager.m
```diff
@@ -17,6 +17,7 @@
 #import "MXCallManager.h"
 
 #import "MXCall.h"
+#import "MXCallKitAdapter.h"
```
manuroe: is it possible to remove this dependency?
I would like to merge this PR as soon as possible but the call kit PR will need a bit more work.

PR #318, opened 2017-06-27
MXFileStore: Improve NSCoding implementation of MXEvent by storing en…

manuroe:
…um values instead of NSStrings when possible, ie for not custom event type.

Loading MatrixHQ room state events takes 840ms instead of 900ms.

We also save some RAM when loading events from the store: all known event types strings point to the same strings consts.


MatrixSDK/Utils/MXTools.m
```diff
@@ -80,6 +81,8 @@ + (void)initialize
                           kMXEventTypeStringReadMarker: @(MXEventTypeReadMarker)
                           };
 
+        revertEventTypesMap = eventTypesMap.allKeys;
```
giomfo: CAUTION: According to ios doc, the order of the elements in the array is not defined.

PR #327, opened 2017-07-14
SwiftMatrixSDK: Add swift refinements to MXSession event listeners.

aapierce0:
Signed-off-by: Avery Pierce <aapierce0@gmail.com>

### DIFF
--- a/MatrixSDK/Contrib/Swift/Data/MXSessionEventListener.swift
+++ b/MatrixSDK/Contrib/Swift/Data/MXSessionEventListener.swift
@@ -1,10 +1,18 @@
-//
-//  MXSessionEventListener.swift
-//  MatrixSDK
-//
-//  Created by Avery Pierce on 7/12/17.
-//  Copyright © 2017 matrix.org. All rights reserved.
-//
+/*
+ Copyright 2017 Avery Pierce
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


MatrixSDK/Contrib/Swift/Data/MXSessionEventListener.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  MXSessionEventListener.swift
+//  MatrixSDK
+//
+//  Created by Avery Pierce on 7/12/17.
+//  Copyright © 2017 matrix.org. All rights reserved.
```
manuroe: Unfortunately, matrix.org is not yet a foundation, nor a legal entity. Could you please remove this copyright line?
For consistency, would it be possible to use the same licence and file header as you did before (like https://github.com/matrix-org/matrix-ios-sdk/blob/6201d5709a634f7c0ea85679c04d12bf582dedab/MatrixSDK/SwiftMatrixSDK.h)?

MatrixSDK/Contrib/Swift/MXSession.swift
```diff
@@ -275,4 +275,28 @@ public extension MXSession {
...
+    
+    /**
+     Register a global listener to events related to the current session.
+     
+     The listener will receive all events including all events of all rooms.
+     
+     - parameters:
+        - types: an array of event types to listen to
+        - block: the block that will be called once a new event has been handled.
+        - returns: a reference to use to unregister the listener
```
manuroe: \"- returns" should have the same indentation as \"- parameters"

MatrixSDK/Contrib/Swift/Data/MXSessionEventListener.swift
aapierce0: Whoops! My mistake. I had fixed this in my local copy already, but I guess I forgot to push the change to Github before submitting the PR. It should be fixed now.

MatrixSDK/Contrib/Swift/MXSession.swift
aapierce0: Good catch! This is fixed in https://github.com/matrix-org/matrix-ios-sdk/pull/327/commits/17383a2b0138f71e5b4bd2c3e235afa9ab978dfb

PR #331, opened 2017-07-18
MXSession unknown token failure callback

aapierce0:
I encountered an issue in my client project where the `-[MXSession start:failure:]` method would not call the failure block if the token was invalid.

signed-off-by: Avery Pierce <aapierce0@gmail.com>

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1008,6 +1008,12 @@ typedef void (^MXOnResumeDone)();
             {
                 NSLog(@"[MXSession] The access token is no more valid. Go to MXSessionStateUnknownToken state.");
                 [self setState:MXSessionStateUnknownToken];
+                
+                // Inform the caller that an error has occurred
+                if (failure)
+                {
+                    failure(error);
+                }
 
                 // Do nothing more because without a valid access_token, the session is useless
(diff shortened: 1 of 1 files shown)


aapierce0:
Actually, I'm looking more closely at this `-[MXSession serverSyncWithServerTimeout:...]` method, and I see that there are several cases where the method can return without calling any callback, for instance when the session was closed or paused before the sync has completed.

Was it designed this way on purpose? Are there known cases where I should *expect* the callback not to be called?

manuroe:
> Was it designed this way on purpose? 
Not at all. Thanks for pointing it out.
There was a lot of changes in this method and we have probably missed some error cases. We should fix all of these holes.

manuroe:
@aapierce0: for reference, there is no issue in this case on Riot because the app listens to the MXSession state change.

ronnie-codes:
This is still happening on `0.26.12`. I don't receive a state change notification either. Client tries to start a session with a token that has been invalidated, and I get the following log but no failure callback. Any ideas?
```
[MXHTTPClient] Error JSONResponse: {
    errcode = "M_UNKNOWN_TOKEN";
    error = "Invalid access token passed.";
    "soft_logout" = 0;
}
```

PR #349, opened 2017-08-08
MXRoomState: Improve algorithm to manage room members displaynames disambiguation

manuroe:
This improves the code in 2 ways:
 - speed: the loop in the memberName method took 4ms per member in a room like MatrixHQ. As there is a lot of membership changes in this room, there was very often a reset on the cache, which made it a bit useless.
 - more robust against race condition: MatrixKit calls the memberName method on one of its processing queue. The new implementation of memberName does only read operations on ivars which is less risky than the write to `membersNamesCache` it did before. It should fix https://github.com/matrix-org/riot-ios-rageshakes/issues/132 (App crashed I don't know why, suspect memory issues / Crash in [MXRoomState copyWithZone:])

### DIFF
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -656,8 +656,7 @@
     {
         return nil;
     }
-    
-    // First, lookup in the cache
+
     NSString *displayName;
 
     // Get the user display name from the member list of the room
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -58,11 +58,11 @@
     NSInteger maxPowerLevel;
 
     /**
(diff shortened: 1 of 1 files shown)


MatrixSDK/Data/MXRoomState.m
```diff
@@ -614,44 +652,35 @@ - (MXRoomThirdPartyInvite *)thirdPartyInviteWithToken:(NSString *)thirdPartyInvi
 - (NSString*)memberName:(NSString*)userId
 {
     // Sanity check (ignore the request if the room state is not initialized yet)
-    if (!userId.length || !membersNamesCache)
+    if (!userId.length || !membersNamesInUse)
     {
         return nil;
     }
     
     // First, lookup in the cache
```
giomfo: please remove this comment

PR #352, opened 2017-08-18
Fix retain cycle between room and eventTimeLine

samuel-gallet:


manuroe:
All right. @samuel-gallet, please sign off.

samuel-gallet:
Samuel Gallet - samuel.gallet@applidium.com

PR #353, opened 2017-08-20
Fix problem with regex in MXPushRuleEventMatchConditionChecker

morozkin:
Now `event_match` conditions don't work if userID or roomID is used as a `pattern` because regex uses word boundary `\b`, so everything that won't be started with a word character will fail match.


giomfo:
@morozkin I confirm we have a problem with this regex. But your suggestion will only fix 2 use cases, and not the problem itself.
We faced a similar issue in MXPushRuleDisplayNameCondtionChecker (https://github.com/matrix-org/matrix-ios-sdk/pull/348).
Don't you think we should replace [NSString stringWithFormat:@"\\b%@\\b", res] with
[NSString stringWithFormat:@"(^|\\W)%@($|\\W)", res]? (without removing any character from `glob`)

morozkin:
@giomfo Yes, you are quite right. This solution will be better

PR #354, opened 2017-08-20
Added an option to send a file and keep it's filename

aramsargsyan:
@giomfo I need this for the keys file to be sent preserving the filename


giomfo:
https://github.com/vector-im/riot-ios/issues/1432

PR #378, opened 2017-10-02
Preserve message order

manuroe:
https://github.com/vector-im/riot-ios/issues/307

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -33,12 +33,41 @@ NSString *const kMXRoomInitialSyncNotification = @"kMXRoomInitialSyncNotificatio
 
 @interface MXRoomOperation : NSObject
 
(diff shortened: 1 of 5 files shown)


MatrixSDK/Data/MXRoom.m
```diff
@@ -1311,6 +1382,132 @@ - (MXHTTPOperation *)reportEvent:(NSString *)eventId
 }
 
 
+#pragma mark - Message order preserving
+/**
+ Make sure that `block` will be called in the order expected by the end user.
+
+ @param localEvent the local echo event corresping to the event being sent.
```
giomfo: - typo "corresping"
- What do you think to consider only the local event id here?

manuroe: > What do you think to consider only the local event id here?
Yeah, possibly. ```localEvent``` could be interesting in the future, or not. Anyway, given it is a private method, we will be free to change it in the future.

PR #408, opened 2017-11-10
Add MXIncomingRoomKeyRequestManager to manage room key requests

manuroe:
This is the equivalent to the existing MXOutgoingRoomKeyRequestManager but for the other side.
Note that js does not have this class. Such management is made outside the sdk, in the react-sdk.

### DIFF
--- a/MatrixSDK/Crypto/KeySharing/MXIncomingRoomKeyRequestManager.m
+++ b/MatrixSDK/Crypto/KeySharing/MXIncomingRoomKeyRequestManager.m
@@ -179,9 +179,8 @@
     [self addPendingKeyRequest:req];
 
     // Broadcast the room key request
-        __weak typeof(self) weakSelf = self;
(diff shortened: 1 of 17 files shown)


MatrixSDK/Crypto/KeySharing/MXIncomingRoomKeyRequestManager.m
```diff
@@ -0,0 +1,288 @@
...
+        return;
+    }
+
+    // Add it to pending key requests
+    [self addPendingKeyRequest:req];
+
+    // Broadcast the room key request
+        __weak typeof(self) weakSelf = self;
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (weakSelf)
```
giomfo: Why do you add here a sanity check on self?

manuroe: you are right. It would be better to make it on `crypto`.

Issue #446, opened 2018-01-24
Medias not loading with an optional client certificate


r2d2leboss:
Hi all,

I'm currently using Matrix behind Nginx. For "security", I use an optional client certificate authentication. I know Riot iOS is not supporting the load of an SSL client certificate to authenticate currently. However, as the option is optional (`ssl_verify_client` in Nginx), then Nginx server checks the IP. If it's fine, I can access my Matrix server. Currently, even if my IP is fine all medias (images, profile pictures..) are spinning or not showing in the discussion. I'm not an iOS developer but my guess is that there is an issue in the `MXMediaLoader` and more precisely in the `willSendRequestForAuthenticationChallenge`, surely causing the issue. Currently, it always go to `kSecTrustResultUnspecified` case. I *guess* the SDK should ignore the request if it can find a client certificate to send to the server and continue without it.

I'm not able to fix the issue, lacking of iOS dev skills but I'm free to test patches.

Edit : removing `willSendRequestForAuthenticationChallenge` delegate solves the issue but I guess it is not the right solution

Pierre

manuroe:
Hi @r2d2leboss,
As you seem to be able to build the app, what happens if you comment the `kSecTrustResultUnspecified` case line at https://github.com/matrix-org/matrix-ios-sdk/blob/58dc1b3075fea9a9a9b0a3bea1d656e3ab9d162d/MatrixSDK/Utils/Media/MXMediaLoader.m#L274 so that we fall down to the `default` case?

r2d2leboss:
Hi @manuroe,

Unfortunately no. With the current version of Riot on iOS, I've a spin over room pictures. With the commented line, they are replaced with the letter (default picture) and no more spin. In the log, I've : 
> [MXMediaLoader] Certificate check failed for <NSURLProtectionSpace: 0x1c4019ec0>: Host:matrix.*****, Server:https, Auth-Scheme:NSURLAuthenticationMethodServerTrust, Realm:(null), Port:443, Proxy:NO, Proxy-Type:(null)

I don't know how it should be handled properly but this article looks interesting : https://chika-kasymov.github.io/authentication_with_client_private_certificate_using_afnetworking
I guess if there is no client certificate to send, the SDK should just ignore it and send nothing. Safari and Chromes manage this opening a "window" with a certificate list if you want to send one.

PR #442, opened 2018-01-11
GZIP 1.2.1

nrakochy:
Bumped `SwiftMatrixSDK.podspec` dependency to `GZIP 1.2.1`. 

Signed-off-by: Nick Rakochy <nick.rakochy@gmail.com>

PR #447, opened 2018-01-25
MXMediaLoader: handles client certificate case

r2d2leboss:
Hi,

This PR handles client certificate case performing the default handling. It also cancels the challenge if the case is not managed. Indeed, according to https://developer.apple.com/documentation/foundation/nsurlconnectiondelegate/1414078-connection?language=objc : 
> In this method, you must invoke one of the challenge-responder methods.

It was not managed before.

It is related to issue #446. It is also quoted in https://github.com/vector-im/riot-ios/issues/1634

Next step would be to support the import of certificate to answer properly to client certificate challenge (https://developer.apple.com/library/content/qa/qa1745/_index.html / https://chika-kasymov.github.io/authentication_with_client_private_certificate_using_afnetworking).

Thanks

### DIFF
--- a/MatrixSDK/Utils/Media/MXMediaLoader.m
+++ b/MatrixSDK/Utils/Media/MXMediaLoader.m
@@ -306,6 +306,14 @@ NSString *const kMXMediaUploadIdPrefix = @"upload-";
             }
         }
     }
+    else if ([protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodClientCertificate])
+    {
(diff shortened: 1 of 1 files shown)


manuroe:
Perfect. We just need a sign-off like in https://github.com/matrix-org/matrix-ios-sdk/pull/442#issue-287662899 as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off.

You can add your sign-off to this PR in new comment.

r2d2leboss:
Signed-off-by: Pierre Dubois pierre.dubois8989@gmail.com

Done @manuroe

PR #456, opened 2018-02-23
API break: MXCrypto: Remove deviceWithDeviceId and devicesForUser methods

manuroe:
because they return local values that may be out of sync.

The downloadKeys method is preferred for better accuracy.
Using downloadKeys will fix https://github.com/vector-im/riot-ios/issues/1782.

### DIFF
--- a/MatrixSDK/Crypto/MXCrypto.h
+++ b/MatrixSDK/Crypto/MXCrypto.h
@@ -194,23 +194,6 @@ FOUNDATION_EXPORT NSString *const kMXCryptoRoomKeyRequestCancellationNotificatio
 - (MXDeviceInfo *)eventDeviceInfo:(MXEvent*)event;
 
 /**
- Get the stored device keys for a user's device.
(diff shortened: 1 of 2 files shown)


MatrixSDK/Crypto/MXCrypto.h
```diff
@@ -235,18 +218,24 @@ FOUNDATION_EXPORT NSString *const kMXCryptoRoomKeyRequestCancellationNotificatio
                complete:(void (^)(void))complete;
 
 /**
- Download the device keys for a list of users and stores them into the crypto store.
+ Get the device keys for a list of users.
+
+ Keys will be downloaded from the matrix homeserver and stored into the crypto store
```
giomfo: This happens only when forceDownload is YES?

manuroe: no. It also happens "if the information in the store is not up-to-date."

PR #468, opened 2018-03-08
MXFileStore: BF: The MXFileStore metadata can be not stored in files.

manuroe:
Part of the https://github.com/vector-im/riot-ios/issues/1797 epic.

When there is no eventStremToken in this MXFileStore metadata, the MXFileStore is reset on the next start, requiring a full initial sync :/

This happened when there are more 2 pending commits. the dispatch saving the copy of the MXFileStoreMetaData data was skipped and the copied data was lost.
As other component in MXFileStore, do not do copy but reuse the same instance.

Found with the checkRoomAccountDataTags test.

### DIFF
--- a/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
+++ b/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
@@ -1258,11 +1258,12 @@ static NSString *const kMXFileStoreRoomReadReceiptsFile = @"readReceipts";
                 [[NSFileManager defaultManager] moveItemAtPath:file toPath:backupFile error:nil];
             }
 
-            // Store new data
-            [NSKeyedArchiver archiveRootObject:metaData toFile:file];
-            
             // Make sure the data will be backed up with the right events stream token from here.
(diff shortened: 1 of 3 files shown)


MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -1262,10 +1259,10 @@ - (void)saveMetaData
             }
 
             // Store new data
-            [NSKeyedArchiver archiveRootObject:metaData2 toFile:file];
+            [NSKeyedArchiver archiveRootObject:metaData toFile:file];
             
             // Make sure the data will be backed up with the right events stream token from here.
-            backupEventStreamToken = metaData2.eventStreamToken;
+            backupEventStreamToken = metaData.eventStreamToken;
```
giomfo: I would prefer to store the `metaData.eventStreamToken` in `backupEventStreamToken` before archiving the `metaData`, just in case the metaData.eventStreamToken is updated on the main thread during the store saving.

PR #469, opened 2018-03-09
MXRestClient: Add [MXRestClient eventWithEventId:] and [MXRestClient …

manuroe:
…eventWithEventId:inRoom:] methods

and use the last one to improve the handling of the redaction of an event so old that we do not have it the local store.

https://github.com/vector-im/riot-ios/issues/1730


MatrixSDK/Data/MXEventTimeline.m
```diff
@@ -663,8 +663,8 @@ - (void)handleRedaction:(MXEvent*)redactionEvent
     {
         // Use a /context request to check whether the redacted event is a state event or not.
```
giomfo: The comment should be updated here

PR #476, opened 2018-03-27
BF: Room summary is not updated after redaction of the room display name

manuroe:
https://github.com/vector-im/riot-ios/issues/1822


MatrixSDK/Data/MXRoomSummary.m
```diff
@@ -314,6 +324,19 @@ - (void)eventDidChangeSentState:(NSNotification *)notif
     }
 }
 
+- (void)roomDidFlushData:(NSNotification *)notif
+{
+    MXRoom *room = notif.object;
+    if (_mxSession == room.mxSession && [_roomId isEqualToString:room.state.roomId])
+    {
+        NSLog(@"[MXRoomSummary] roomDidFlushData: %@", _roomId);
```
giomfo: Why don't you use the existing `resetRoomStateData` method?

manuroe: Idk. It may be better.

PR #508, opened 2018-06-07
MXRestClient: Factorise processing and completion blocks handling

manuroe:
vector-im/riot-ios#1898

and kill all "Block implicitly retains self" warnings.


manuroe:
> SGTM but we should also add weak self / strong self on success and failure blocks of httpClient calls.

True. There is only 82 of them to add...

PR #527, opened 2018-07-05
 MXRestClient: Add Matrix filter API

manuroe:
Required for https://github.com/vector-im/riot-ios/issues/1931


manuroe:
This PR has also a test that validates all new MXFilter* classes.

PR #532, opened 2018-07-17
Reduce dependency to room.state

manuroe:
Its access is going to be async.

https://github.com/vector-im/riot-ios/issues/1931

### DIFF
--- a/MatrixSDK/Data/MXRoomMembersCount.m
+++ b/MatrixSDK/Data/MXRoomMembersCount.m
@@ -35,10 +35,11 @@
     return isEqual;
 }
 
+
 #pragma mark - NSCopying
 - (id)copyWithZone:(NSZone *)zone
 {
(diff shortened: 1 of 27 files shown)


MatrixSDK/Data/MXRoomMembersCount.m
```diff
@@ -0,0 +1,71 @@
...
+        && _invited == other.invited;
+    }
+
+    return isEqual;
+}
+
+#pragma mark - NSCopying
+- (id)copyWithZone:(NSZone *)zone
+{
+    MXRoomMembersCount *roomMembersCount = [MXRoomMembersCount new];
```
SBiOSoftWhare: Do not forget to take the `NSZone` into account. Use`MXRoomMembersCount *roomMembersCount = [[MXRoomMembersCount allocWithZone:zone] init];` or even `MXRoomMembersCount *roomMembersCount = [[[self class] allocWithZone:zone] init];`

PR #533, opened 2018-07-23
MXRoom: Make access to liveTimeline data async

manuroe:
Part of https://github.com/vector-im/riot-ios/issues/1931 in order to decouple MXRoom.members from MXRoom.state

### DIFF
--- a/MatrixSDK/Data/MXPeekingRoom.m
+++ b/MatrixSDK/Data/MXPeekingRoom.m
@@ -61,11 +61,13 @@
         }
         self->httpOperation = nil;
 
-        [self.liveTimeline initialiseState:roomInitialSync.state];
+        [self liveTimeline:^(MXEventTimeline *liveTimeline) {
+            [liveTimeline initialiseState:roomInitialSync.state];
 
-        // @TODO: start the events stream
+            // @TODO: start the events stream
 
-        onServerSyncDone();
+            onServerSyncDone();
+        }];
 
     } failure:failure];
 }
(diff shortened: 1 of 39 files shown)


MatrixSDK/Data/MXPeekingRoom.m
```diff
@@ -78,12 +80,11 @@ - (MXRoomSummary *)summary
 
 - (void)close
 {
+    [super close];
```
SBiOSoftWhare: It is better to call [super close] at the end. In general clean current object resources first and then clean parent resources.

MatrixSDKTests/MXRoomStateDynamicTests.m
```diff
@@ -203,73 +205,76 @@ - (void)testLiveEventsForScenario1
...
-                        XCTAssertNil(roomState.topic, @"The room topic is not yet defined. Found: %@", roomState.topic);
-                        XCTAssertNil(room.state.topic, @"The room topic is not yet defined. Found: %@", roomState.topic);
-                        break;
-                        
-                    case 1:
-                        //  3 - Bob changes the room topic to "Topic #1"
-                        XCTAssertEqual(event.eventType, MXEventTypeRoomTopic);
+                    // Check each expected event and their roomState contect
+                    // Events are live. Then comes in order
+                    switch (eventCount++) {
```
SBiOSoftWhare: Maybe more readable to switch on `eventCount` and then call `eventCount++` after switching

manuroe: I guess that depends on your C background :)

PR #539, opened 2018-08-02
Lazy loading server side

manuroe:
Manage LL server side and the new summary API.

https://github.com/vector-im/riot-ios/issues/1931

### DIFF
--- a/MatrixSDK/Data/MXEventTimeline.m
+++ b/MatrixSDK/Data/MXEventTimeline.m
@@ -104,15 +104,6 @@ NSString *const kMXRoomInviteStateEventIdPrefix = @"invite-";
         _state = [[MXRoomState alloc] initWithRoomId:room.roomId andMatrixSession:room.mxSession andDirection:YES];
         
         _roomEventFilter = [[MXRoomEventFilter alloc] init];
-
-        // If the event stream runs with lazy loading, the timeline must do the same
-        // @TODO: This should be synchronous here but :/
-        [room.mxSession filterWithFilterId:room.mxSession.syncFilterId success:^(MXFilterJSONModel *filter) {
-            if (filter.room.state.lazyLoadMembers)
-            {
-                 self.roomEventFilter.lazyLoadMembers = YES;
-            }
-        } failure:nil];
     }
     return self;
 }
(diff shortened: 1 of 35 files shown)


MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -148,4 +161,108 @@ - (BOOL)session:(MXSession *)session updateRoomSummary:(MXRoomSummary *)summary
...
+                                       [memberNames componentsJoinedByString:@", "],
+                                       lastMemberName];
+            }
+            else
+            {
+                NSUInteger otherCount = summary.membersCount.members - 1 - serverRoomSummary.heroes.count;
+                summary.displayname = [NSString stringWithFormat:@"%@ & %@ %@",
+                                       [memberNames componentsJoinedByString:@", "],
+                                       @(otherCount),
+                                       (1 < otherCount) ? @"others" : @"other"];
```
giomfo: How to localize "others"/"other" string? Maybe you should add TODO here to track this point

MatrixSDK/MXSession.m
```diff
@@ -428,6 +429,20 @@ - (void)startWithSyncFilterId:(NSString *)syncFilterId onServerSyncDone:(void (^
         _store.syncFilterId = syncFilterId;
```
giomfo: I think we should flush the store when the filter id changes

MatrixSDK/MXSession.m
```diff
@@ -417,6 +417,7 @@ - (void)startWithSyncFilterId:(NSString *)syncFilterId onServerSyncDone:(void (^
 
     [self setState:MXSessionStateSyncInProgress];
 
+    // Check update of the filter used for /sync requests
```
giomfo: You should check first if `syncFilterId` and `_store.syncFilterId` are defined

MatrixSDK/Data/MXRoomSummaryUpdater.m
manuroe: Yeap. This could be done in the kit or here with stg like `MXSendReplyEventStringsLocalizable`.

MatrixSDK/MXSession.m
manuroe: Ftm, this is the role of the app. Some filter changes may be applied at runtime. We just log a warning.

MatrixSDK/Data/MXEventTimeline.m
```diff
@@ -377,6 +389,13 @@ - (void)handleJoinedRoomSync:(MXRoomSync *)roomSync
         [store deleteRoom:self.state.roomId];
     }
 
+    // In case of lazy-loading, we may not have the membership event for our user.
```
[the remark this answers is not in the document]
manuroe: There is now a test for that: `testRoomAfterLeavingFromAnotherDevice`.

PR #540, opened 2018-08-03
Enable lazy-loading for timelines on permalinks

manuroe:
That does not work yet as expected on server side but the client is ready

I have also added one test on search.

Still part of https://github.com/vector-im/riot-ios/issues/1931.

### DIFF
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -2357,7 +2357,7 @@ MXAuthAction;
                                           {
                                               NSLog(@"[MXRestClient] eventWithEventId: The homeserver does not support `/rooms/{roomId}/event/{eventId}` API. Try with `/context`");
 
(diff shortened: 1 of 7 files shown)


MatrixSDK/MXRestClient.m
```diff
@@ -2378,17 +2378,24 @@ - (MXHTTPOperation*)eventWithEventId:(NSString*)eventId
...
                            failure:(void (^)(NSError *error))failure
 {
     NSString *path = [NSString stringWithFormat:@"%@/rooms/%@/context/%@", apiPathPrefix, roomId, eventId];
 
+    NSMutableDictionary *parameters = [NSMutableDictionary dictionary];
+    parameters[@"limit"] = @(limit);
+
+    if (filter)
+    {
+        parameters[@"filter"] = filter.dictionary;
```
giomfo: You may use filter.jsonString here, no?

PR #550, opened 2018-08-21
MXSession: Add MXSessionStateSyncError state and MXSession.syncError 

manuroe:
to manage homeserver resource quota on /sync requests

Point 3 of of vector-im/riot-ios#1937

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1253,11 +1253,13 @@ typedef void (^MXOnResumeDone)(void);
         // on 64 bits devices, the error codes are huge integers.
         int32_t code = (int32_t)error.code;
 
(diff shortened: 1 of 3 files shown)


MatrixSDK/MXSession.m
```diff
@@ -1178,96 +1189,119 @@ - (void)serverSyncWithServerTimeout:(NSUInteger)serverTimeout
...
+    // Check whether the caller wants to handle error himself
+    if (failure)
+    {
+        failure(error);
+    }
+    else
+    {
+        // Handle error here
+        // on 64 bits devices, the error codes are huge integers.
+        int32_t code = (int32_t)error.code;
```
SBiOSoftWhare: Check error domain before check error code. You could have an error from other domain with same error code.

PR #578, opened 2018-10-15
Disable CallKit in China (#1941)

aaronraimist:
Fixes https://github.com/vector-im/riot-ios/issues/1941
Fixes https://github.com/vector-im/riot-ios/issues/1891

I decided to just disable in the same way that we disable it if you are on iOS 9.

I think it would be pointless to display the grayed out CallKit settings to Chinese users and add a new string that says basically sorry we can't give you this feature.

### DIFF
--- a/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
+++ b/MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
@@ -192,7 +192,14 @@ NSString * const kMXCallKitAdapterAudioSessionDidActive = @"kMXCallKitAdapterAud
 
 + (BOOL)callKitAvailable
 {
-    return [NSProcessInfo.processInfo isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){10,0,0}];
+	if (@available(iOS 10.0, *)) {
+		// CallKit currently illegal in China
+		// https://github.com/vector-im/riot-ios/issues/1941
+
+		return ![NSLocale.currentLocale.countryCode isEqual: @"CN"];
+	}
+
+	return false;
 }
 
 #pragma mark - CXProviderDelegate


MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -192,7 +192,14 @@ - (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date
 
 + (BOOL)callKitAvailable
 {
-    return [NSProcessInfo.processInfo isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){10,0,0}];
+	if (@available(iOS 10.0, *)) {
+		// CallKit currently illegal in China
+		// https://github.com/vector-im/riot-ios/issues/1941
+
+		return ![NSLocale.currentLocale.countryCode isEqual: @"CN"];
```
manuroe: https://stackoverflow.com/a/51319158 checks also "CHN". We could do the same.

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
```diff
@@ -192,7 +192,14 @@ - (void)reportCall:(MXCall *)call connectedAtDate:(nullable NSDate *)date
...
+	}
+
+	return false;
```
manuroe: Please use NO ;)

MatrixSDK/VoIP/CallKit/MXCallKitAdapter.m
aaronraimist: screen shot 2018-10-16 at 11 04 38 am

It seems like they are all two letters as least as far as iOS is concerned.

aaronraimist:
Should be good to go

manuroe:
> I think it would be pointless to display the grayed out CallKit settings to Chinese users and add a new string that says basically sorry we can't legally give you this feature.

If `[MXCallKitAdapter callKitAvailable] == NO`, this section is not displayed in the settings so we are good.

aaronraimist:
Yeah I was just responding to https://github.com/vector-im/riot-ios/issues/1941 which talked about adding a new string and disabling it in a different way so it would still show that section of the settings but explain that it can't be enabled legally.

PR #599, opened 2018-11-20
Lazy loading: Try to mitigate unexpected "Empty room"

manuroe:
https://github.com/vector-im/riot-ios/issues/2020

This will not work every time but, at least, we now log the issue when it happens.

### DIFF
--- a/MatrixSDK/Data/MXRoomSummaryUpdater.m
+++ b/MatrixSDK/Data/MXRoomSummaryUpdater.m
@@ -299,7 +299,10 @@
             for (MXRoomMember *member in otherMembers)
             {
                 NSString *memberName = [roomState.members memberName:member.userId];
-                [memberNames addObject:memberName];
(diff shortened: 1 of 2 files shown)


MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -339,6 +346,82 @@ - (BOOL)updateSummaryDisplayname:(MXRoomSummary *)summary session:(MXSession *)s
...
+
+        case 1:
+            if (memberCount == 2)
+            {
+                NSLog(@"[MXRoomSummaryUpdater] fixUnexpectedEmptyRoomDisplayname: Fixed 1");
+                displayname = memberNames[0];
+            }
+            else
+            {
+                //
```
SBiOSoftWhare: Maybe a legacy comment ?

PR #637, opened 2019-02-21
 Crypto: Fix crash in MXKeyBackup

manuroe:
Fixes https://github.com/vector-im/riot-ios/issues/2281.

Fixed by removing the dependency on MXCrypto.cryptoQueue that can become nil in some conditions.

### DIFF
--- a/MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
+++ b/MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
@@ -46,6 +46,9 @@ NSUInteger const kMXKeyBackupSendKeysMaxCount = 100;
 {
     MXCrypto *crypto;
 
+    // The queue to run background tasks
+    dispatch_queue_t cryptoQueue;
+
     // Observer to kMXKeyBackupDidStateChangeNotification when backupAllGroupSessions is progressing
     id backupAllGroupSessionsObserver;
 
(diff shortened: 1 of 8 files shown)


MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
```diff
@@ -44,7 +44,10 @@
 
 @interface MXKeyBackup ()
 {
-    MXSession *mxSession;
+    MXCrypto *crypto;
```
SBiOSoftWhare: Use a weak reference here to avoid retain cycle with `MXCrypto.keybackup`. Currently MXKeyBackup as a strong reference to MXCrypto and MXCrypto has a strong reference to MXKeyBackup.

MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1398,7 +1398,7 @@ - (instancetype)initWithMatrixSession:(MXSession*)matrixSession cryptoQueue:(dis
 
         oneTimeKeyCount = -1;
 
-        _backup = [[MXKeyBackup alloc] initWithMatrixSession:_mxSession];
+        _backup = [[MXKeyBackup alloc] initWithCrypto:self];
```
SBiOSoftWhare: As mentioned above a retain cycle is created as MXKeyBackup has a strong reference to MXCrypto and MXCrypto has a strong reference to MXKeyBackup.

PR #658, opened 2019-05-14
Aggregations relations

manuroe:
Part of https://github.com/vector-im/riot-ios/issues/2392

This PR:
- parses and stores aggregated data
- parses relates_to events 
- sends reactions
- provides reactions (but after an initial /sync)

Next PR will:
- provide updated reactions
- have a mechanism to notify updated reactions
- implement MXReactionCount.myUserHasReacted

### DIFF
--- a/MatrixSDKTests/MXEventAnnotationTests.swift
+++ b/MatrixSDKTests/MXEventAnnotationTests.swift
@@ -1,0 +1,84 @@
+/*
+ Copyright 2019 New Vector Ltd
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+import XCTest
+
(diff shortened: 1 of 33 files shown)


MatrixSDKTests/MXEventAnnotationTests.swift
```diff
@@ -0,0 +1,84 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+import XCTest
+
+import MatrixSDK
+
+class MXEventAnnotationTests: XCTestCase {
+
+    let eventJSON = [
```
SBiOSoftWhare: Set type here: `let eventJSON: [String : AnyObject]` and remove `as [String : AnyObject]`

MatrixSDKTests/MXEventAnnotationTests.swift
```diff
@@ -0,0 +1,84 @@
...
+            XCTAssertEqual(annotation.type, MXEventAnnotationReaction)
+            XCTAssertEqual(annotation.key, "👍")
+            XCTAssertEqual(annotation.count, 3)
+        }
+    }
+
+    func testJSONDictionary() {
+        let event = MXEvent(fromJSON: eventJSON)
+
+        let jsonDictionary = event?.jsonDictionary() as! [String : AnyObject]
```
SBiOSoftWhare: If `event?.jsonDictionary()` returns nil `as! [String : AnyObject]` will crash and it will not hit `XCTAssertNotNil(jsonDictionary)`. You could write `let jsonDictionary = event?.jsonDictionary() as? [String : AnyObject]` and use `if let jsonDictionary = jsonDictionary` below.

PR #733, opened 2019-10-13
MXIdentityService: failed on checkAPIVersionAvailability

giomfo:
The fallback to v1 based on 404 error code failed because the id server return 403 (forbidden)


giomfo:
Closed this should be fixed on server side

PR #741, opened 2019-10-29
MXHTTPOperation: Fix a crash in `-mutateTo:` method when operation parameter is nil

SBiOSoftWhare:


manuroe:
What is the call stack when the crash occurred?
We generally make this sanity check before calling this method.

Note this outside sanity check is boring. Every time, I c+p this pattern, I want to change it and make this parameter optional.

PR #744, opened 2019-10-30
Update JitsiMeetSDK dependency to ~> 2.3.1

SBiOSoftWhare:


manuroe:
It worths noting that we do not use 2.4.0 because Riot is not built with Swift 5.1 yet.

PR #776, opened 2020-02-05
MXAggregations: Optimise `referenceEventsForEvent` method, use cache when possible

SBiOSoftWhare:
vector-im/riot-ios/issues/2927


MatrixSDK/JSONModels/Aggregations/MXAggregationPaginatedResponse_Private.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ The `MXAggregationPaginatedResponse` extension exposes internal operations.
+ */
+@interface MXAggregationPaginatedResponse()
+
+- (instancetype)initWithOriginalEvent:(nullable MXEvent*)originalEvent
```
manuroe: `originalEvent` should be nonnull.
The fact it is null from server response is a synapse bug.

PR #801, opened 2020-03-26
Cross-Signing: Detect and expose new sign-ins

SBiOSoftWhare:
vector-im/riot-ios/issues/2918

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ Improvements:
  * MXStore: Add a method to get related events for a specific event.
  * MXPublicRoom: Add canonical alias property.
  * MXLogger: Add a parameter to indicate the number of log files.
+ * Cross-Signing: Detect and expose new sign-ins (/vector-im/riot-ios/issues/2918).
 
 Bug fix:
  * MXEventType: Fix Swift refinement.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ Improvements:
  * MXStore: Add a method to get related events for a specific event.
  * MXPublicRoom: Add canonical alias property.
  * MXLogger: Add a parameter to indicate the number of log files.
- * Cross-Signing: Detect and expose new sign-ins (/vector-im/riot-ios/issues/2918).
+ * MXDeviceList: Post `MXDeviceListDidUpdateUsersDevicesNotification` notification when users devices list are updated.
 
 Bug fix:
  * MXEventType: Fix Swift refinement.
--- a/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
+++ b/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
@@ -332,7 +332,7 @@ NSString *const MXCrossSigningErrorDomain = @"org.matrix.sdk.crosssigning";
         _crossSigningTools = [MXCrossSigningTools new];
         
         [self computeState];
-        [self registerUserDevicesUpdateNotification];
+        [self registerUsersDevicesUpdateNotification];
      }
     return self;
 }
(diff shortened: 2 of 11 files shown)


CHANGES.rst
```diff
@@ -15,6 +15,7 @@ Improvements:
  * MXStore: Add a method to get related events for a specific event.
  * MXPublicRoom: Add canonical alias property.
  * MXLogger: Add a parameter to indicate the number of log files.
+ * Cross-Signing: Detect and expose new sign-ins (/vector-im/riot-ios/issues/2918).
```
manuroe: We should mention `MXDeviceListDidUpdateUsersDevicesNotification`.
All Cross-Signing things will finish in one line: *Add support of cross-signing*.

MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
```diff
@@ -496,6 +500,39 @@ - (NSString *)makeSigningKey:(OLMPkSigning * _Nullable *)signing privateKey:(NSD
     return pubKey;
 }
 
+- (void)registerUserDevicesUpdateNotification
+{
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(userDevicesDidUpdate:) name:MXDeviceListDidUpdateUsersDevicesNotification object:nil];
```
manuroe: We need to listen to the right object. See my remark about `postNotificationName:MXDeviceListDidUpdateUsersDevicesNotification` below.

MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
```diff
@@ -496,6 +500,39 @@ - (NSString *)makeSigningKey:(OLMPkSigning * _Nullable *)signing privateKey:(NSD
...
+            {
+                [newDeviceIds addObject:deviceInfo.deviceId];
+            }
+        }
+        
+        if (newDeviceIds.count)
+        {
+            NSDictionary *userInfo = @{ MXCrossSigningNotificationDeviceIdsKey: newDeviceIds };
+            
+            [[NSNotificationCenter defaultCenter] postNotificationName:MXCrossSigningMyUserDidSignInOnNewDeviceNotification object:nil userInfo:userInfo];
```
manuroe:
```suggestion
            [[NSNotificationCenter defaultCenter] postNotificationName:MXCrossSigningMyUserDidSignInOnNewDeviceNotification object:self userInfo:userInfo];
```

MatrixSDK/Crypto/Data/MXDeviceListOperationsPool.m
```diff
@@ -169,8 +171,12 @@ - (void)doKeyDownloadForUsers:(NSArray<NSString *> *)users token:(NSString *)tok
                 // Update the store
                 // Note that devices which aren't in the response will be removed from the store
                 [self->crypto.store storeDevicesForUser:userId devices:mutabledevices];
+                
+                usersDevices[userId] = mutabledevices.allValues;
             }
         }
+        
+        [[NSNotificationCenter defaultCenter] postNotificationName:MXDeviceListDidUpdateUsersDevicesNotification object:nil userInfo:usersDevices];
```
manuroe: We need to post the notification from a defined instance. Else, in case of multi-account, we are going to be completely lost.
Unfortunately, we cannot use `self` because this class is private. The best candidate is probably `self->crypto`  but we need to move the notification definition to `MXCrypto`.

MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
```diff
@@ -609,7 +609,10 @@ - (MXHTTPOperation*)sendToOtherInRequest:(MXKeyVerificationRequest*)request
             break;
             
         case MXKeyVerificationTransportToDevice:
-            operation = [self sendToDevice:request.otherUser deviceId:request.otherDevice eventType:eventType content:content success:success failure:failure];
+            if (request.otherDevice)
```
manuroe: I am not confortable with this patch. I am even happy it crashes.
If we cannot communicate with the other peer, we have a serious problem. The bug seems in another place or at another level.

Can you describe the issue?

MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
```diff
@@ -944,7 +947,25 @@ - (void)handleSASKeyVerificationStart:(MXSASKeyVerificationStart*)keyVerificatio
...
+            if (existingQRCodeTransaction.state == MXQRCodeTransactionStateUnknown)
+            {
+                isExistingTransactionTemporary = YES;
+                
+                // Remove fake QR code transaction
+                [self removeQRCodeTransactionWithTransactionId:existingQRCodeTransaction.transactionId];
+            }
+        }
+        
+        if (existingQRCodeTransaction && !(isExistingTransactionTemporary))
```
manuroe: This code does not cancel anymore if there is already an existing SAS transaction.

manuroe: why not:
```suggestion
        if ([existingTransaction isKindOfClass:MXQRCodeTransaction.class])
        {
             MXQRCodeTransaction *existingQRCodeTransaction = (MXQRCodeTransaction*)existingTransaction;

            if (existingQRCodeTransaction.state == MXQRCodeTransactionStateUnknown)
            {
                existingTransaction = nil;

                // Remove fake QR code transaction
                [self removeQRCodeTransactionWithTransactionId:existingQRCodeTransaction.transactionId];
            }
        }

        if (existingTransaction)
```

PR #814, opened 2020-04-08
Fix remote ice candidates being added before remote description is setup

Legi429:
The peer connection requires the remote description to be setup before any remote ice candidates are added to it.

In the current implementation, ice candidates are sent before the called device has responded with an answer. Therefore there can be a case where the called device handles incoming ice candidates before it handles the initial offer from the caller.

This often happens when the called device is not having the app open **and** is not in the same network (e.g. one device using wifi, one using cellular), leaving the call to be stuck in "connecting" until the call is either ended or times out.

This potentially fixes multiple issues in vector-im/riot-ios. I think this is the initial one describing the issue best:
https://github.com/vector-im/riot-ios/issues/1784

### DIFF
--- a/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
+++ b/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
@@ -60,6 +60,11 @@
      Success block for the async `startCapturingMediaWithVideo` method.
      */
     void (^onStartCapturingMediaWithVideoSuccess)(void);
+
+    /**
+     Ice candidate cache
+     */
+    NSMutableArray<RTCIceCandidate *> *iceCandidateCache;
 }
 
 @property (nonatomic, strong) RTCVideoCapturer *videoCapturer;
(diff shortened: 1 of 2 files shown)


MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
```diff
@@ -159,22 +164,48 @@ - (void)handleRemoteCandidate:(NSDictionary<NSString *, NSObject *> *)candidate
...
+    {
+        // Cache ice candidates until remote description is set
+        if (iceCandidateCache == nil)
+        {
+            iceCandidateCache = [NSMutableArray arrayWithObject:iceCandidate];
+        }
+        else
+        {
+            [iceCandidateCache addObject:iceCandidate];
+        }
```
ismailgulek, line 179: @Legi429, here the first iceCandidate is being ignored, because of the else condition, if the cache is nil.

Legi429: I think this should still work since I'm initializing the array _with_ the first iceCandidate for that reason. Or am I missing something maybe?

ismailgulek: Sorry @Legi429, you're right. It seems I need a break.

ismailgulek:
Thanks @Legi429, it'll be available in 0.16.1.

Legi429:
Great :) Thanks for letting me know.

PR #815, opened 2020-04-08
Remove workaround in MXJingleCallAudioSessionConfigurator

Legi429:
This particular commit doesn't fix any issues to my knowledge. I just noticed this workaround and saw that it is no longer needed, since the required methods in RTCAudioSession are now available.


ismailgulek:
Thanks @Legi429, it'll be available in 0.16.1.

PR #824, opened 2020-04-16
MXCrypto: Change the threading model

manuroe:
to make [MXCrypto decryptEvent:] less blocking when the UI calls it.

This is noticeable when keys import takes 20s. This operation blocked decryptEvent: before.
This threading model change is possible because we rely on the thread safety of the Realm DB.

Fixes https://github.com/vector-im/riot-ios/issues/3105.

### DIFF
--- a/MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmDecryption.m
+++ b/MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmDecryption.m
@@ -1,6 +1,7 @@
 /*
  Copyright 2016 OpenMarket Ltd
  Copyright 2017 Vector Creations Ltd
+ Copyright 2020 The Matrix.org Foundation C.I.C
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmDecryption.m
```diff
@@ -114,7 +115,10 @@ - (MXEventDecryptionResult *)decryptEvent:(MXEvent*)event inTimeline:(NSString*)
         }
         else if ([olmError.domain isEqualToString:MXDecryptingErrorDomain] && olmError.code == MXDecryptingErrorUnknownInboundSessionIdCode)
         {
-            [self addEventToPendingList:event inTimeline:timeline];
+            // Do nothing more on the calling thread
+            dispatch_async(crypto.cryptoQueue, ^{
+                [self addEventToPendingList:event inTimeline:timeline];
+            });
```
ismailgulek, line 121: There is another call to addEventToPendingList in line 105. It can be a good idea to move it to cryptoQueue also. Also asserting thread in addEventToPendingList method would make us sure.

PR #827, opened 2020-04-16
BF: Cross-sign is broken from the 3rd generation

manuroe:
Fixes https://github.com/vector-im/riot-ios/issues/3112.

I tested it only with SAS. This will worth to test it with QR code when it will be more stable.

This is done by trusting locally our own user. Thus, we trust their MSK.
Commits can be checked one by one.

Screenshot 2020-04-16 at 16 40 55

### DIFF
--- a/MatrixSDK/Crypto/MXCrypto.m
+++ b/MatrixSDK/Crypto/MXCrypto.m
@@ -41,6 +41,7 @@
 
 #import "MXKeyVerificationManager_Private.h"
 #import "MXDeviceInfo_Private.h"
+#import "MXCrossSigningInfo_Private.h"
 #import "MXCrossSigning_Private.h"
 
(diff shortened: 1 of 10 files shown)


MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -933,6 +934,100 @@ - (void)setDevicesKnown:(MXUsersDevicesMap<MXDeviceInfo *> *)devices complete:(v
 #endif
 }
 
+- (void)setUserVerification:(BOOL)verificationStatus forUser:(NSString*)userId
```
SBiOSoftWhare, line 937: We can maybe use `- (void)setUserWithId:(NSString*)userId verified:(BOOL)isVerified` for readability.

PR #850, opened 2020-05-15
Fixed cancel reason response for verification request.

tladesignz:
Without this fix, the initiator of a verification request will always see a `MXTransactionCancelCode.user` reason for the cancel.

Not sure if reason should be set back on failure. Feel free to improve!

### Pull Request Checklist

Signed-off-by: Benjamin Erhart <berhart@netzarchitekten.com>


pixlwave (commented):
Thanks for this contribution @tladesignz this looks like a solid improvement to me. There appear to be a few unrelated changes included in the PR however. I think the only files we need would be `MXKeyVerificationManager.m` and `MXKeyVerificationRequest.m`? If that's correct would you mind updating for us.

Also, if you could switch from the old CHANGES.md to our new [changelog](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog) entries, that would be fab 👍

PR #858, opened 2020-05-28
MXCrypto: Only create one olm session at a time per device

manuroe:
Fix vector-im/riot-ios/issues/2331

### DIFF
--- a/MatrixSDK/Crypto/MXCrypto.m
+++ b/MatrixSDK/Crypto/MXCrypto.m
@@ -2013,6 +2013,7 @@ NSTimeInterval kMXCryptoMinForceSessionPeriod = 3600.0; // one hour
             [ensureOlmSessionsInProgress addObject:deviceIdentityKey];
         }
         
+        // In both case, we need to wait for the creation of the olm session for this device
         [devicesInProgress addObject:deviceIdentityKey];
     }
     
@@ -2026,7 +2027,7 @@ NSTimeInterval kMXCryptoMinForceSessionPeriod = 3600.0; // one hour
     observer = [[NSNotificationCenter defaultCenter] addObserverForName:kMXCryptoOneTimeKeyClaimCompleteNotification object:self queue:nil usingBlock:^(NSNotification * _Nonnull note) {
         MXStrongifyAndReturnIfNil(self);
         
-        NSMutableArray<NSString*> *devices = note.userInfo[kMXCryptoOneTimeKeyClaimCompleteNotificationDevicesKey];
+        NSArray<NSString*> *devices = note.userInfo[kMXCryptoOneTimeKeyClaimCompleteNotificationDevicesKey];
         NSError *error = note.userInfo[kMXCryptoOneTimeKeyClaimCompleteNotificationErrorKey];
         
         // Was it a /claim request for us?
--- a/MatrixSDK/Crypto/MXCrypto.m
+++ b/MatrixSDK/Crypto/MXCrypto.m
@@ -2038,7 +2038,7 @@ NSTimeInterval kMXCryptoMinForceSessionPeriod = 3600.0; // one hour
                 NSLog(@"[MXCrypto] ensureOlmSessionsForDevices: Got a notification failure for %@ devices. Fail our current pool of %@ devices", @(devices.count), @(devicesInProgress.count));
                 
                 // Consider the failure for all requests of the current pool
-                [self->ensureOlmSessionsInProgress removeObjectsInArray:devicesInProgress];
+                [self->ensureOlmSessionsInProgress removeObjectsInArray:devices];
                 [devicesInProgress removeAllObjects];
                 
                 // The game is over for this pool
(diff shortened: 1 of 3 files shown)


MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1964,27 +1976,118 @@ - (MXHTTPOperation*)ensureOlmSessionsForDevices:(NSDictionary<NSString* /* userI
...
     MXUsersDevicesMap<NSString*> *usersDevicesToClaim = [[MXUsersDevicesMap<NSString*> alloc] init];
+    // The same but devices are listed by their identity key
+    NSMutableArray<NSString*> *devicesToClaim = [NSMutableArray array];
+    
+    // Devices (by their identity key) that are waiting for a response to /claim request
+    // That can be devices for which we are going to make a /claim request OR devices that
+    // already have a pending requests.
+    // Once we have emptied this array, we can call the success or the failure block. The
+    // operation is complete.
+    NSMutableArray<NSString*> *devicesInProgress = [NSMutableArray array];
```
giomfo, line 2000: I had some difficulties to understand the meaning of this array. I think we should rename it by removing the suffix "InProgress"
I suggest: currentDevicesPool, fullDevicesList, currentDevices...

I would like to highlight this is the full list of the devices for this request whether or not we decide to make a /claim request here or waiting for an existing one

MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1964,27 +1976,118 @@ - (MXHTTPOperation*)ensureOlmSessionsForDevices:(NSDictionary<NSString* /* userI
...
+        // Claim only if a request is not yet pending
+        if (![ensureOlmSessionsInProgress containsObject:deviceIdentityKey])
+        {
+            [usersDevicesToClaim setObject:oneTimeKeyAlgorithm forUser:device.userId andDevice:device.deviceId];
+            [devicesToClaim addObject:deviceIdentityKey];
+            
+            [ensureOlmSessionsInProgress addObject:deviceIdentityKey];
+        }
+        
+        [devicesInProgress addObject:deviceIdentityKey];
```
giomfo, line 2016: I would move this line `[devicesInProgress addObject:deviceIdentityKey];` at the beginning of the block instead of at the end (to highlight that I don't need to know whether a request is in progress or not)

MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1964,27 +1976,118 @@ - (MXHTTPOperation*)ensureOlmSessionsForDevices:(NSDictionary<NSString* /* userI
...
+        
+        // Was it a /claim request for us?
+        if ([devicesInProgress mx_intersectArray:devices])
+        {
+            if (error)
+            {
+                NSLog(@"[MXCrypto] ensureOlmSessionsForDevices: Got a notification failure for %@ devices. Fail our current pool of %@ devices", @(devices.count), @(devicesInProgress.count));
+                
+                // Consider the failure for all requests of the current pool
+                [self->ensureOlmSessionsInProgress removeObjectsInArray:devicesInProgress];
```
giomfo, line 2040: I think you are wrong here, you have to remove objects in `devices` array.

Indeed the error concerns only these devices. If I observe here an error about a device for which I did not trigger /claim request, your code will remove the devices for which I triggered a request. Let my request finish its job and perhaps work for other requests

MatrixSDK/Crypto/MXCrypto.m
manuroe: well. This is a local variable. Its scope is thus well defined. The IDE colorises it accordingly.
I cannot find a better name for the moment.
 I am sticked on the fact that there are devices we are processing for this call. There will be a progression of the processing.Their number will decrease.

MatrixSDK/Crypto/MXCrypto.m
manuroe: > I think you are wrong here, you have to remove objects in devices array.

I updated `devices`. It is clearly immutable. You cannot remove things from it.

> If I observe here an error about a device for which I did not trigger /claim request, your code will remove the devices for which I triggered a request.

We check `/ Was it a /claim request for us?` before going here. If the device is in `devicesInProgress`, a /claim request was triggered for it. The request could have been made by the current call of `ensureOlmSessionsForDevices` or a previous one. We do not take care and we do not need to.

> Let my request finish its job and perhaps work for other requests

We fail for all devices in `devicesInProgress ` because we cannot have a grey answer or half success.
If `ensureOlmSessionsForDevices` is called 2 times in parallel with `[DeviceA]` then `[DeviceA, DeviceB, DeviceC]`. If /claim for DeviceA fails, we need to fail both calls even if /claim worked for DeviceB & DeviceC.

MatrixSDK/Crypto/MXCrypto.m
giomfo: I understood its usage - I'm still thinking a rename would be useful:
wdyt of "pendingDevicesList"?
If you disagree that's ok, this was just a suggestion

PR #871, opened 2020-06-25
MXCrossSigning: Gossip the master key

manuroe:
fixes vector-im/riot-ios/issues/3346

### DIFF
--- a/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
+++ b/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
@@ -48,6 +48,15 @@ NSString *const MXCrossSigningErrorDomain = @"org.matrix.sdk.crosssigning";
     return (_state >= MXCrossSigningStateTrustCrossSigning);
 }
 
+- (BOOL)hasAllPrivateKeys
+{
+    id<MXCryptoStore> cryptoStore = self.crypto.store;
+    
+    return ([cryptoStore secretWithSecretId:MXSecretId.crossSigningMaster]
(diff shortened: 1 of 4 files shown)


MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
```diff
@@ -282,12 +291,51 @@ - (void)requestPrivateKeysToDeviceIds:(nullable NSArray<NSString*>*)deviceIds
...
+            // Accept the secret anyway (It should not happen)
+            isSecretValid = YES;
+        }
+        
+        NSLog(@"[MXCrossSigning] requestPrivateKeysToDeviceIds: Got MSK. isSecretValid: %@", @(isSecretValid));
+        if (isSecretValid)
+        {
+            [self.crypto.store storeSecret:secret withSecretId:MXSecretId.crossSigningMaster];
+            dispatch_group_leave(onPrivateKeysReceivedGroup);
+        }
```
SBiOSoftWhare, line 325: We are missing balance call to `dispatch_group_leave(onPrivateKeysReceivedGroup);` when `isSecretValid == NO` for each key.
From `dispatch_group_enter` [documentation](https://developer.apple.com/documentation/dispatch/1452803-dispatch_group_enter): 
> A call to this function must be balanced with a call to dispatch_group_leave.

manuroe: The current algorithm seems correct to me. If the secret is not valid, we need to wait for another device to send another secret.
This blocks return NO to keep receiving secrets.

PR #892, opened 2020-08-04
Test update

manuroe:
Commits can be reviewed separately.

Changes and explanation:
 * Fix "fastlane ios test" and generate html report
Available at build/test/report.html. Other available artefacts: MatrixSDK-macOS.xcresult and report.junit

image

 * Make tests fail instantly if no local synapse is running
In order to save CI time

 * Do not use anymore NSAssert in tests
If there is a crash when running tests that will be for a good reason. Plus it avoids xcode to rerun crashed tests 7 times.

### DIFF
--- a/MatrixSDKTests/MatrixSDKTestsData.h
+++ b/MatrixSDKTests/MatrixSDKTestsData.h
@@ -148,12 +148,18 @@ onUnrecognizedCertificateBlock:(MXHTTPClientOnUnrecognizedCertificate)onUnrecogn
 - (void)breakTestCase:(XCTestCase*)testCase reason:(NSString *)reason, ...;
 
 // Logout the user on the server and log the user in with a new device
-- (void)relogUserSession:(MXSession*)session withPassword:(NSString*)password onComplete:(void (^)(MXSession *newSession))onComplete;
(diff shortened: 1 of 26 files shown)


AliSoftware:
Just updated the Fastlane `test` lane to always `zip` the `.xcresult` bundle (in all cases, whether `scan` succeeds or fails) so that the CI can upload it as an artefact to be downloaded. [this nightly build on BuildKite for example](See https://buildkite.com/matrix-dot-org/matrix-ios-sdk/builds/100#892cfacd-4f4e-4dba-a9fd-e11640f77b0a)

MatrixSDKTests/MatrixSDKTestsData.h
```diff
@@ -45,7 +45,8 @@ FOUNDATION_EXPORT NSString * const kMXTestsAliceAvatarURL;
 
 // Get credentials asynchronously
 // The user will be created if needed
-- (void)getBobCredentials:(void (^)(void))success;
+- (void)getBobCredentials:(XCTestCase*)testCase
```
SBiOSoftWhare, line 48: We could use:
`- (void)getBobCredentialsFromTestCase:(XCTestCase*)testCase completion:(void (^)(void))completion;`

PR #941, opened 2020-11-11
Tools: Add MXProfiler to track performances

manuroe:
The default implementation, `MXBaseProfiler`, reports duration in logs and to the `MXAnalytics` interface.
Logs look like:
```
[MXBaseProfiler] Task startup - storePreload for 578 units completed in 1682.477ms
[MXBaseProfiler] Task startup - mountData for 578 units completed in 2267.157ms
[MXBaseProfiler] Task startup - launchScreen for 1 units completed in 2282.014ms
...
[MXBaseProfiler] Task startup - incrementalSync for 53 units completed in 481.103ms
[MXBaseProfiler] Task notifications - timelineDisplay for 1 units completed in 652.003ms
```
 
The app can call `MXProfiler.pause()` when it goes in background. This avoids to measure and to report bad durations.

### DIFF
--- a/MatrixSDK/Utils/Profiling/MXTaskProfile.h
+++ b/MatrixSDK/Utils/Profiling/MXTaskProfile.h
@@ -20,13 +20,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MXTaskProfile : NSObject
 
+// Task name
 @property (nonatomic, readonly) NSString *name;
+
+// Category to group related tasks
 @property (nonatomic, readonly) NSString *category;
 
+// Task timing
 @property (nonatomic, readonly) NSDate *startDate;
 @property (nonatomic, readonly, nullable) NSDate *endDate;
 @property (nonatomic, readonly) NSTimeInterval duration;
 
+// Number of items managed by the task
 @property (nonatomic) NSUInteger units;
 
 // YES, if the task was interrupted by a pause
--- a/MatrixSDK/Utils/Profiling/MXTaskProfile.h
+++ b/MatrixSDK/Utils/Profiling/MXTaskProfile.h
@@ -1,0 +1,38 @@
+// 
+// Copyright 2020 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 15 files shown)


MatrixSDK/Utils/Profiling/MXTaskProfile.h
```diff
@@ -0,0 +1,38 @@
...
+@interface MXTaskProfile : NSObject
+
+@property (nonatomic, readonly) NSString *name;
+@property (nonatomic, readonly) NSString *category;
+
+@property (nonatomic, readonly) NSDate *startDate;
+@property (nonatomic, readonly, nullable) NSDate *endDate;
+@property (nonatomic, readonly) NSTimeInterval duration;
+
+@property (nonatomic) NSUInteger units;
```
SBiOSoftWhare, line 30: Can we have more description about what this field is intended to carry?
Maybe we can add `metadata: NSDictionary` field even if Matomo does not support at the moment. We can use known dictionary keys like: `count` and let the possibility to set custom ones.

MatrixSDK/Utils/Profiling/MXTaskProfile.m
```diff
@@ -0,0 +1,55 @@
...
+}
+
+- (NSTimeInterval)duration
+{
+    NSTimeInterval duration = 0;
+    if (_endDate)
+    {
+        duration = [_endDate timeIntervalSinceDate:_startDate];
+    }
+    return duration;;
```
SBiOSoftWhare, line 52:
```suggestion
    return duration;
```

MatrixSDK/Utils/Profiling/MXTaskProfile.h
manuroe: > Maybe we can add metadata: NSDictionary field even if Matomo does not support at the moment. We can use known dictionary keys like: count and let the possibility to set custom ones.

count = units here, which is the genetic term that matches dollars, which is the metric you find in every analytics tools.

This is the plan to have a `metadata` or `context` dict but let's see first how count.ly behaves with its `context` while integrating it.

SBiOSoftWhare (approved):

PR #949, opened 2020-11-18
Room membership: Introduce room membership change notifications and getter

SBiOSoftWhare:


MatrixSDK/Room/Membership/MXMembershipTransitionState.h
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/**
+ `MXMembershipTransitionState` is an ephemeral state that represents the transitivity between the membership (MXMembership) states.
+ */
+typedef NS_ENUM(NSInteger, MXMembershipTransitionState) {
+    MXMembershipTransitionStateUnknown,
+    MXMembershipTransitionStatePending,
```
manuroe, line 22: Can we use `MXMembershipTransitionStateInvited`?

`Pending` alone is not clear enough. I would prefer to avoid a "ing" word because this is not a transitioning state.

manuroe (approved):

PR #965, opened 2020-12-03
Support multiple SSO Identity Providers (MSC2858)

SBiOSoftWhare:
Implement [MSC2858](https://github.com/matrix-org/matrix-doc/pull/2858)

Part of vector-im/element-ios/issues/3846

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,13 +5,14 @@ Changes to be released in next version
  * 
 
 🙌 Improvements
- * 
+ * Login flow: Add MXLoginSSOFlow to support multiple SSO Identity Providers ([MSC2858](https://github.com/matrix-org/matrix-doc/pull/2858)) (vector-im/element-ios/issues/3846).
 
 🐛 Bugfix
  * 
 
 ⚠️ API Changes
- * 
+ * Add MXLoginSSOFlow which represents a SSO login flow and offers multiple SSO Identity Providers through MXLoginSSOIdentityProvider class.
+ * MXAuthenticationSession: `flows` property which represents all supported login flows now handles MXLoginFlow and MXLoginSSOFlow classes.
 
 🗣 Translations
  * 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,8 +11,7 @@ Changes to be released in next version
  * 
 
 ⚠️ API Changes
- * Add MXLoginSSOFlow which represents a SSO login flow and offers multiple SSO Identity Providers through MXLoginSSOIdentityProvider class.
- * MXAuthenticationSession: `flows` property which represents all supported login flows now handles MXLoginFlow and MXLoginSSOFlow classes.
+ *
 
 🗣 Translations
  * 
--- a/MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
+++ b/MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
@@ -1,0 +1,37 @@
+// 
+// Copyright 2020 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+#import "MXJSONModels.h"
+#import "MXLoginSSOIdentityProvider.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+extern NSString *const MXLoginSSOFlowIdentityProvidersKey;
+
+/**
+ `MXLoginSSOFlow` represents a SSO login or a register flow supported by the home server (See MSC2858 https://github.com/matrix-org/matrix-doc/pull/2858).
+ */
+@interface MXLoginSSOFlow : MXLoginFlow
+
+/**
+ List of all SSO Identity Providers supported
+ */
+@property (nonatomic, readonly) NSArray<MXLoginSSOIdentityProvider*> *identityProviders;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
+++ b/MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
@@ -1,0 +1,60 @@
+// 
+// Copyright 2020 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import "MXLoginSSOFlow.h"
+
+NSString *const MXLoginSSOFlowIdentityProvidersKey = @"identity_providers";
+
+@interface MXLoginSSOFlow()
+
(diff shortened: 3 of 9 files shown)


CHANGES.rst
```diff
@@ -5,13 +5,14 @@ Changes to be released in next version
...
 🙌 Improvements
- * 
+ * Login flow: Add MXLoginSSOFlow to support multiple SSO Identity Providers ([MSC2858](https://github.com/matrix-org/matrix-doc/pull/2858)) (vector-im/element-ios/issues/3846).
 
 🐛 Bugfix
  * 
 
 ⚠️ API Changes
- * 
+ * Add MXLoginSSOFlow which represents a SSO login flow and offers multiple SSO Identity Providers through MXLoginSSOIdentityProvider class.
```
manuroe, line 14: We usually use the API changes for api breaks. 
It will be annoying if we start to describe every new api here.

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
```diff
@@ -0,0 +1,60 @@
...
+
+@implementation MXLoginSSOFlow
+
++ (id)modelFromJSON:(NSDictionary *)JSONDictionary
+{
+    MXLoginSSOFlow *loginFlow = [MXLoginSSOFlow new];
+    if (loginFlow)
+    {
+        MXJSONModelSetString(loginFlow.type, JSONDictionary[@"type"]);
+        MXJSONModelSetArray(loginFlow.stages, JSONDictionary[@"stages"]);
```
ismailgulek, line 35: Can't we delegate this parsing to the superclass? I mean with another method like `parseJSON:` in `MXLoginfFlow`. My concern is that If another property added to `MXLoginFlow`, we'll have to maintain this one also.

MatrixSDK/JSONModels/Login/MXLoginSSOIdentityProvider.m
```diff
@@ -0,0 +1,49 @@
...
+{
+    NSString *identifier;
+    NSString *name;
+    
+    MXJSONModelSetString(identifier, JSONDictionary[@"id"]);
+    MXJSONModelSetString(name, JSONDictionary[@"name"]);
+    
+    MXLoginSSOIdentityProvider *identityProvider = [MXLoginSSOIdentityProvider new];
+    
+    if (identityProvider && identifier && name)
```
ismailgulek, line 39: Have two concerns.
1. We're declaring that `identifier` and `name` will be non-nil with `NS_ASSUME_NONNULL_BEGIN`, but if they are nil after parsing, we still return them nil here.
2. This introduces some logic in a model class, which doesn't seem to be a good idea.

MatrixSDK/JSONModels/MXJSONModels.m
```diff
@@ -196,13 +197,51 @@ + (id)modelFromJSON:(NSDictionary *)JSONDictionary
...
+{
+    NSMutableArray *loginFlows;
+    
+    for (NSDictionary *JSONDictionary in JSONDictionaries)
+    {
+        MXLoginFlow *loginFlow;
+        
+        NSString *type;
+        
+        MXJSONModelSetString(type, JSONDictionary[@"type"]);
```
ismailgulek, line 220: I'm not very comfortable with exposing the name of `type` field outside of `MXLoginFlow` class. Maybe we can define an extern variable for that.

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
```diff
@@ -0,0 +1,37 @@
...
+#import "MXLoginSSOIdentityProvider.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+extern NSString *const MXLoginSSOFlowIdentityProvidersKey;
+
+/**
+ `MXLoginSSOFlow` represents a SSO login or a register flow supported by the home server (See MSC2858 https://github.com/matrix-org/matrix-doc/pull/2858).
+ */
+@interface MXLoginSSOFlow : MXLoginFlow
```
ismailgulek, line 28: I'm not very comfortable with the existence of this class and deciding that a flow is SSO or not by class check. Why don't we reuse `MXLoginFlow` for that? And maybe a method like `-(BOOL) isSSo`, which checks the type.

MatrixSDKTests/MXAuthenticationSessionTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        guard let flows = authenticationSession?.flows else {
+            XCTFail("flows shouldn not be nil")
+            return
+        }
+        
+        XCTAssertEqual(flows.count, 3)
+        
+        if let ssoFlow = flows.first(where: { $0.type == kMXLoginFlowTypeSSO }) {
+            
+            if let loginSSOFlow = ssoFlow as? MXLoginSSOFlow {
```
ismailgulek, line 72: This is what i wanted to avoid on above comment.

CHANGES.rst
ismailgulek: I agree this

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
SBiOSoftWhare: Yes he have an architecture problem. Here the class method is not really convenient. Member method like other libs would help to make calls to `super` implementation. 
In `MXLoginSSOFlow` Now we can do something like:
```
+ (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
{
    MXLoginSSOFlow *loginFlow = [super modelFromJSON:JSONDictionary];
```
And in `MXLoginFlow` we should have: 
```
+ (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
{
    MXLoginSSOFlow *loginFlow = [self new];
```

MatrixSDK/JSONModels/Login/MXLoginSSOIdentityProvider.m
SBiOSoftWhare: Yes I should return nil in this case. In the MSC identifier && name have to be non optional.

MatrixSDK/JSONModels/MXJSONModels.m
SBiOSoftWhare: Sure

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
SBiOSoftWhare: I made the same as Web implementation only expose relevant properties for dedicated flow. It's not a really good idea either to expose a lot of properties that can be nil or not and used in only one case. 
In MXLoginSSOFlow case we expose the identity providers list can't be nil. And you know that is useful only in this case.

MatrixSDKTests/MXAuthenticationSessionTests.swift
SBiOSoftWhare: He have to differentiate `m.login.sso` from `m.login.cas` they are both the same kind of flow but not exactly the same :)

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
SBiOSoftWhare: Web implementation here https://github.com/matrix-org/matrix-react-sdk/blob/8593845e3d1120bb8d9f7f9744f4d9a3a809a578/src/Login.ts#L42

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
ismailgulek: I think this seems better.

MatrixSDK/JSONModels/MXJSONModels.m
ismailgulek: This should be avoided after https://github.com/matrix-org/matrix-ios-sdk/pull/965#discussion_r536024589

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.h
ismailgulek: Ok, let's continue as is then. thanks for the explanation.

MatrixSDK/JSONModels/Login/MXLoginSSOIdentityProvider.m
ismailgulek: Cool, thanks.

MatrixSDKTests/MXAuthenticationSessionTests.swift
ismailgulek: This will continue as the same, explained in https://github.com/matrix-org/matrix-ios-sdk/pull/965#discussion_r536027445

ismailgulek (approved):

PR #968, opened 2020-12-08
MXBackgroundSyncService: Keep its stores in sync with MXSession store

manuroe:
In real life, the same `MXBackgroundSyncService` instance runs for hours.  `MXSession` can run in parallel with it.

Here, we make sure that `MXBackgroundSyncService` uses the latest data stored by `MXSession`, instead of using obsolete data we have in memory.
Else `MXSession` will consider `MXBackgroundSyncService` data obsolete on its next startup or resume, making `MXBackgroundSyncService` previous work useless.

MXBackgroundStore.event(withEventId:) was not working. It should work now.

### DIFF
--- a/MatrixSDK/Background/MXBackgroundStore.swift
+++ b/MatrixSDK/Background/MXBackgroundStore.swift
@@ -87,6 +87,13 @@ class MXBackgroundStore: NSObject, MXStore {
         }
     }
     
+    func event(withEventId eventId: String, inRoom roomId: String) -> MXEvent? {
+        // TODO
+        
+        NSLog("[MXBackgroundStore] eventWithEventId: \(eventId): TODO")
+        return nil
+    }
+    
     
     //  MARK: - Stubs
     
@@ -103,10 +110,6 @@ class MXBackgroundStore: NSObject, MXStore {
     
     func eventExists(withEventId eventId: String, inRoom roomId: String) -> Bool {
         return false
-    }
-    
-    func event(withEventId eventId: String, inRoom roomId: String) -> MXEvent? {
-        return nil
     }
     
     func deleteAllMessages(inRoom roomId: String) {
--- a/MatrixSDK/Background/MXBackgroundStore.swift
+++ b/MatrixSDK/Background/MXBackgroundStore.swift
@@ -24,57 +24,29 @@ enum MXBackgroundStoreErrorCode: Int {
     case userIDMissing = 1001   // User ID is missing in credentials
 }
 
-/// Fake memory store implementation. Uses some real values from an MXFileStore instance.
-class MXBackgroundStore: MXMemoryStore {
+/// Minimalist MXStore implementation. It uses some real values from an MXFileStore instance.
+class MXBackgroundStore: NSObject, MXStore {
 
-    //  In-memory value for eventStreamToken. Will be used as eventStreamToken if provided.
-    private var lastStoredEventStreamToken: String?
-    private var credentials: MXCredentials
     //  real store
     private var fileStore: MXFileStore
-    private var myUser: MXUser?
     
     init(withCredentials credentials: MXCredentials) {
-        self.credentials = credentials
         fileStore = MXFileStore(credentials: credentials)
         //  load real eventStreamToken
         fileStore.loadMetaData()
     }
     
-    override func open(with credentials: MXCredentials, onComplete: (() -> Void)?, failure: ((Error?) -> Void)? = nil) {
-        super.open(with: credentials, onComplete: {
-            guard let userId = credentials.userId else {
-                failure?(NSError(domain: MXBackgroundStoreErrorDomain,
-                                 code: MXBackgroundStoreErrorCode.userIDMissing.rawValue,
-                                 userInfo: nil))
-                return
-            }
-            //  load session user before calling onComplete
-            self.fileStore.asyncUsers(withUserIds: [userId], success: { (users) in
-                if let user = users.first {
-                    self.myUser = user
-                }
-                onComplete?()
-            }, failure: failure)
-        }, failure: failure)
-    }
-    
     //  Return real eventStreamToken, to be able to launch a meaningful background sync
-    override var eventStreamToken: String? {
+    var eventStreamToken: String? {
         get {
-            //  if more up-to-date token exists, use it
-            if let token = lastStoredEventStreamToken {
-                return token
-            }
             return fileStore.eventStreamToken
         } set {
-            //  store new token values in memory, and return these values in future reads
-            lastStoredEventStreamToken = newValue
+            //  no-op
         }
     }
     
     //  Return real userAccountData, to be able to use push rules
-    override var userAccountData: [AnyHashable : Any]? {
+    var userAccountData: [AnyHashable : Any]? {
         get {
             return fileStore.userAccountData
         } set {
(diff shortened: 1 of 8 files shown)


MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -144,6 +143,11 @@ public enum MXBackgroundSyncServiceError: Error {
                         completion: @escaping (MXResponse<MXEvent>) -> Void) {
         NSLog("[MXBackgroundSyncService] fetchEvent: \(eventId). allowSync: \(allowSync)")
         
+        // Check local stores on every request so that we use up-to-data data from the MXSession store
+        if allowSync {
+            updateBackgroundServiceStoresIfNeeded()
```
ismailgulek, line 148: It may be a good idea to move this in `launchBackgroundSync`, as there are other cases triggering background sync.

MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -522,4 +527,24 @@ public enum MXBackgroundSyncServiceError: Error {
...
+        // by checking that the event stream token we have in memory is the same that in the last version of MXSession store
+        let eventStreamToken = store.eventStreamToken
+ 
+        let upToDateStore = MXBackgroundStore(withCredentials: credentials)
+        let upToDateEventStreamToken = upToDateStore.eventStreamToken
+
+        if (eventStreamToken != upToDateEventStreamToken) {
+            // MXSession continued to work in parallel with the background sync service
+            // MXSession has updated its stream token. We need to use t
+            NSLog("[MXBackgroundSyncService] resetBackgroundServiceStoresIfNeeded: Update MXBackgroundStore. Its event stream token (\(String(describing: eventStreamToken))) does not match current MXStore.eventStreamToken (\(String(describing: upToDateEventStreamToken)))")
```
ismailgulek, line 541:
```suggestion
            NSLog("[MXBackgroundSyncService] updateBackgroundServiceStoresIfNeeded: Update MXBackgroundStore. Its event stream token (\(String(describing: eventStreamToken))) does not match current MXStore.eventStreamToken (\(String(describing: upToDateEventStreamToken)))")
```

MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -522,4 +527,24 @@ public enum MXBackgroundSyncServiceError: Error {
...
+            store = upToDateStore
+            
+            // syncResponseStore has obsolete data. Reset it
+            NSLog("[MXBackgroundSyncService] resetBackgroundServiceStoresIfNeeded: Reset MXSyncResponseStore. Its prevBatch was token \(String(describing: syncResponseStore.prevBatch))")
```
ismailgulek, line 545:
```suggestion
            NSLog("[MXBackgroundSyncService] updateBackgroundServiceStoresIfNeeded: Reset MXSyncResponseStore. Its prevBatch was token \(String(describing: syncResponseStore.prevBatch))")
```

MatrixSDK/Background/MXBackgroundStore.swift
```diff
@@ -83,45 +58,177 @@ class MXBackgroundStore: MXMemoryStore {
     }
     
     //  This store should act like as a permanent one
-    override var isPermanent: Bool {
+    var isPermanent: Bool {
         return true
```
ismailgulek, line 62:
```suggestion
        return false
```

I think we may stop acting like a permanent store, as we don't use a session anymore.

MatrixSDK/Background/MXBackgroundStore.swift
```diff
@@ -83,45 +58,177 @@ class MXBackgroundStore: MXMemoryStore {
     }
     
     //  This store should act like as a permanent one
```
ismailgulek, line 60:
```suggestion
```

MatrixSDK/Background/MXBackgroundStore.swift
```diff
@@ -83,45 +58,177 @@ class MXBackgroundStore: MXMemoryStore {
...
+    }
+    
+    func store(_ user: MXUser) {
+    }
+    
+    func users() -> [MXUser]? {
+        return nil
+    }
+    
+    func user(withUserId userId: String) -> MXUser? {
```
ismailgulek, line 188: We use the result of this function at `MXBackgroundSyncService.swift:132`. Why did you remove fetching our user from the store?

MatrixSDK/Background/MXBackgroundStore.swift
```diff
@@ -24,57 +24,32 @@ enum MXBackgroundStoreErrorCode: Int {
...
-                                 userInfo: nil))
-                return
-            }
-            //  load session user before calling onComplete
-            self.fileStore.asyncUsers(withUserIds: [userId], success: { (users) in
-                if let user = users.first {
-                    self.myUser = user
-                }
-                onComplete?()
-            }, failure: failure)
```
ismailgulek, line 58: We need this, commented about at `user(withUserId userId: String) -> MXUser?` method.

MatrixSDK/Background/MXBackgroundStore.swift
```diff
@@ -83,45 +58,177 @@ class MXBackgroundStore: MXMemoryStore {
...
+    }
+    
+    func storeEvent(forRoom roomId: String, event: MXEvent, direction: __MXTimelineDirection) {
+    }
+    
+    func replace(_ event: MXEvent, inRoom roomId: String) {
+    }
+    
+    func eventExists(withEventId eventId: String, inRoom roomId: String) -> Bool {
+        return false
```
ismailgulek, line 136: It'd be perfect to implement this in an efficient way in order to avoid reading whole events for only one. Not mandatory for now but it can be a good improvement point in future.

MatrixSDKTests/MXBackgroundSyncServiceTests.swift
```diff
@@ -584,6 +584,163 @@ class MXBackgroundSyncServiceTests: XCTestCase {
...
+                        self.bgSyncService = MXBackgroundSyncService(withCredentials: bobCredentials)
+                        self.bgSyncService?.event(withEventId: eventId, inRoom: roomId) { _ in
+                            
+                            let syncResponseStore = MXSyncResponseFileStore()
+                            syncResponseStore.open(withCredentials: bobCredentials)
+                            let syncResponseStorePrevBatch = syncResponseStore.prevBatch
+                            
+                            // Test check
+                            // MXSyncResponseFileStore should point to the same sync token as MXSession store
+                            XCTAssertNotEqual(bobSessionStartEventStreamToken1, syncResponseStorePrevBatch)
```
ismailgulek, line 639: This should be `XCTAssertEqual` in theory, but this is worthless. Because Bob's session is not stopped in this test, we cannot guarantee that his store's eventStreamToken will be same as sync response store's prevBatch.

ismailgulek (changes requested):

MatrixSDK/Background/MXBackgroundStore.swift
manuroe: I will check again why we need `user(withUserId userId: String) -> MXUser?`
My issue was this is an async call within a synchronous method. I saw cases where the code was doing things whereas the async call was not complete.

MatrixSDK/Background/MXBackgroundStore.swift
manuroe: I do not think we use this method in the bg service. If I am wrong, we need to implement it.

MatrixSDK/Background/MXBackgroundStore.swift
manuroe: Thanks for the catch. I missed it. I implemented `MXBackgroundSyncService.swift:132` in another way.

ismailgulek: Cool, thanks

MatrixSDK/Background/MXBackgroundStore.swift
ismailgulek: No, you right, as we're not using it right now. Mine was just a wish.

MatrixSDKTests/MXBackgroundSyncServiceTests.swift
```diff
@@ -584,6 +584,157 @@ class MXBackgroundSyncServiceTests: XCTestCase {
...
+                                    
+                                    // -> MXBackgroundSyncService should have detected that the MXSession ran in parallel.
+                                    //    It must have reset its cache. syncResponseStore.prevBatch must not be the same
+                                    let syncResponseStore = MXSyncResponseFileStore()
+                                    syncResponseStore.open(withCredentials: bobCredentials)
+                                    
+                                    XCTAssertNotEqual(syncResponseStorePrevBatch, syncResponseStore.prevBatch)
+                                    
+                                    // -> MXSyncResponseFileStore must be in sync with MXSession store
+                                    XCTAssertNotEqual(bobSessionStartEventStreamToken2, syncResponseStore.prevBatch)
```
ismailgulek, line 654: Seems same with https://github.com/matrix-org/matrix-ios-sdk/pull/968#discussion_r538401678

MatrixSDK/Background/MXBackgroundSyncService.swift
ismailgulek: > Do you think we need to call this method for every call of `launchBackgroundSync`?

Yes, just in case for parallel session syncs.

ismailgulek (approved):

PR #969, opened 2020-12-09
VoIP: Use Tracks Instead of Transceivers

ismailgulek:
Fix vector-im/element-ios#3862

### DIFF
--- a/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
+++ b/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
@@ -607,7 +607,7 @@ didRemoveIceCandidates:(NSArray<RTCIceCandidate *> *)candidates;
         if ([transceiver currentDirection:&direction])
         {
             if (direction == RTCRtpTransceiverDirectionInactive ||
-                direction == RTCRtpTransceiverDirectionSendOnly ||
+                direction == RTCRtpTransceiverDirectionRecvOnly ||  //  remote party can set a hold tone with 'sendonly'
                 direction == RTCRtpTransceiverDirectionStopped)
             {
                 return NO;
@@ -619,7 +619,7 @@ didRemoveIceCandidates:(NSArray<RTCIceCandidate *> *)candidates;
     
     if (activeReceivers.count == 0)
     {
-        //  if there is no active receivers left, we can say this call is holded
+        //  if there is no active receivers (on the other party) left, we can say this call is holded
         dispatch_async(dispatch_get_main_queue(), ^{
             [self.delegate callStackCallDidRemotelyHold:self];
         });
--- a/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
+++ b/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
@@ -27,7 +27,7 @@
 #import "MXJingleCameraCaptureController.h"
 #import <WebRTC/WebRTC.h>
 
-NSString *const kMXJingleCallWebRTCMainStreamID = @"ARDAMS";
+NSString *const kMXJingleCallWebRTCMainStreamID = @"userMedia";
 
 @interface MXJingleCallStackCall () <RTCPeerConnectionDelegate>
 {
--- a/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
+++ b/MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
@@ -27,6 +27,8 @@
 #import "MXJingleCameraCaptureController.h"
 #import <WebRTC/WebRTC.h>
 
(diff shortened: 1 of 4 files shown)


MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
```diff
@@ -434,6 +416,8 @@ - (void)peerConnection:(RTCPeerConnection *)peerConnection
 {
     NSLog(@"[MXJingleCallStackCall] didRemoveStream");
     
+    //  TODO: Check this stream has the main stream ID (kMXJingleCallWebRTCMainStreamID) after all platforms decided to use the same stream id.
```
dbkr, line 419: Not sure that the IDs should really matter, only how many are left. We don't really care about stream ID until we start dealing with multiple video or audio streams.

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
```diff
@@ -27,6 +27,8 @@
 #import "MXJingleCameraCaptureController.h"
 #import <WebRTC/WebRTC.h>
 
+NSString *const kMXJingleCallWebRTCMainStreamID = @"ARDAMS";
```
dbkr, line 30: I'm going to suggest 'userMedia' for now: anything would be better than 'ARDAMS'

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
```diff
@@ -625,36 +602,32 @@ - (void)setCameraPosition:(AVCaptureDevicePosition)theCameraPosition
...
+    NSUInteger numberOfAudioTracks = [self numberOfMatchesOfKeyword:@"m=audio" inString:sdpOffer];
+    NSUInteger numberOfVideoTracks = [self numberOfMatchesOfKeyword:@"m=video" inString:sdpOffer];
+
     if (numberOfAudioTracks == 0 && numberOfVideoTracks == 0)
     {
         //  no audio or video tracks
         return YES;
     }
+
+    NSUInteger numberOfInactiveTracks = [self numberOfMatchesOfKeyword:@"a=inactive" inString:sdpOffer];
```
dbkr, line 617: I still think we should be able to get this without sdp parsing by setting the remote description and then querying the tracks, as per https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/webrtc/call.ts#L1025 / https://github.com/matrix-org/matrix-js-sdk/blob/46cebcd1ca51a6dacd9802f9aa3593cef2015436/src/webrtc/call.ts#L690

dbkr (commented):
looks great otherwise, and simpler 🎉

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
ismailgulek: 👍

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
ismailgulek: It's not used right now, just didn't delete for possible future uses.

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
ismailgulek: Updated

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
```diff
@@ -602,7 +599,34 @@ - (void)setCameraPosition:(AVCaptureDevicePosition)theCameraPosition
...
+        }
+        
+        return YES;
+    }]];
+    
+    //  if there is no active receivers left, we can say this call is holded
+    if (activeReceivers.count == 0)
+    {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self.delegate callStackCallDidHold:self];
```
dbkr, line 624: Should there be a call to say it's un-held here if there is an active receiver? Is this relying on AddStream being called when the call is un-held?

ismailgulek: Yes, it's relying on `didAddStream` call.

dbkr: ok - as long as this is reliable

dbkr (approved):

MatrixSDKExtensions/VoIP/Jingle/MXJingleCallStackCall.m
ismailgulek: I didn't like this afterwards, as we can add more streams to calls in future, it may lead to unexpected results. Code updated.

manuroe (approved):

PR #970, opened 2020-12-10
Added MXKeyProvider to share encryption keys between application and Kit

gileluard:
part of the fix for https://github.com/vector-im/element-ios/issues/3866

### DIFF
--- a/MatrixSDK/Crypto/KeyProvider/MXKeyProvider.h
+++ b/MatrixSDK/Crypto/KeyProvider/MXKeyProvider.h
@@ -19,6 +19,8 @@
 #ifndef KeyProvider_h
 #define KeyProvider_h
 
+NS_ASSUME_NONNULL_BEGIN
+
 /// This delegate will be in charged to effectively give the encryption keys configured in the application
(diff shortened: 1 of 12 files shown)


MatrixSDK/Crypto/KeyProvider/MXKeyProvider.h
```diff
@@ -0,0 +1,55 @@
...
+
+@end
+
+@interface MXKeyProvider : NSObject
+
++ (nonnull instancetype)sharedInstance;
+
+@property (nonatomic, strong, nullable) id<MXKeyProviderDelegate> delegate;
+
+- (nullable MXKeyData *)requestKeyForDataOfType:(MXDataType)dataType;
```
manuroe, line 45: Can we use string ids instead of enum for more flexibility?

manuroe: The SDK should not need to know what keys will be needed for the the kit or the app.
That means the "mandatory" check must be done by the consumer.

manuroe (approved):

PR #971, opened 2020-12-11
Call Transfer Initiation

ismailgulek:
Fix vector-im/element-ios#3872


dbkr (commented):
This is looking good - suggest we keep it either on a branch or behind a feature flag if possible until we have a bridge or similar to test it against.

ismailgulek:
> This is looking good - suggest we keep it either on a branch or behind a feature flag if possible until we have a bridge or similar to test it against.

Thanks, disabled with https://github.com/matrix-org/matrix-ios-sdk/pull/971/commits/41d94fa6d0e188a1f8468ec962b8d0f58f7c5954

dbkr (approved):
lgtm - we will probably also want to support the simpler case of just sending a replaces event to the other side telling them what user to call.

ismailgulek:
> lgtm - we will probably also want to support the simpler case of just sending a replaces event to the other side telling them what user to call.

Merging this now. This request will be handled in future, after we decide when to switch to that flow.

PR #972, opened 2020-12-11
MXRestClient: Fix uploadKeys method that was using the /keys/upload/ in a bad way

manuroe:
Fix vector-im/element-ios/issues/3737

The CS API does not take a deviceId parameter.

Tests are still happy:
image


bmarty (approved):
LGTM :)

PR #974, opened 2020-12-15
MXRoomSummary: Fix crash when lastMessageData cannot be decrypted

ismailgulek:
Fix https://github.com/vector-im/element-ios/issues/3879


gileluard (approved):
LGTM

PR #987, opened 2021-01-12
VoIP: PSTN Support

ismailgulek:
Adds PSTN support check and implements `thirdparty/user/$protocol` api. Fix vector-im/element-ios#3910


dbkr (approved):
Looks right afaics!

PR #988, opened 2021-01-12
Try to fix in Swift runtime failure: Unexpectedly found nil while imp…

gileluard:
fix https://github.com/vector-im/element-ios/issues/3916

Copy event array before processing it in order to make sure it's not altered during the process.


manuroe:
The root issue has been fixed. We do not need to complexify this part of the code

PR #989, opened 2021-01-13
MXUIKitBackgroundTask: Avoid thread switching when creating a background task

manuroe:
to keep threading model as it was before

Fix vector-im/element-ios/issues/3917.

### DIFF
--- a/MatrixSDK/Utils/MXUIKitApplicationStateService.swift
+++ b/MatrixSDK/Utils/MXUIKitApplicationStateService.swift
@@ -1,0 +1,116 @@
+// 
+// Copyright 2021 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


MatrixSDK/Utils/MXUIKitApplicationStateService.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+    
+    
+    //  MARK: - Method Overrides
+    
+    public override init() {
+        applicationState = .inactive
+        super.init()
+        
+        // Assume the service is created on the main tread
```
ismailgulek, line 69:
```suggestion
        // The service must be created on the main tread
        assert(Thread.isMainThread, "[MXUIKitApplicationStateService] initialized on non-main thread.")
```

ismailgulek: It may be a good idea to assert the thread for debugging purposes.

ismailgulek (approved):

PR #997, opened 2021-01-20
MXBackgroundSyncService: Clear the bg sync crypto db if needed 

manuroe:
so that we will be able to even clear corrupted db. This should fix vector-im/element-ios#3956.

It does not prevent the corruption (even Realm does not know how it can happen) but the app and the NSE will auto fix themselves. The user will just need to open the app. 
Opening the app will clear the background sync cache. On the next push, the bg sync service will detect that the bg sync cache is empty, it will then clear the realm db. Any corrupted db issue is solved.
Some notifications will not be able to decrypt but messages will be readable within the app.

This required some changes in `MXRealmCryptoStore` to delete the db without opening it.

### DIFF
--- a/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
+++ b/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
@@ -1440,43 +1440,10 @@ RLM_ARRAY_TYPE(MXRealmSecret)
         
         config.fileURL = realmFileURL;
         
-        // Check whether an existing db file has to be be moved from the default folder to the shared container.
-        if ([NSFileManager.defaultManager fileExistsAtPath:[defaultRealmFileURL path]])
+        // Make sure the full path exists before giving it to Realm
+        if (![NSFileManager.defaultManager fileExistsAtPath:realmFileFolderURL.path])
         {
-            if (![NSFileManager.defaultManager fileExistsAtPath:[realmFileURL path]])
-            {
-                // Move this db file in the container directory associated with the application group identifier.
-                NSLog(@"[MXRealmCryptoStore] Move the db file to the application group container");
-                
-                if (![NSFileManager.defaultManager fileExistsAtPath:realmFileFolderURL.path])
-                {
-                    [[NSFileManager defaultManager] createDirectoryAtPath:realmFileFolderURL.path withIntermediateDirectories:YES attributes:nil error:nil];
-                }
-                
-                NSError *fileManagerError = nil;
-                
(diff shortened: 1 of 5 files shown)


MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -1779,6 +1703,57 @@ + (RLMRealm*)realmForUser:(NSString*)userId andDevice:(NSString*)deviceId
...
+                                  URLByAppendingPathExtension:@"realm"];
+    
+    // Check for a potential application group id.
+    NSString *applicationGroupIdentifier = [MXSDKOptions sharedInstance].applicationGroupIdentifier;
+    if (applicationGroupIdentifier)
+    {
+        // Use the shared db file URL.
+        NSURL *sharedContainerURL = [[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:applicationGroupIdentifier];
+        NSURL *realmFileFolderURL = [sharedContainerURL URLByAppendingPathComponent:kMXRealmCryptoStoreFolder];
+        realmFileURL = [[realmFileFolderURL URLByAppendingPathComponent:userId] URLByAppendingPathExtension:@"realm"];
```
ismailgulek, line 1735:
```suggestion
        realmFileURL = [[realmFileFolderURL URLByAppendingPathComponent:realmFile] URLByAppendingPathExtension:@"realm"];
```

ismailgulek: I know it's coming from the old code, but it may be a good idea to fix this.

ismailgulek (approved):

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -337,15 +337,38 @@ + (void)deleteStoreWithCredentials:(MXCredentials*)credentials
...
 
+    // Delete db file directly
+    // So that we can even delete corrupted realm db
     RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
     NSURL *realmFileURL = [self realmFileURLForUserWithUserId:credentials.userId andDevice:credentials.deviceId];
     config.fileURL = realmFileURL;
     
+    if (![RLMRealm fileExistsForConfiguration:config])
+    {
+        NSLog(@"[MXRealmCryptoStore] deleteStore: Realm db does not exist");
```
ismailgulek, line 348: Maybe we could return in this case?

manuroe: oh, yes

ismailgulek (approved):

PR #982, opened 2020-12-22
Add xcframework argument to build.sh

pixlwave:
This PR adds an optional `xcframework` argument that can be passed to `build.sh`. When used it will create `build/MatrixSDK.xcframework` with targets for iOS, macOS, Catalyst and the Simulator with all arm64 and x86_64 slices where relevant. This distributable would make matrix integration easier for Xcode projects that aren't using cocoapods.

Depends on #983 to successfully build.

Signed-off-by: Doug Earnshaw pixlwave@users.noreply.github.com

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CHANGES.rst)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

PR #1002, opened 2021-01-27
Freeze MXResponse enum

pixlwave:
When using [niochat/MatrixSDK](https://github.com/niochat/MatrixSDK) (or MatrixSDK.xcframework) any switch against `MXResponse` produces a warning that unknown cases are unhandled.

Hoping this change is welcome as I can't foresee any other cases that could be added to MXResponse?

Please let me know if there's any changes you would like.

Signed-off-by: Doug Earnshaw pixlwave@users.noreply.github.com

### Pull Request Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@ Changes to be released in next version
  * 
 
 ⚠️ API Changes
- * 
+ * MXResponse has been frozen for binary compatibility.
 
 🗣 Translations
  * 
--- a/MatrixSDK/Contrib/Swift/MXResponse.swift
+++ b/MatrixSDK/Contrib/Swift/MXResponse.swift
@@ -46,9 +46,10 @@ import Foundation
      }
  
  */
-public enum MXResponse<T> {
+@frozen public enum MXResponse<T> {
     case success(T)
     case failure(Error)
+    // Note: Additional cases break binary compatibility
     
     /// Indicates whether the API call was successful
     public var isSuccess: Bool {


SBiOSoftWhare:
Hi @pixlwave thanks to point the issue. But by default user defined Swift enums are frozen: `All enums written in Swift outside of the standard library and overlays will implicitly be considered frozen in Swift 4.2. Enums imported from C will be non-frozen by default, with a new C-side annotation to treat them as frozen.` (from https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md). 
If you don't use CocoaPods can you check the build options you are using when you compile the SDK? There are also other consideration on build settings option regarding `frozen` attribute you can find here https://docs.swift.org/swift-book/ReferenceManual/Attributes.html

pixlwave:
Hey, thanks for looking into this 😄. We're building it with:
```
xcodebuild archive -workspace MatrixSDK.xcworkspace -scheme MatrixSDK-iOS -destination "generic/platform=iOS" -archivePath build/MatrixSDK-iOS SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES IPHONEOS_DEPLOYMENT_TARGET=11.0 GCC_WARN_ABOUT_DEPRECATED_FUNCTIONS=NO
```
(See #982)

I think it'll be `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` which is causing issues as this is enabling library evolution (I think with `-enable-library-evolution`). There's more here: https://swift.org/blog/library-evolution/ but sadly I think that building XCFrameworks results in differences to the compiled code 😕

pixlwave:
Interestingly that post notes:
> Swift Package Manager packages or binary frameworks that are internal to your app, should not be built with library evolution support.

I'll check and see if it's possible to build an XCFramework without `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` as that would definitely be preferable for our use case (I had thought it was a requirement 🤔 after watching the [WWDC session](https://developer.apple.com/videos/play/wwdc2019/416/))

pixlwave:
Hmmmm, it seems not 😣 The XCFramework gets created but none of the slices are added.

SBiOSoftWhare:
Indeed from Swift [documentation](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID620) 
> When the compiler isn’t in library evolution mode, all structures and enumerations are implicitly frozen, and this attribute is ignored.

SBiOSoftWhare (approved):

pixlwave:
Closed in favour of #1330 which includes an updated changelog. (The remote repository this was based on was deleted and so no further changes can be pushed to this PR is it seems).

PR #1005, opened 2021-01-29
Improve realm DB compaction

manuroe:
This PR offers 2 improvements on realm DB compaction. The 2 commits can be reviewed separately.

### DIFF
--- a/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
+++ b/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
@@ -1765,6 +1765,7 @@ RLM_ARRAY_TYPE(MXRealmSecret)
         }
     };
     
+    config.shouldCompactOnLaunch = nil;
     if ([self shouldCompactReamDBForUserWithUserId:userId andDevice:deviceId])
     {
         config.shouldCompactOnLaunch = ^BOOL(NSUInteger totalBytes, NSUInteger bytesUsed) {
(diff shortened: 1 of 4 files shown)


MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -1765,26 +1765,23 @@ + (RLMRealm*)realmForUser:(NSString*)userId andDevice:(NSString*)deviceId
...
-    config.shouldCompactOnLaunch = ^BOOL(NSUInteger totalBytes, NSUInteger bytesUsed) {
-        // totalBytes refers to the size of the file on disk in bytes (data + free space)
-        // usedBytes refers to the number of bytes used by data in the file
-        
-        static BOOL logDBFileSizeAtLaunch = YES;
-        if (logDBFileSizeAtLaunch)
-        {
+    if ([self shouldCompactReamDBForUserWithUserId:userId andDevice:deviceId])
+    {
+        config.shouldCompactOnLaunch = ^BOOL(NSUInteger totalBytes, NSUInteger bytesUsed) {
```
gileluard, line 1770: Your assumption is that `config.shouldCompactOnLaunch` is `false` by default but nothing in the [RLMRealmConfiguration documentation](https://realm.io/docs/objc/latest/api/Classes/RLMRealmConfiguration.html#/c:objc(cs)RLMRealmConfiguration(py)shouldCompactOnLaunch) say so. I would rather move the `if ([self shouldCompactReamDBForUserWithUserId:userId andDevice:deviceId])` in the closure to be safer.

gileluard (approved):

PR #1011, opened 2021-02-03
MXLogger: Reduce log size

manuroe:
Part of vector-im/element-ios/issues/#3903.

This PR contains 2 improvements. Commits can be reviewed separately.


gileluard (approved):
LGTM. Single weak point of you're algorithm is that logs 1,2,3,4 reach the size limit but 1,2,3,5 don't for example in case of 4 can be bigger than 5.

PR #1024, opened 2021-02-15
MXRoomSummary: Add creator user id property

SBiOSoftWhare:


manuroe (commented):
MXRoomSummary data are computed once. You need to persist `MXRoomSummary.creatorUserId` by:
 - Updating NSCoding implementation
 - Force a cache reset in MXFileStoreVersion

manuroe (approved):

PR #973, opened 2020-12-14
Added encryption for MXRealmCryptoStore

gileluard:

PR #1027, opened 2021-02-18
MXRealmCryptoStore: Make sure secrets in the DB are encrypted

manuroe:
Fix https://github.com/vector-im/element-ios/issues/3867
Require https://gitlab.matrix.org/matrix-org/olm/-/merge_requests/17.

It embeds the realm db migration code refactoring made in https://github.com/matrix-org/matrix-ios-sdk/pull/973 even if no migration is required here.

We cannot use the built-in encryption offered by Realm because it does not support multi-process. So, let's encrypt every single thing in a clear Realm db.

- [x] Olm data (account, megolm keys, etc). Done by using v2 pickle of OLMKit which use single global pickle key the app can store in keychain
- [x] Secrets used for 4S (private keys for key backup, MSDK, USK, SSK)


gileluard (approved):
LGTM

PR #1028, opened 2021-02-19
Support adding DISABLE_CALLKIT preprocessor macro to disable compilation of CallKit

qychrisyang:
Due to the policy, CallKit needs to be disabled.

> Chinese Ministry of Industry and Information Technology (MIIT) requested that CallKit functionality be deactivated in all apps available on the China App Store.


hrool:
+1, element-ios should be available for China. Currently, element-ios can't be found in China Appstore.

PR #1035, opened 2021-03-03
MXCrossSigning: Fix setupWithPassword method crash when a grace period is enabled

SBiOSoftWhare:
Fix vector-im/element-ios#4099

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Changes to be released in next version
  * 
 
 🐛 Bugfix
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,4 +1,4 @@
-Changes to be released in next version
+Changes in 0.18.4 (2021-03-03)
```
ismailgulek, line 1: I think we don't need to (maybe must not) change this line, as it'll be updated by our scripts.

ismailgulek (approved):

PR #1040, opened 2021-03-07
MXRoomState/MXRoomMembers: Fix memory leak and copying

Johennes:
`MXRoomState` and `MXRoomMembers` currently hold mutual strong references to each other via the `members` property and the `state` ivar, respectively. This leads to a memory leak preventing both objects from being deallocated. The leak is easy to reproduce in the Element iOS app. Just launch the app, use a few rooms and start the memory graph debugger. At this point you'll find orphaned objects retaining each other such as:

Screenshot 2021-03-06 at 20 21 28

Additionally, the leak is masqueraded by a bug in `MXRoomState`'s `copyWithZone:` method. The latter copies the `MXRoomMembers` object but doesn't update its `state` ivar to point to the `MXRoomState` copy. As a result, copying an `MXRoomState` object does *not* introduce another leak. I've noticed this when writing a unit test for the leak. Instead of the retain cycle, there now is a chain where `MXRoomMembers` is owned by one room state but points "back" to another.

Screenshot 2021-03-07 at 19 51 58

The present pull request addresses both issues. First, the `state` ivar on `MXRoomMembers` is made weak. This seems safe because `MXRoomState` retains its `members` property and `MXRoomMembers` should not exist on its own without a room state owning it.

Secondly, a new method `copyWithZone:andState:` is introduced on `MXRoomMembers`. The latter copies the receiver and updates the `state` ivar to the specified `MXRoomState`.

### Pull Request Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Changes to be released in next version
  * 
 
 🐛 Bugfix
- * 
+ * MXRoomState/MXRoomMembers: Fix memory leak and copying.
 
 ⚠️ API Changes
  * 
--- a/MatrixSDK/Data/MXRoomMembers.h
+++ b/MatrixSDK/Data/MXRoomMembers.h
@@ -32,12 +32,17 @@
 /**
  Create a `MXRoomMembers` instance.
 
- @paran state the room state it depends on.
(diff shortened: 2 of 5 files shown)


manuroe:
Thanks @Johennes !

PR #1050, opened 2021-03-17
E2E improvements

manuroe:
There are 2 separates improvements:
- Do not load room members in e2e rooms after an initial sync (1st commit)
- Add MXRoomSummary:.enableTrustTracking() to compute and maintain trust value for the given room (vector-im/element-ios/issues/4115).

### DIFF
--- a/MatrixSDK/Data/MXRoomSummary.m
+++ b/MatrixSDK/Data/MXRoomSummary.m
@@ -480,6 +480,30 @@ static NSUInteger const kMXRoomSummaryTrustComputationDelayMs = 1000;
 
 #pragma mark - Trust management
 
+- (void)enableTrustTracking:(BOOL)enable
+{
+    if (enable)
+    {
+        if (!_isEncrypted || _trust)
+        {
+            // Not applicable or already enabled
+            return;
+        }
+        
+        NSLog(@"[MXRoomSummary] enableTrustTracking: YES");
+        
+        // Bootstrap trust computation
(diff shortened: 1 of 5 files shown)


MatrixSDK/Data/MXRoomSummary.m
```diff
@@ -480,6 +480,30 @@ - (void)unregisterEventEditsListener
...
+    {
+        NSLog(@"[MXRoomSummary] enableTrustTracking: NO");
+        [self unregisterTrustLevelDidChangeNotifications];
+        _trust = nil;
+    }
+}
+
 - (void)setIsEncrypted:(BOOL)isEncrypted
 {
     _isEncrypted = isEncrypted;
```
ismailgulek, line 509: Should we also consider `_trust` here at line `if (_isEncrypted && [MXSDKOptions sharedInstance].computeE2ERoomSummaryTrust)`, like we did below? Or it's not applicable because the room is newly being encrypted, so we won't have a `_trust` yet?

`if (_isEncrypted && ([MXSDKOptions sharedInstance].computeE2ERoomSummaryTrust || _trust))`

ismailgulek (approved):

MatrixSDK/Data/MXRoomSummary.m
manuroe: This is the latter. `_trust` will be nil.

ismailgulek: Great, thanks.

PR #1051, opened 2021-03-19
MXCrypto: Split network request `/keys/query` into smaller requests (250 users max)

manuroe:
Fix vector-im/element-ios/issues/4123.


gileluard (approved):
LGTM

PR #1056, opened 2021-03-25
[Spaces] Create a space

SBiOSoftWhare:
Part of vector-im/element-ios/issues/4069

### DIFF
--- a/MatrixSDKTests/MXSpaceServiceTest.swift
+++ b/MatrixSDKTests/MXSpaceServiceTest.swift
@@ -1,0 +1,159 @@
+//
+// Copyright 2021 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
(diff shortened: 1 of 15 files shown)


MatrixSDKTests/MXSpaceServiceTest.swift
```diff
@@ -0,0 +1,159 @@
...
+                NotificationCenter.default.removeObserver(token)
+            }
+            
+            completion(roomSummary)
+        })
+    }
+    
+    // MARK: - Tests
+    
+    func testCreateSpace() throws {
```
manuroe, line 76: Can you describe the scenario of the test like https://github.com/matrix-org/matrix-ios-sdk/blob/37a6e164dcc46f6747393d19b96bc5396a63bf14/MatrixSDKTests/MXCrossSigningTests.m#L332-L337?
This is very helpful for maintenance.

MatrixSDKTests/MXSpaceServiceTest.swift
```diff
@@ -0,0 +1,159 @@
...
+            }
+            readyToTest(spaceService, mxSession!, expectation!)
+        }
+    }
+    
+    private func waitRoomSummaryUpdate(for roomId: String, completion: @escaping ((MXRoomSummary) -> Void)) {
+        
+        var token: NSObjectProtocol?
+        
+        token = NotificationCenter.default.addObserver(forName: NSNotification.Name.mxRoomSummaryDidChange, object: nil, queue: OperationQueue.main, using: { notification in
```
manuroe, line 55: I wonder if we should put this logic in the SDK, either in MXRoom or MXSpace.

SBiOSoftWhare: In fact this method only wait for the first catch of room summary update notification. I don't know if it can be useful when using the SDK.

manuroe (approved):

PR #1057, opened 2021-03-26
Notifications: Fix background sync out of memory

manuroe:
Fix https://github.com/vector-im/element-ios/issues/3957

This PR allows to use less RAM. Cached sync responses are no more aggregated to a single file but into several ones.
OOM started when this single big file was more than 2.5MB. This PR limits the size for separated files to 1MB.

Commits can be mainly reviewed one by one.
Most of commits are a refactoring to make possible to store sync responses on several files.


MatrixSDK/Background/Store/MXSyncResponseStoreManager.swift
```diff
@@ -0,0 +1,292 @@
...
+            if let event = event {
+                return event
+            }
+        }
+        
+        NSLog("[MXSyncResponseStoreManager] event: Not found event \(eventId) in room \(roomId)")
+        return nil
+    }
+    
+    func event(withEventId eventId: String, inRoom roomId: String, inSyncResponse response: MXCachedSyncResponse) -> MXEvent? {
```
ismailgulek, line 181:
```suggestion
    private func event(withEventId eventId: String, inRoom roomId: String, inSyncResponse response: MXCachedSyncResponse) -> MXEvent? {
```

MatrixSDK/Background/Store/MXSyncResponseStoreManager.swift
```diff
@@ -0,0 +1,292 @@
...
+            //  no current sync response, directly save the new one
+            NSLog("[MXSyncResponseStoreManager] updateStore: Start storing sync response")
+            let cachedSyncResponse = MXCachedSyncResponse(syncToken: syncToken,
+                                                         syncResponse: newSyncResponse)
+            _ = syncResponseStore.addSyncResponse(syncResponse: cachedSyncResponse)
+        }
+        
+        // Manage user account data
+        if let accountData = newSyncResponse.accountData {
+            syncResponseStore.accountData = accountData
```
ismailgulek, line 154: Shouldn't we merge account data? Is it right to override that?

manuroe: We have to override it. There is no incremental update of it in matrix.
Overriding is the only way to remove data for example.

MatrixSDK/Background/Store/Model/MXCachedSyncResponse.m
```diff
@@ -14,31 +14,46 @@
...
+}
 
 + (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
 {
-    MXSyncResponseStoreModel *syncResponseModel = [[MXSyncResponseStoreModel alloc] init];
+    MXCachedSyncResponse *cachedSyncResponse;
+    
+    NSString *syncToken;
+    MXSyncResponse *syncResponse;
+    MXJSONModelSetString(syncToken, JSONDictionary[@"sync_token"]);
```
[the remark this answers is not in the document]
manuroe: `MXSyncResponseFileStore` has a new file organisation. We will not decode an old file that used `prev_batch`. The code will not crash because of this.

I do not want to manage a migration because `MXSyncResponseFileStore` is just a skippable cache for the moment.

What is true is that we do not erase previous data. I am going to fix it.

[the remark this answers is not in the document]
ismailgulek: Makes sense, thanks.

MatrixSDK/Background/Store/MXSyncResponseStoreManager.swift
ismailgulek: Can you check again this Manu? Afaics in `-[MXSession handleAccountData:]`, we're updating the existing account data with new events and then store it again.

I'm afraid that some account data events will be ignored here, and even `newSyncResponse.accountData` can be empty dictionary.

ismailgulek (approved):

PR #1061, opened 2021-04-06
MXSpaceService: Update public space creation initial state events

SBiOSoftWhare:
Part of vector-im/element-ios/issues/4069


MatrixSDK/Room/State/MXRoomStateEventBuilder.swift
```diff
@@ -0,0 +1,77 @@
...
+            "content": content
+        ]
+    }
+}
+
+// MARK: - Public
+
+/// MXRoomStateEventBuilder enables to build initial state events
+@objcMembers
+public class MXRoomStateEventBuilder: NSObject {
```
manuroe, line 40: Can we call it `MXRoomInitialStateEventBuilder`?
It only creates such type of JSON, not JSON for full room state events.

manuroe (approved):

PR #1092, opened 2021-05-11
Cache Initial Sync Response

ismailgulek:
Fix vector-im/element-ios#4317


MatrixSDK/MXSession.m
```diff
@@ -1343,6 +1390,12 @@ - (void)serverSyncWithServerTimeout:(NSUInteger)serverTimeout
                 return;
             }
             
+            //  clear initial sync cache after handling sync response
+            for (NSString *responseId in self.initialSyncResponseCache.outdatedSyncResponseIds)
+            {
+                [self.initialSyncResponseCache deleteSyncResponseWithId:responseId];
```
manuroe, line 1396: It is too early. We need to wait for the completion of async operations launched by MXFileStore.commit().
We are missing an interface in MXStore to be able to synchronise on the end of those asynchronous operations.

One way to test it is to create a crash or restart your app just after this line. On a real account, MXFileStore will be empty on the next startup.

MatrixSDKTests/MXSessionTests.m
```diff
@@ -65,6 +66,71 @@ - (void)tearDown
...
+
+        [expectation fulfill];
+    }];
+}
+
+// Check MXSession updates initial sync cache when an error occurs handling the sync response.
+//
+// - Have Bob start a new session
+// - Run initial sync on Bob's session
+// - In the middle of the process, pause the session
```
manuroe, line 89: Closing the session and deleting the MXFileStore would be closer to the problem we are trying to improve.

ismailgulek: Updated to close the session, no need to delete data in store, it's already using a `MXNoStore` instance.

MatrixSDK/MXSession.m
ismailgulek: Thanks, updated.

MatrixSDK/MXSession.m
```diff
@@ -615,24 +619,31 @@ - (void)handleSyncResponse:(MXSyncResponse *)syncResponse
...
+                                                              }];
+        };
 
-        // Commit store changes done in [room handleMessages]
-        if ([self.store respondsToSelector:@selector(commit)])
+        // Commit store changes
+        if ([self.store respondsToSelector:@selector(commitWithCompletion:)])
         {
-            [self.store commit];
+            [self.store commitWithCompletion:completionBlock];
```
manuroe, line 641: We should continue to call `completion()` as soon as possible like before.
The data is loaded in memory. The app can work. The /sync stream can run.

If we wait for `commitWithCompletion`, we will delay the usage of the app for 70s in M's case.

CHANGES.rst
```diff
@@ -30,6 +30,8 @@ Changes in 0.18.12 (2021-05-12)
 
 🙌 Improvements
  * MXPushGatewayRestClient: Add timeout param to the HTTP method.
+ * MXSession: Cache initial sync response until it is fully handled (vector-im/element-ios/issues/4317).
```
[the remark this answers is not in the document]
ismailgulek: Weird, did not complain for a conflict, updated.

MatrixSDK/MXSession.m
ismailgulek: Updated.

ismailgulek: Updated.

MatrixSDKTests/MXSessionTests.m
```diff
@@ -65,6 +67,110 @@ - (void)tearDown
...
+                        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+
+                            [NSThread sleepForTimeInterval:1.0];
+
+                            XCTAssertEqual(cache.syncResponseIds.count,
+                                           0,
+                                           @"Initial sync cache must be used after successful restart");
+
+                            [expectation fulfill];
+                            [store deleteAllData];
```
manuroe, line 144:
```suggestion
```

There is no need. Plus, it should happen before the line above.

ismailgulek: You right, removed.

manuroe (approved):

PR #1106, opened 2021-05-28
MXRealmCryptoStore: ReadOnly Realms

ismailgulek:
Part of vector-im/element-ios#4352

### DIFF
--- a/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
+++ b/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
@@ -27,6 +27,8 @@
 #import "MXKeyProvider.h"
 #import "MXRawDataKey.h"
 #import "MXAes.h"
+#import "MatrixSDKSwiftHeader.h"
+#import "MXRealmHelper.h"
 
 
 NSUInteger const kMXRealmCryptoStoreVersion = 16;
@@ -306,6 +308,8 @@ RLM_ARRAY_TYPE(MXRealmSecret)
 
 #pragma mark - MXRealmCryptoStore
 
+NSString *const MXRealmCryptoStoreReadonlySuffix = @"readonly";
+
 @interface MXRealmCryptoStore ()
 {
     NSString *userId;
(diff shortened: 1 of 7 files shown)


MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -1827,6 +1867,30 @@ + (BOOL)shouldCompactReamDBForUserWithUserId:userId andDevice:(NSString*)deviceI
...
 
+#pragma mark - readOnly
+
+- (void)setReadOnly:(BOOL)readOnly
+{
+    NSLog(@"[MXRealmCryptoStore] setReadOnly: %@", readOnly ? @"YES" : @"NO");
+    _readOnly = readOnly;
+}
+
+static BOOL copyReadonlyDBNextTime = YES;
```
manuroe, line 1878: This static variable can lead to unpredictable result.
The app and extension run on their own process. copyReadonlyDBNextTime will be always YES for them at init. But the behavior will be different is different in our tests where we have only one process.

I wonder if we should remove the copyReadonlyDBNextTime, that is a bit hard to understand and add a method `deleteReadOnlyCache` that we use when needed.

Then, we could use the existence of the cache file in https://github.com/matrix-org/matrix-ios-sdk/pull/1106/files?diff=split&w=1#diff-0a4aae3a9f88fcb6a3fc5f1f0d20d2afb3ebbbb55951e55025dfd0c0aa311a26R1640.

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -1621,7 +1635,30 @@ + (RLMRealm*)realmForUser:(NSString*)userId andDevice:(NSString*)deviceId
...
+            NSError *error;
+            NSDictionary *fileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:config.fileURL.path error:nil];
+            unsigned long long fileSize = [[fileAttributes objectForKey:NSFileSize] unsignedLongLongValue];
+            MXStopwatch *stopwatch = [MXStopwatch new];
+            [[NSFileManager defaultManager] removeItemAtURL:readOnlyURL error:nil];
+            [[NSFileManager defaultManager] copyItemAtURL:config.fileURL toURL:readOnlyURL error:&error];
+            NSLog(@"[MXRealmCryptoStore] realmForUser: readonly copy file lasted %@, fileSize: %@", [stopwatch readableIn:MXStopwatchMeasurementUnitMilliseconds], [MXTools fileSizeToString:fileSize round:NO]);
+            if (!error)
+            {
+                config.fileURL = readOnlyURL;
```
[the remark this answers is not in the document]
ismailgulek: Updated

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
ismailgulek: Refactored.

manuroe (approved):

PR #1116, opened 2021-06-03
Use `MXKeyProvider` for Room Last Message Encryption Key

ismailgulek:
Additional improvements for vector-im/element-ios#4360

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -473,7 +473,7 @@ typedef void (^MXOnResumeDone)(void);
                             // Make sure the last message has been decrypted
                             // In case of an initial sync, we save decryptions to save time. Only unread messages are decrypted.
                             // We need to decrypt already read last message.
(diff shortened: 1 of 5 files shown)


MatrixSDK/MXSession.m
```diff
@@ -2832,7 +2832,7 @@ - (void)fixRoomsSummariesLastMessageWithMaxServerPaginationCount:(NSUInteger)max
     
     for (MXRoomSummary *summary in self.roomsSummaries)
     {
-        if (!summary.lastMessage)
+        if (!summary.lastMessage || !summary.lastMessage.text)
```
manuroe, line 2835: It supposes that the code using the SDK has this specific behavior but we cannot guarantee it.

As we already have a dispatchGroup here, we could fetch the `MXEvent` to check that it was successfully decrypted.

ismailgulek: You're right. Updated as you said, kit PR closed.

PR #1117, opened 2021-06-04
Support full rehydration feature.

gileluard:
needed for https://github.com/vector-im/element-ios/issues/4255


MatrixSDK/MXRestClient.m
```diff
@@ -4014,6 +4067,46 @@ - (MXHTTPOperation*)uploadKeys:(NSDictionary*)deviceKeys oneTimeKeys:(NSDictiona
                                  }];
 }
 
+- (MXHTTPOperation*)uploadKeys:(NSDictionary*)deviceKeys oneTimeKeys:(NSDictionary*)oneTimeKeys
+               forDeviceWithId:(NSString*)deviceId
```
uhoreg, line 4071: this parameter doesn't seem to be used anywhere.  It should be included in the path as `/keys/upload/<deviceId>` (after URL-encoding the device ID).

MatrixSDK/Crypto/Dehydration/MXDehydrationManager.m
```diff
@@ -0,0 +1,178 @@
...
+//                    [NSString stringWithFormat:@"ed25519:%@", deviceId]: account.identityKeys[@"ed25519"],
+//                    [NSString stringWithFormat:@"curve25519:%@", deviceId]: account.identityKeys[@"curve25519"]
+//            }
+//        };
+//
+//        NSString *deviceSignature = [account signMessage:[NSJSONSerialization dataWithJSONObject:deviceKeys options:0 error:&error]];
+
+
+        NSError *error = nil;
+        NSString *signature = [account signMessage:[NSJSONSerialization dataWithJSONObject:deviceInfo.signalableJSONDictionary options:0 error:&error]];
```
uhoreg, line 121: Signatures need to be done using canonical JSON, like in  https://github.com/matrix-org/matrix-ios-sdk/blob/develop/MatrixSDK/Crypto/MXOlmDevice.m#L116 (or you can use the `signJSON` method)

MatrixSDK/Crypto/Dehydration/MXDehydrationService.m
```diff
@@ -0,0 +1,260 @@
...
+        [self->session.matrixRestClient claimDehydratedDeviceWithId:device.deviceId Success:^(BOOL isClaimed) {
+            if (!isClaimed)
+            {
+                MXLogDebug(@"[MXDehydrationManager] rehydrateDevice: device already claimed.");
+                success();
+                return;
+            }
+
+            MXLogDebug(@"[MXDehydrationManager] rehydrateDevice: using dehydrated device");
+            self->session.matrixRestClient.credentials.deviceId = device.deviceId;
```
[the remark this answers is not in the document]
manuroe: To go further on this architectural change, we could have some stateless static methods in MXCrypto that apply the change for the rehydrated device.
Thus, the next instantiated MXSession will start with the right data.
It seems safer than calling this method at the right time where right time window is quite small.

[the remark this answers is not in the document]
manuroe: MXDehydrationService can do those changes internally for the SDK user and just return the device id. But it means that no MXSession must run and this service should not depend on MXSession for that part.

The flow would be:
- Have a MXRestClient
- Init MXDehydrationService with it
- MXDehydrationService.rehydrate(). It will return a device id
- Let the SDK user store the device id
- Start a new MXRestClient with new credentials
- Init MXSession with it as usual. There should be no change on this existing code for this call

MatrixSDK/Crypto/Dehydration/MXDehydrationService.h
```diff
@@ -0,0 +1,73 @@
...
+ Service in charge of dehydrating and rehydrating a device.
+ 
+ @see https://github.com/uhoreg/matrix-doc/blob/dehydration/proposals/2697-device-dehydration.md for more details
+ */
+@interface MXDehydrationService : NSObject
+
+/**
+ Singleton method
+ */
++ (instancetype)sharedInstance;
```
manuroe, line 41: Do we really need a shared instance?
The current implementation does not support multi-account well.

MatrixSDK/Crypto/Dehydration/MXDehydrationService.m
```diff
@@ -0,0 +1,295 @@
...
+                success(nil);
+            });
+            return;
+        }
+        
+        if (![device.algorithm isEqual:MXDehydrationAlgorithm])
+        {
+            MXLogError(@"[MXDehydrationManager] rehydrateDevice: Wrong algorithm for dehydrated device.");
+            dispatch_async(dispatch_get_main_queue(), ^{
+                failure([NSError errorWithDomain:kMXNSErrorDomain code:0 userInfo:@{NSLocalizedDescriptionKey: @"Wrong algorithm for dehydrated device"}]);
```
manuroe, line 161: MXDehydrationService should have its own error domain and code like `MXAesErrorDomain` for an objc example.

MatrixSDK/Crypto/Dehydration/MXDehydrationService.m
```diff
@@ -0,0 +1,295 @@
...
+#import "MXRawDataKey.h"
+#import "MXSession.h"
+
+NSString *const MXDehydrationAlgorithm = @"org.matrix.msc2697.v1.olm.libolm_pickle";
+
+NSString *const MXDehydrationServiceKeyDataType = @"org.matrix.sdk.dehydration.service.key";
+
+@interface MXDehydrationService()
+{
+    BOOL inProgress;
```
manuroe, line 33: Could we expose it publicly?

That will avoid to call `dehydrateDeviceWithMatrixRestClient` when it is useless.

MatrixSDK/Crypto/Dehydration/MXDehydrationService.h
```diff
@@ -0,0 +1,73 @@
...
+ @param restClient client used to call the dehydration API
+ @param crypto crypto used to self sign the dehydrated device
+ @param dehydrationKey key used to pickle the Olm account
+ @param success callback called in case of success (deviceId not null) or if the process is canceled (deviceId is null)
+ @param failure callback called in case of unexpected failure
+ */
+- (void)dehydrateDeviceWithMatrixRestClient:(MXRestClient*)restClient
+                                     crypto:(MXCrypto*)crypto
+                             dehydrationKey:(NSData*)dehydrationKey
+                                    success:(void (^)(NSString * _Nullable deviceId))success
```
manuroe, line 55: We should not hide errors behind `success(nil)` but report proper errors instead.

manuroe (approved):

PR #1124, opened 2021-06-11
MXSession: Fix app that can fail to resume

manuroe:
Fix vector-im/element-ios/issues/4417

The root cause of the issue is a race condition in the MXSession pause management between the sdk state machine and the kit state machine.

The SDK was blocked and dead in MXSessionStateSyncInProgress state. The kit thought it does not need to resume it on app resume.

This fix is very related on how the kit and the sdk work together. But it could make sense in a more general manner.

I was able to reproduce the issue ("[MXSession] The connection has been cancelled.") by adding a delay of 25s in /sync response. Then, if you background the app just after starting it, you will hit this case.

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1519,9 +1519,9 @@ typedef void (^MXOnResumeDone)(void);
                 // This happens when the SDK cannot make any more requests because the app is in background
                 // and the background task is expired or going to expire.
                 // The app should have paused the SDK before but it did not. So, pause the SDK ourselves.
-                // Note that we need to come back to MXSessionStateRunning in order to be able to pause.
+                // Note that we need to come back to MXSessionStatePauseRequested in order to be able to pause.
                 MXLogDebug(@"[MXSession] -> Go to pause");
-                [self setState:MXSessionStateRunning];
+                [self setState:MXSessionStatePauseRequested];
                 [self pause];
             }
         }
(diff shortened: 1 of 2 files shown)


MatrixSDK/MXSession.m
```diff
@@ -1512,7 +1512,18 @@ - (void)handleServerSyncError:(NSError*)error forRequestWithServerTimeout:(NSUIn
...
+            MXLogDebug(@"[MXSession] The connection has been cancelled in state %@", @(_state));
+
+            if (_state == MXSessionStateSyncInProgress)
+            {
+                // This happens when the SDK cannot make any more requests because the app is in background
+                // and the background task is expired or going to expire.
+                // The app should have paused the SDK before but it did not. So, pause the SDK ourselves.
+                // Note that we need to come back to MXSessionStateRunning in order to be able to pause.
+                MXLogDebug(@"[MXSession] -> Go to pause");
+                [self setState:MXSessionStateRunning];
```
ismailgulek, line 1524: Instead of going into _running_ state, `MXSessionStatePauseRequested` makes more sense to me, in terms of state machine logically, which should also be applicable to pause the session.

I'm just imagining that if we cannot pause the session for some reason, we'll stay in running state and next time there will be a similar issue when we tried to resume the session, we'll see the session is already _running_ and probably do nothing.

ismailgulek (approved):

PR #1126, opened 2021-06-15
Fix Last Message Regression

ismailgulek:


manuroe (approved):
Ok for the change but I created https://github.com/vector-im/element-ios/issues/4440 to have a better solution.

PR #1127, opened 2021-06-15
VoIP: Change Audio Output Routes

ismailgulek:
Use headphones and Bluetooth devices when available.


simonwiedmer:
Verified "Fix audio routing when there is no external device" and it is working now

manuroe (approved):

PR #1134, opened 2021-06-21
Auto Accept Room Invites

ismailgulek:
Introduces `autoAcceptRoomInvites` flag on `MXSDKOptions` which is false by default.

### DIFF
--- a/MatrixSDK/Background/MXBackgroundSyncService.swift
+++ b/MatrixSDK/Background/MXBackgroundSyncService.swift
@@ -504,6 +504,21 @@ public enum MXBackgroundSyncServiceError: Error {
             handleToDeviceEvent(event)
         }
         
+        if MXSDKOptions.sharedInstance().autoAcceptRoomInvites,
+           let invitedRooms = syncResponse.rooms?.invite {
+            invitedRooms.forEach { roomId, roomSync in
+                MXLog.debug("[MXBackgroundSyncService] handleSyncResponse: Auto-accepting room invite for \(roomId)")
+                restClient.joinRoom(roomId) { response in
+                    switch response {
+                    case .success:
+                        MXLog.debug("[MXBackgroundSyncService] handleSyncResponse: Joined room: \(roomId)")
+                    case .failure(let error):
+                        MXLog.error("[MXBackgroundSyncService] handleSyncResponse: Failed to join room: \(roomId), error: \(error)")
+                    }
+                }
+            }
+        }
+        
         MXLog.debug("[MXBackgroundSyncService] handleSyncResponse: Next sync token: \(syncResponse.nextBatch)")
     }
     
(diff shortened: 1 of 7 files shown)


MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -504,6 +504,21 @@ public enum MXBackgroundSyncServiceError: Error {
...
+        if MXSDKOptions.sharedInstance().autoAcceptRoomInvites,
+           let invitedRooms = syncResponse.rooms?.invite {
+            invitedRooms.forEach { roomId, roomSync in
+                MXLog.debug("[MXBackgroundSyncService] handleSyncResponse: Auto-accepting room invite for \(roomId)")
+                restClient.joinRoom(roomId) { response in
+                    switch response {
+                    case .success:
+                        MXLog.debug("[MXBackgroundSyncService] handleSyncResponse: Joined room: \(roomId)")
+                    case .failure(let error):
+                        MXLog.error("[MXBackgroundSyncService] handleSyncResponse: Failed to join room: \(roomId), error: \(error)")
```
manuroe, line 516: Do you know if the `/join` request generates an error if we double join a room?
This is what will happen if the auto-join is firstly managed by the NSE.

manuroe (commented):
With https://github.com/vector-im/element-ios/pull/4452, invites will be hidden forever if there is a network failure on `/join`.

We need another mechanism to retry. We can do it on start but it will be great to be able to retry while the app is running.

MatrixSDK/Background/MXBackgroundSyncService.swift
ismailgulek: You're right. I've checked this and seems we don't have an error. I need to fix this.

ismailgulek: Error added with https://github.com/matrix-org/matrix-ios-sdk/pull/1134/commits/260d68416d5d0c92914b48af08c2182468a8f898

ismailgulek: Retry mechanism added with https://github.com/matrix-org/matrix-ios-sdk/pull/1134/commits/93aca3b8481d8d606dee8bb99c2984e8a4182feb

MatrixSDK/MXSession.m
```diff
@@ -2217,6 +2224,21 @@ - (MXHTTPOperation*)joinRoom:(NSString*)roomIdOrAlias
                      success:(void (^)(MXRoom *room))success
                      failure:(void (^)(NSError *error))failure {
     
+    if ([self isJoinedOnRoom:roomIdOrAlias])
```
manuroe, line 2227: Why not using `RoomSummary.membershipTransitionState` and `MXMembershipTransitionStateJoining`?

ismailgulek: Updated for both is joined and is invited logic.

manuroe (approved):

PR #1138, opened 2021-06-25
VoIP: Audio Routing

ismailgulek:


manuroe (commented):
LGTM.

Can we rename `MXAudioOutputRoute*` to `MXiOSAudioOutputRoute*`? 
So that we know that is target specific. We may have `MXMacOSAudioOutputRoute` in the future.

manuroe (approved):

PR #1142, opened 2021-06-30
MXSDKOptions: Add wellknownDomainUrl to customise the domain for wellknown

manuroe:
fix https://github.com/vector-im/element-ios/issues/#4489.

I also added a commit to refresh the wellknown on every start to get latest services available on the server.

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -4094,18 +4094,20 @@ typedef void (^MXOnResumeDone)(void);
     MXLogDebug(@"[MXSession] refreshHomeserverWellknown");
     if (!autoDiscovery)
     {
-        NSString *homeServer;
-        
-        // Retrieve the domain from the user id as it can be different from the `MXRestClient.homeserver` that uses the client-server API endpoint domain.
-        NSString *userDomain = [MXTools serverNameInMatrixIdentifier:self.myUserId];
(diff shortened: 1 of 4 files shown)


MatrixSDK/MXSession.m
```diff
@@ -4094,18 +4091,20 @@ - (MXHTTPOperation *)refreshHomeserverWellknown:(void (^)(MXWellKnown *))success
...
-        
-        // Retrieve the domain from the user id as it can be different from the `MXRestClient.homeserver` that uses the client-server API endpoint domain.
-        NSString *userDomain = [MXTools serverNameInMatrixIdentifier:self.myUserId];
-        
-        if (userDomain)
-        {
-            homeServer =  [NSString stringWithFormat:@"https://%@", userDomain];
-        }
-        else
+        NSString *homeServer = [MXSDKOptions sharedInstance].wellknownDomainUrl;
```
giomfo, line 4094: When the autoDiscovery is already instantiated, don't you think we should compare its url with the potential `wellknownDomainUrl` from MXKSDKOptions, in order to create it again if the urls don't match

manuroe: SDK options are build settings. We must not need to manage complex logic.

giomfo (approved):

PR #1145, opened 2021-07-01
Improve video transcoding quality

pixlwave:
Draft for now as I have a couple of questions.

Add `maxUploadSize:failure:` to `MXRestClient` and stores the size in `MXSession`.
Pass a target file size to `MXTools` and use `AVAssetExportPreset3840x2160` when transcoding.

### DIFF
--- a/MatrixSDK/Data/Store/MXStore.h
+++ b/MatrixSDK/Data/Store/MXStore.h
@@ -324,6 +324,21 @@
 - (void)close;
 
 
+#pragma mark - Media repository
+
+/**
+ The maximum size an upload can be in bytes.
+ */
+@property (nonatomic, readonly) NSInteger maxUploadSize;
+
+/**
+ Store the maximum upload size.
+
+ @param maxUploadSize The maximum upload size to store.
+ */
+- (void)storeMaxUploadSize:(NSInteger)maxUploadSize;
+
+
 #pragma mark - Permanent storage -
 
 /**
--- a/MatrixSDK/MXRestClient.h
+++ b/MatrixSDK/MXRestClient.h
@@ -1925,6 +1925,17 @@ typedef MXHTTPOperation* (^MXRestClientIdentityServerAccessTokenHandler)(void (^
                           failure:(void (^)(NSError *error))failure
                    uploadProgress:(void (^)(NSProgress *uploadProgress))uploadProgress NS_REFINED_FOR_SWIFT;
 
+/**
+Get the maximum size a media upload can be in bytes.
+ 
+@param success A block object called when the operation succeeds. It provides the maximum size an upload can be in bytes.
+@param failure A block object called when the operation fails.
+
+@return a MXHTTPOperation instance.
+*/
+- (MXHTTPOperation*) maxUploadSize:(void (^)(NSInteger maxUploadSize))success
+                           failure:(void (^)(NSError *error))failure;
+
 
 #pragma mark - Antivirus server API
 
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -3687,7 +3687,7 @@ MXAuthAction;
                                      {
                                          __block NSInteger maxUploadSize;
                                          [self dispatchProcessing:^{
-                                             MXJSONModelSetInteger(maxUploadSize, JSONResponse[@"m.upload.size"]);
+                                             MXJSONModelSetInteger(maxUploadSize, JSONResponse[@"m.upload.size"] ?: [NSNumber numberWithInteger:-1]);
                                          } andCompletion:^{
                                              success(maxUploadSize);
                                          }];
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -3685,7 +3685,7 @@ MXAuthAction;
 
                                      if (success)
                                      {
-                                         __block NSInteger maxUploadSize;
+                                         __block NSInteger maxUploadSize = 0;
                                          [self dispatchProcessing:^{
                                              MXJSONModelSetInteger(maxUploadSize, JSONResponse[@"m.upload.size"] ?: [NSNumber numberWithInteger:-1]);
                                          } andCompletion:^{
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -3673,6 +3673,32 @@ MXAuthAction;
                                  }];
 }
 
+- (MXHTTPOperation*) maxUploadSize:(void (^)(NSInteger maxUploadSize))success
+                           failure:(void (^)(NSError *error))failure
+{
+    MXWeakify(self);
+    return [httpClient requestWithMethod:@"GET"
+                                    path:[NSString stringWithFormat:@"%@/config", contentPathPrefix]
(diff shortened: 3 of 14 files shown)


MatrixSDK/MXRestClient.h
```diff
@@ -1925,6 +1925,17 @@ typedef MXHTTPOperation* (^MXRestClientIdentityServerAccessTokenHandler)(void (^
...
 
+/**
+Get the maximum size a media upload can be in bytes.
+ 
+@param success A block object called when the operation succeeds. It provides the maximum size an upload can be in bytes.
+@param failure A block object called when the operation fails.
+
+@return a MXHTTPOperation instance.
+*/
+- (MXHTTPOperation*) maxUploadSize:(void (^)(NSInteger maxUploadSize))success
```
pixlwave, line 1936: `maxUploadSize` or `getMaxUploadSize`? I couldn't really see a convention one way or the other.

MatrixSDK/MXRestClient.m
```diff
@@ -3673,6 +3673,32 @@ - (MXHTTPOperation*) uploadContent:(NSData *)data
...
+                                    path:[NSString stringWithFormat:@"%@/config", contentPathPrefix]
+                              parameters:nil
+                                 success:^(NSDictionary *JSONResponse) {
+                                     MXStrongifyAndReturnIfNil(self);
+
+                                     if (success)
+                                     {
+                                         __block NSInteger maxUploadSize;
+                                         [self dispatchProcessing:^{
+                                             MXJSONModelSetInteger(maxUploadSize, JSONResponse[@"m.upload.size"]);
```
pixlwave, line 3690: The spec [mentions](https://matrix.org/docs/spec/client_server/r0.6.1#id406) that `m.upload.size` can be missing or null. Does this need to be handled here?

MatrixSDK/MXSession.h
```diff
@@ -398,6 +398,11 @@ FOUNDATION_EXPORT NSString *const kMXSessionNoRoomTag;
  */
 @property (nonatomic, readonly) MXMediaManager *mediaManager;
 
+/**
+ The the maximum size an upload can be in bytes.
+ */
+@property (nonatomic, readonly) NSInteger maxUploadSize;
```
pixlwave, line 404: I was split between storing this here or in the media manager?

MatrixSDK/MXSession.m
```diff
@@ -946,6 +946,12 @@ - (void)resume:(void (^)(void))resumeDone
     [self handleBackgroundSyncCacheIfRequiredWithCompletion:^{
         [self _resume:resumeDone];
     }];
+    
+    [self.matrixRestClient maxUploadSize:^(NSInteger maxUploadSize) {
```
pixlwave, line 950: I'm pretty sure this is in the wrong place. Would appreciate a suggestion of where it should really go 🙂

MatrixSDK/MXRestClient.m
manuroe: `MXJSONModelSetInteger` does the check for you. We could default to -1 if the server does not provide it

MatrixSDK/MXRestClient.h
manuroe: I would say no "get" in objc. It is fine like this for me.

MatrixSDK/MXSession.h
manuroe: MXSession is ok. Media manager should be a seperate component. I would like us to replace the cache it manages by another 3rd party library.

MatrixSDK/Utils/MXTools.m
```diff
@@ -827,8 +828,9 @@ + (void)convertVideoToMP4:(NSURL*)videoLocalURL
     // Convert video container to mp4
     // Use medium quality to save bandwidth
     AVURLAsset* videoAsset = [AVURLAsset URLAssetWithURL:videoLocalURL options:nil];
-    AVAssetExportSession *exportSession = [AVAssetExportSession exportSessionWithAsset:videoAsset presetName:AVAssetExportPresetMediumQuality];
+    AVAssetExportSession *exportSession = [AVAssetExportSession exportSessionWithAsset:videoAsset presetName:AVAssetExportPreset3840x2160];
```
manuroe, line 831: 4K scares me :/

We should make it configurable with `MXSDKOptions` with a property like `videoConversionPresetName` and use HD as the default preset.

I do not think it worths to pass it as a parameter to `MXRoom.sendVideo()`(at least for the current behavior of Element).

MatrixSDK/MXSession.m
pixlwave: Done!

MatrixSDK/Utils/MXTools.m
pixlwave: > 4K scares me :/

Agreed 😄

Added to `MXSDKOptions`.

MatrixSDK/MXRestClient.m
pixlwave: Done and `MXTools` now checks for a target size > 0 before assigning the limit.

MatrixSDK/Data/Store/MXStore.h
```diff
@@ -324,6 +324,21 @@
 - (void)close;
 
 
+#pragma mark - Media repository
```
pixlwave, line 327: Note: these follow an `@optional` further up, and are only implemented in `MXFileStore`.

MatrixSDK/MXRestClient.m
```diff
@@ -3673,6 +3673,32 @@ - (MXHTTPOperation*) uploadContent:(NSData *)data
...
+    MXWeakify(self);
+    return [httpClient requestWithMethod:@"GET"
+                                    path:[NSString stringWithFormat:@"%@/config", contentPathPrefix]
+                              parameters:nil
+                                 success:^(NSDictionary *JSONResponse) {
+                                     MXStrongifyAndReturnIfNil(self);
+
+                                     if (success)
+                                     {
+                                         __block NSInteger maxUploadSize;
```
manuroe, line 3688: Local variables must be initialised in objc.
They may be 0 in debug but not in release build...

```suggestion
                                         __block NSInteger maxUploadSize = 0;
```


Note that ivars are initialised to 0 or nil or NO because the instance memory allocator does it.

MatrixSDK/MXSession.m
```diff
@@ -881,6 +881,13 @@ - (void)_startWithSyncFilterId:(NSString *)syncFilterId onServerSyncDone:(void (
 
     // Refresh wellknown data
     [self refreshHomeserverWellknown:nil failure:nil];
+    
+    // Get the maxmium file size allowed for uploading media
+    [self.matrixRestClient maxUploadSize:^(NSInteger maxUploadSize) {
+        [self.store storeMaxUploadSize:maxUploadSize];
+    } failure:^(NSError *error) {
+        MXLogDebug(@"[MXSession] Failed to get maximum upload size.");
```
manuroe, line 889:
```suggestion
        MXLogError(@"[MXSession] Failed to get maximum upload size.");
```

manuroe (commented):
We also need a change to CHANGES

manuroe (approved):

PR #1167, opened 2021-07-21
MXRoomSummary: Cache local unread event count 

SBiOSoftWhare:
Fix https://github.com/vector-im/element-ios/issues/4585


manuroe (commented):
MXFileStore must be reset

manuroe (approved):

PR #1170, opened 2021-07-26
MXCryptoStore: Keep current store version after resetting data

manuroe:
Fix vector-im/element-ios/issues/4594

to avoid dead state on an initial sync.

There is a db version update to fix possible wrong value for MXRealmOlmAccount.cryptoVersion.

### DIFF
--- a/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
+++ b/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
@@ -416,9 +416,6 @@ NSString *const MXRealmCryptoStoreReadonlySuffix = @"readonly";
                 [realm transactionWithBlock:^{
                     [realm deleteAllObjects];
                 }];
-                
-                // We must keep the last version
-                [self setVersion:MXCryptoVersionLast];
             }
             else
             {
@@ -450,7 +447,8 @@ NSString *const MXRealmCryptoStoreReadonlySuffix = @"readonly";
             {
                 MXLogDebug(@"[MXRealmCryptoStore] Credentials do not match");
                 [MXRealmCryptoStore deleteStoreWithCredentials:credentials];
-                return [MXRealmCryptoStore createStoreWithCredentials:credentials];
+                self = [MXRealmCryptoStore createStoreWithCredentials:credentials];
+                self.cryptoVersion = MXCryptoVersionLast;
             }
         }
         
--- a/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
+++ b/MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
@@ -32,7 +32,7 @@
 #import "MXBackgroundModeHandler.h"
 
 
-NSUInteger const kMXRealmCryptoStoreVersion = 16;
+NSUInteger const kMXRealmCryptoStoreVersion = 17;
 
 static NSString *const kMXRealmCryptoStoreFolder = @"MXRealmCryptoStore";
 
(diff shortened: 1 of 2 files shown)


MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -416,10 +416,13 @@ + (void)_deleteStoreWithCredentials:(MXCredentials*)credentials readOnly:(BOOL)r
                 [realm transactionWithBlock:^{
                     [realm deleteAllObjects];
                 }];
+                
+                // We must keep the last version
+                [self setVersion:MXCryptoVersionLast];
```
ismailgulek, line 421: This seems useless to me. We don't have a stored `version` property on this class.

You probably wanted to use `cryptoVersion` but that won't help neither, because we removed the account object above. Maybe we need a metadata object to store the version separately.

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -2165,6 +2168,20 @@ + (BOOL)finaliseMigrationWith:(RLMMigration *)migration oldSchemaVersion:(uint64
...
+                MXLogDebug(@"[MXRealmCryptoStore] Migration from schema #16 -> #17");
+                
+                MXLogDebug(@"[MXRealmCryptoStore]    Make sure MXRealmOlmAccount.cryptoVersion is MXCryptoVersion2");
+                [migration enumerateObjects:MXRealmOlmAccount.className block:^(RLMObject *oldObject, RLMObject *newObject) {
+                    NSNumber *version;
+                    MXJSONModelSetNumber(version, oldObject[@"cryptoVersion"]);
+                    if (version && version.intValue == 0)
+                    {
+                        MXLogDebug(@"[MXRealmCryptoStore]    -> Fix MXRealmOlmAccount.cryptoVersion");
+                        newObject[@"cryptoVersion"] = @(MXCryptoVersion2);
```
ismailgulek, line 2182:
```suggestion
                        newObject[@"cryptoVersion"] = @(MXCryptoVersionLast);
```

ismailgulek: `MXCryptoVersionLast` makes more sense to me to use here.

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
```diff
@@ -2165,6 +2168,20 @@ + (BOOL)finaliseMigrationWith:(RLMMigration *)migration oldSchemaVersion:(uint64
                 
             case 15:
                 MXLogDebug(@"[MXRealmCryptoStore] Migration from schema #15 -> #16: Nothing to do (added optional MXRealmSecret.encryptedSecret)");
+                
+            case 16:
+                MXLogDebug(@"[MXRealmCryptoStore] Migration from schema #16 -> #17");
+                
+                MXLogDebug(@"[MXRealmCryptoStore]    Make sure MXRealmOlmAccount.cryptoVersion is MXCryptoVersion2");
```
ismailgulek, line 2175:
```suggestion
                MXLogDebug(@"[MXRealmCryptoStore]    Make sure MXRealmOlmAccount.cryptoVersion is MXCryptoVersionLast");
```

ismailgulek: Same as below.

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
manuroe: Argh. Thanks. Autocompletion led me to a wrong thing.

MatrixSDK/Crypto/Data/Store/MXRealmCryptoStore/MXRealmCryptoStore.m
manuroe: We need an accurate value to do the correct migration steps.

When we will move to MXCryptoVersion3, let's say in db schema 26, there will certainly be some migration work to do. If you use MXCryptoVersionLast here, the db will move to MXCryptoVersion3 earlier than expected, specially if someone needs to migrate from schema 15 to 26.

PR #1177, opened 2021-08-02
Add excludingUserIDs parameter to displayname and avatar updater methods.

pixlwave:
Allows specified users to be excluded when computing the room's name and avatar. The new functions contain the implementation, with the older ones now calling into these.

Includes tests for the functions.


manuroe (approved):

manuroe (approved):
Once https://github.com/vector-im/element-ios/pull/4643#discussion_r684974410 is fixed, we can merge those PRs and close the issue.

PR #1180, opened 2021-08-04
[Spaces] M10.5 SDK: Support suggested rooms #4500

gileluard:
Fixes https://github.com/vector-im/element-ios/issues/4500

### DIFF
--- a/MatrixSDK/Space/MXSpace.swift
+++ b/MatrixSDK/Space/MXSpace.swift
@@ -58,8 +58,6 @@ public class MXSpace: NSObject {
     public private(set) var otherMembersId: [String] = []
     public private(set) var membersId: [String] = []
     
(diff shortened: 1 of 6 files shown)


MatrixSDK/Space/MXSpace.swift
```diff
@@ -57,6 +57,8 @@ public class MXSpace: NSObject {
     public private(set) var childRoomIds: [String] = []
     public private(set) var otherMembersId: [String] = []
     public private(set) var membersId: [String] = []
+    
+    public var lastSpaceChildrenSummary: MXSpaceChildrenSummary?
```
SBiOSoftWhare, line 61: Can we add a comment for the purpose of this property?

gileluard: This member is not needed anymore.

SBiOSoftWhare (approved):

PR #1185, opened 2021-08-05
Fix #1184: Allow custom HTTP headers on registration and login.

### LINKED ISSUE #1184: Allow custom HTTP headers on registration and login
**Is your feature request related to a problem? Please describe.**

We currently run federated servers for a very vulnerable audience.

We want to be a good citizen in the fediverse, so we want to stop spammers from using our
systems.

The currently available solutions have the following problems for us:

1. We don't want to sell our audience to Google via Recaptcha.
2. They're not English speakers, might not even understand latin scripture, therefore will have problems filling that captcha anyway.
3. We don't want to make them give us any identifying things like phone numbers or mail addresses to keep them safe.

**Describe the solution you'd like**

In order to sort out spam bots on our servers, we like our client apps to send an API key
on registration via HTTP headers.

We're not interested in supporting other clients than our own, but we want to stay federated.

I'll send a PR with the code changes we need to make this happen.

**Describe alternatives you've considered**

We'd really hate to maintain patches on our own.

Other workarounds would need to make use of reflection (aka Objective-C-KVO) and a bunch of code duplication.

Another idea would be to add proprietary steps to the registration process. However, it seems, as pluggable, as the API design looks, Synapse isn't actually that pluggable in that regard.

Considering the imminent threat with spam bots, HTTP headers seem like a way quicker and more viable solution.

tladesignz:
### Pull Request Checklist

### DIFF
--- a/MatrixSDK/Contrib/Swift/MXRestClient.swift
+++ b/MatrixSDK/Contrib/Swift/MXRestClient.swift
@@ -102,13 +102,14 @@ public extension MXRestClient {
      
      - parameters:
          - username: The user name to test.
+         - headers: Optional HTTP headers to add to the request.
          - completion: A block object called when the operation is completed.
          - inUse: Whether the username is in use
      
      - returns: a `MXHTTPOperation` instance.
      */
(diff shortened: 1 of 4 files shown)


tladesignz:
#1184

tladesignz:
As discussed in issue #1184, this is superfluous. There's already similar functionality available, it was just hard to find.

PR #1191, opened 2021-08-10
VoIP: Add Logs for Recurring Calls

ismailgulek:
Only adds logs for vector-im/element-ios#4687. That's why i didn't add a changelog.


stefanceriu (approved):
👍

PR #1200, opened 2021-08-13
Doug/4484 contacts access

pixlwave:
Part of https://github.com/vector-im/element-ios/issues/4484.

Adds an `areAllTermsAgreed` property to `MXIdentityService` which is updated by `MXSession` as the account data changes and cached in the store.


stefanceriu (approved):
LGTM! 👍

PR #1220, opened 2021-09-07
Add MXURLPreview and a previewForURL method to MXRestClient

pixlwave:
Part of https://github.com/vector-im/element-ios/pull/4790.

`MXURLPreview` only focuses on the core OpenGraph fields for now (there are potentially more in the response). Image type, width, height and size aren't actively used in the above PR, but seem useful to keep in here anyway.


MatrixSDK/JSONModels/Media/MXURLPreview.m
```diff
@@ -0,0 +1,121 @@
...
+@implementation MXURLPreview
+
++ (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
+{
+    MXURLPreview *urlPreview = [MXURLPreview new];
+    
+    NSString *siteName, *title, *description, *imageURL, *imageType;
+    NSNumber *imageWidth, *imageHeight, *imageFileSize;
+    
+    MXJSONModelSetString(siteName, JSONDictionary[@"og:site_name"]);
```
SBiOSoftWhare, line 28: Do not hesitate to create constants when you use a key twice. Like `static NSString * const kSiteNameJSONKey = @"og:site_name";`

MatrixSDK/MXRestClient.h
```diff
@@ -1951,6 +1952,21 @@ Get the maximum size a media upload can be in bytes.
...
+ 
+Note: Clients should consider avoiding this endpoint for URLs posted in encrypted rooms.
+ 
+@param url The URL to get the preview data for.
+@param success A block object called when the operation succeeds. It provides an `MXURLPreview` object for the requested URL.
+@param failure A block object called when the operation fails.
+
+@return a MXHTTPOperation instance.
+*/
+- (MXHTTPOperation*) previewForURL:(NSURL*)url
```
SBiOSoftWhare, line 1966:
```suggestion
- (MXHTTPOperation*)previewForURL:(NSURL*)url
```

MatrixSDK/JSONModels/Media/MXURLPreview.m
pixlwave: Good point, that's much better. Are these good, kept private to the implementation?

SBiOSoftWhare (approved):

PR #1230, opened 2021-09-20
[Spaces] pre release fixes

gileluard:
- Added `MXSpaceNotificationCounter`

### DIFF
--- a/MatrixSDK/Space/MXSpaceNotificationCounter.swift
+++ b/MatrixSDK/Space/MXSpaceNotificationCounter.swift
@@ -1,23 +1,41 @@
 //
-//  MXSpaceNotificationCounter.swift
-//  MatrixSDK
+// Copyright 2021 The Matrix.org Foundation C.I.C
 //
-//  Created by Gil Eluard on 19/09/2021.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 //
 
 import Foundation
 
+/// MXSpaceNotificationState stores the categorized number of unread messages
 @objcMembers
 public class MXSpaceNotificationState: NSObject {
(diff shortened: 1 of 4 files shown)


MatrixSDK/Space/MXSpaceNotificationCounter.swift
```diff
@@ -0,0 +1,160 @@
+//
+//  MXSpaceNotificationCounter.swift
+//  MatrixSDK
+//
+//  Created by Gil Eluard on 19/09/2021.
+//
```
SBiOSoftWhare: This is weird that default project header was not used here.

MatrixSDK/Space/MXSpaceNotificationCounter.swift
```diff
@@ -0,0 +1,160 @@
...
+
+import Foundation
+
+@objcMembers
+public class MXSpaceNotificationState: NSObject {
```
SBiOSoftWhare, line 21: Do not hesitate to create separate file for this one.

MatrixSDK/Space/MXSpaceNotificationCounter.swift
```diff
@@ -0,0 +1,160 @@
...
+        left = left + right
+    }
+}
+
+@objcMembers
+public class MXSpaceNotificationCounter: NSObject {
+    
+    // MARK: - Properties
+    
+    public var homeNotificationState = MXSpaceNotificationState()
```
SBiOSoftWhare, line 65: Is this property can be modified from the outside? Otherwise can make the setter private:
```suggestion
    public private(set) var homeNotificationState = MXSpaceNotificationState()
```

MatrixSDK/Space/MXSpaceNotificationCounter.swift
```diff
@@ -0,0 +1,160 @@
...
+        
+        MXLog.debug("[Spaces] computeNotificationCount ended after \(Date().timeIntervalSince(startDate))s")
+    }
+    
+    /// Notification state for a given space
+    /// - Parameters:
+    ///   - spaceId: ID of the space
+    /// - Returns: a `MXSpaceNotificationState` instance with the number of notifications for the given space
+    public func notificationState(forSpaceWithId spaceId: String) -> MXSpaceNotificationState {
+        return notificationStatePerSpaceId[spaceId] ?? MXSpaceNotificationState()
```
SBiOSoftWhare, line 109: Should we create in this case the instance in the dictionary? Because we are returning an instance that will not be the same the next time it will be in `notificationStatePerSpaceId ` dict.

SBiOSoftWhare (approved):

PR #1232, opened 2021-09-23
Fixes #4717 - Renaming DM rooms to [User Name](Left) after the only o…

stefanceriu:
…ther participant leaves.


pixlwave (approved):

pixlwave:
We had a call to discuss the path without room heroes as users that left won't be included there. However this doesn't seem to be touched when the other user leaves the room.

PR #1234, opened 2021-09-24
avoid reentrance

gileluard:
Avoid reentrance in space graph builder

### DIFF
--- a/MatrixSDK/Space/MXSpaceService.swift
+++ b/MatrixSDK/Space/MXSpaceService.swift
@@ -73,6 +73,8 @@ public class MXSpaceService: NSObject {
     private var orphanedRooms: [MXRoom] = []
     private var orphanedDirectRooms: [MXRoom] = []
     
+    private var isGraphBuilding = false;
+    
     public let notificationCounter = MXSpaceNotificationCounter()
     
     public private(set) var rootSpaceSummaries: [MXRoomSummary] = []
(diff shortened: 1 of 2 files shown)


MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -101,52 +103,63 @@ public class MXSpaceService: NSObject {
...
-            MXLog.debug("\(spaces), \(spacesPerId), \(roomsPerId), \(directRooms)")
-            var parentIdsPerRoomId: [String : Set<String>] = [:]
-            spaces.forEach { space in
-                space.updateChildSpaces(with: spacesPerId)
-                space.updateChildDirectRooms(with: directRooms)
-                space.childRoomIds.forEach { roomId in
-                    var parentIds = parentIdsPerRoomId[roomId] ?? Set<String>()
-                    parentIds.insert(space.spaceId)
-                    parentIdsPerRoomId[roomId] = parentIds
+        guard !self.isGraphBuilding else {
```
SBiOSoftWhare, line 106: If new rooms `[RoomC, RoomD]` came in the sync but we are building graph for other rooms `[RoomA, RoomB]`, the method returns, but it will never build the graph with `[RoomC, RoomD]` until new rooms related events trigger this method.
An option can be to use `Operation` class or just a pending rooms array. 
* When you receive new rooms you build the graph with them. 
* If new rooms come in between while you are building the graph you create a new operation or change `pendingRooms` value. You made this until the current graph building is done (cancel the last waiting operation or update the pending rooms).
* When the graph build is done you take the last pending operation or pending rooms array and you made the build graph. 
* It's not needed but you can use a debounce timer when you receive new rooms or when the graph building ends.

SBiOSoftWhare (approved):

PR #1240, opened 2021-09-27
Simplify last message types

pixlwave:
Part of https://github.com/vector-im/element-ios/issues/4451.

• Renames eventsFilterForMessages into eventsFilterForLastMessage.
• Removes ignoreMemberProfileChanges as this option can actually come from eventsFilterForLastMessage.

These changes in the SDK are not technically necessary but do make the API clearer. The downside is they change the API, so whilst I think they're helpful I can understand if we should leave this as is.


stefanceriu (approved):
lgtm 👍

stefanceriu (approved):

PR #1243, opened 2021-09-28
vector-im/element-ios/issues/4899 - Added dynamism and compile time s…

stefanceriu:
…afety to room name and send reply event localizable strings.


pixlwave (commented):
Looks good to me. Maybe the changelog should have a `.api` entry to mention any breaking changes that other users of the SDK would face?

Small thought on the protocol naming: Localizable makes more sense to me as it aligns with other protocol naming like Identifiable, Hashable, Codable etc and itself isn't actually a Localizer.

stefanceriu:
We've had a separate chat an agreed to use the Protocol suffix.

changelog.d/4899.api
```diff
@@ -0,0 +1 @@
+Added dynamism and compile time safety to room name and send reply event localizable strings.
```
pixlwave, line 1:
```suggestion
`MXSendReplyEventStringsLocalizable` is now `MXSendReplyEventStringLocalizerProtocol` and `MXRoomNameStringsLocalizable` is now `MXRoomNameStringLocalizerProtocol`
```

pixlwave (approved):

PR #1246, opened 2021-09-29
MXSpaceService: Fix a crash due to recursion depth limit

manuroe:
Fix https://github.com/vector-im/element-ios/issues/4919.


gileluard (approved):
LGTM

PR #1250, opened 2021-10-04
Fix performance issues with Spaces on very large accounts #4898

gileluard:
Needed for https://github.com/vector-im/element-ios/issues/4898

- Moved space graph data structure into separate class
- Cache space graph data structure

### DIFF
--- a/MatrixSDK/Space/MXSpaceService.swift
+++ b/MatrixSDK/Space/MXSpaceService.swift
@@ -58,31 +58,34 @@ public class MXSpaceService: NSObject {
     private let processingQueue: DispatchQueue
     private let completionQueue: DispatchQueue
     
-    private var spaces: [MXSpace] = []
-    private var spacesPerId: [String : MXSpace] = [:]
-    private var parentIdsPerRoomId: [String : Set<String>] = [:]
-    private var flattenedParentIds: [String: Set<String>] = [:]
-
-    private var rootSpaces: [MXSpace] = [] {
+    private var graph: MXSpaceGraphData = MXSpaceGraphData() {
         didSet {
-            self.rootSpaceSummaries = rootSpaces.compactMap { space in
-                return space.summary
+            self.graph.spaceRoomIds.forEach { spaceId in
+                if let space = self.getSpace(withId: spaceId) {
+                    self.spacesPerId[spaceId] = space
+                }
             }
         }
     }
-    private var orphanedRooms: [MXRoom] = []
-    private var orphanedDirectRooms: [MXRoom] = []
+    private var spacesPerId: [String:MXSpace] = [:]
     
     private var isGraphBuilding = false;
     
     public let notificationCounter = MXSpaceNotificationCounter()
     
-    public private(set) var rootSpaceSummaries: [MXRoomSummary] = []
+    public var rootSpaceSummaries: [MXRoomSummary] {
+        return self.graph.rootSpaceIds.compactMap { spaceId in
+            self.session.roomSummary(withRoomId: spaceId)
+        }
+    }
     
     public private(set) var needsUpdate: Bool = true
     
     public var graphUpdateEnabled = true
     
+    private var sessionStateDidChangeObserver: Any?
+    private var sessionDidSyncObserver: Any?
+
     // MARK: - Setup
     
     public init(session: MXSession) {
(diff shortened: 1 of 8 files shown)


MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -317,18 +367,54 @@ public class MXSpaceService: NSObject {
...
+        self.sessionStateDidChangeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name.mxSessionStateDidChange, object: session, queue: nil) { [weak self] notification in
+            guard let session = self?.session, session.state == .storeDataReady, let store = self?.session.store else {
+                return
+            }
+            
+            self?.loadData(from: store)
+        }
+        
+        self.sessionDidSyncObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name.mxSessionDidSync, object: session, queue: nil) { [weak self] notification in
+            guard let self = self, let syncResponse = notification.userInfo?[kMXSessionNotificationSyncResponseKey] as? MXSyncResponse else {
```
manuroe, line 380: Note: if you need to call `buildGraph` in an accurate time during the processing of the sync response, you can make a call from `MXSession.handleSyncResponse()` like before. The execution order will be more controlled than via a notification.
But it seems that the space service has its own life cycle with no strong dependency with the MXSession life cycle. So, notification is fine here.

MatrixSDK/Space/MXSpaceStore.swift
```diff
@@ -0,0 +1,124 @@
...
+    }
+    
+    // MARK: - Properties
+
+    private let userId: String
+    private var storeUrl: URL?
+
+    // MARK - Setup
+    
+    init(session: MXSession) {
```
manuroe, line 37: Please use userId/deviceId instead of a dependency on MXSession.
DeviceId is important because of integration tests that can run 2 devices for the same user in parallel.

MatrixSDK/Space/MXSpaceStore.swift
```diff
@@ -0,0 +1,124 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+/// `MXSpaceStore` is used to store the spaces related data into a permanent store
+class MXSpaceStore {
```
manuroe, line 20: Can we have a protocol for it?
MXFileSpaceStore will be this implementation.

MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -162,71 +183,42 @@ public class MXSpaceService: NSObject {
...
+        self.prepareData(with: roomIds, index: 0, output: output) { result in
+            while !output.computingSpaces.isEmpty {
+                sleep(1)
             }
             
-            // TODO improve updateNotificationsCount and call the method to all spaces once subspaces will be supported
-            self.notificationCounter.computeNotificationCount(for: rootSpaces, with: rooms, ancestorsPerRoomId: ancestorsPerRoomId)
-
-            MXLog.debug("[Spaces] buildGraph ended after \(Date().timeIntervalSince(startDate))s")
+            MXLog.debug("[Spaces] data prepared in \(Date().timeIntervalSince(startDate))")
```
manuroe, line 201: Can we use the common formatting for log, ie `MXLog.debug("[MXSpaceService] buildGraph: Data prepared ...")`?
Can you apply it everywhere, please? I just noticed it.

manuroe (commented):
Thanks. The threading model is good now. The new algo looks good to me.

MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -37,6 +37,9 @@ extension MXSpaceServiceError: CustomNSError {
 
 // MARK: - MXSpaceService notification constants
 extension MXSpaceService {
+    /// Posted once the first graph as been built or loaded
+    public static let didInitialised = Notification.Name("MXSpaceServiceDidInitialised")
```
manuroe, line 41: Is it normal that is not used by the app?
I expected to see it in the EI [PR](https://github.com/vector-im/element-ios/pull/4961).

gileluard: It will be used for https://github.com/vector-im/element-ios/issues/4570

MatrixSDK/Space/MXSpaceFileStore.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    // MARK: - Properties
+
+    private let userId: String
+    private var storeUrl: URL?
+
+    // MARK - Setup
+    
+    init(userId: String) {
```
manuroe, line 37: We also need deviceId to avoid clashes in integration tests.

manuroe (approved):

PR #1257, opened 2021-10-11
Crash in MXSpaceService.prepareData() #4979

gileluard:
fix https://github.com/vector-im/element-ios/issues/4979


MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -372,7 +377,7 @@ public class MXSpaceService: NSObject {
         
         func setComputing(_ isComputing: Bool, forSpace space: MXSpace) {
             self.serialQueue.sync {
-                if isComputing {
+                if _isComputing {
```
gileluard, line 380: this one should be `isComputing` as it is a parameter of the method.

PR #1260, opened 2021-10-13
Avoid calling SDK synchroniously in spaces classes #4999

gileluard:
- Fixed

### DIFF
--- a/MatrixSDK/Space/MXSpaceService.swift
+++ b/MatrixSDK/Space/MXSpaceService.swift
@@ -454,24 +454,31 @@ public class MXSpaceService: NSObject {
             }
             
             var _room: MXRoom?
-            var _space: MXSpace?
+            var space: MXSpace?
             var isRoomDirect = false
-            var _directUserId: String?
-            self.sdkProcessingQueue.sync {
+            var directUserId: String?
+            self.sdkProcessingQueue.async {
                 _room = self.session.room(withRoomId: roomIds[index])
                 
                 if let room = _room {
-                    _space = self.spacesPerId[room.roomId] ?? room.toSpace()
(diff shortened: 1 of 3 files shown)


MatrixSDK/Space/MXSpaceService.swift
```diff
@@ -454,24 +454,31 @@ public class MXSpaceService: NSObject {
...
+                self.prepareData(with: roomIds, index: index, output: output, room: room, space: space, isRoomDirect: isRoomDirect, directUserId: directUserId, completion: completion)
             }
-            
-            guard let room = _room else {
-                self.prepareData(with: roomIds, index: index+1, output: output, completion: completion)
-                return
-            }
-            
+        }
+    }
```
ismailgulek, line 477: You could simplify the method like:

```
private func prepareData(with roomIds:[String], index: Int, output: PrepareDataResult, completion: @escaping (_ result: PrepareDataResult) -> Void) {
    self.processingQueue.async {
        guard index < roomIds.count else {
            self.completionQueue.async {
                output.isPreparingData = false
                if !output.isComputing {
                    completion(output)
                }
            }
            return
        }
        
        self.sdkProcessingQueue.async {
            guard let room = self.session.room(withRoomId: roomIds[index]) else {
                self.prepareData(with: roomIds, index: index+1, output: output, completion: completion)
                return
            }
            
            let space = self.spacesPerId[room.roomId] ?? room.toSpace()
            
            self.prepareData(with: roomIds, index: index, output: output, room: room, space: space, isRoomDirect: room.isDirect, directUserId: room.directUserId, completion: completion)
        }
    }
}
```

ismailgulek (approved):

PR #1279, opened 2021-10-29
Client permalinks

langleyd:
To accompany https://github.com/vector-im/element-ios/issues/4981 and https://github.com/vector-im/element-ios/pull/5060


giomfo (approved):
LGTM

PR #1281, opened 2021-10-29
Initialize imagesCacheLruCache before caching

ferologics:
Caching operations would fail silently because cache was not initialized.


ismailgulek (approved):

ismailgulek:
Thanks @ferologics for this contribution.

PR #1283, opened 2021-11-03
Fix room previews

pixlwave:
Required by https://github.com/vector-im/element-ios/pull/5087, fixes https://github.com/vector-im/element-ios/issues/5083.

### DIFF
--- a/MatrixSDK/Data/RoomList/MXRoomListDataFilterOptions.swift
+++ b/MatrixSDK/Data/RoomList/MXRoomListDataFilterOptions.swift
@@ -40,6 +40,10 @@ public struct MXRoomListDataFilterOptions: Equatable {
     /// Flag to filter only suggested rooms, if set to `true`, `dataTypes` and `notDataTypes` are not valid.
     public let onlySuggested: Bool
     
+    /// Flag to hide any rooms where the user's membership is unknown. This has no effect when `onlySuggested` is `true`.
+    /// When set to `false`, rooms that have been cached during peeking may be included in the filtered results.
+    public let hideUnknownMembershipRooms: Bool
+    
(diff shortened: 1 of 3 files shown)


MatrixSDK/MXSession.h
```diff
@@ -503,6 +503,16 @@ FOUNDATION_EXPORT NSString *const kMXSessionNoRoomTag;
...
 
+/**
+ Whether the user is part of a room with the membership state of `join` or
+ they are in the process of joining.
+
+ @param roomIdOrAlias The ID or alias of the room to check.
+
+ @return YES if they are.
+ */
+- (BOOL)isJoinedOnRoom:(NSString *)roomIdOrAlias;
```
ismailgulek, line 514: Doesn't have an implementation.

MatrixSDK/Data/RoomList/MXRoomListDataFilterOptions.swift
```diff
@@ -121,6 +121,11 @@ public final class MXRoomListDataFilterOptions: NSObject {
         }
         
         if !onlySuggested {
+            // Filter out any room previews by making sure the user has valid membership.
+            let memberPredicate = NSPredicate(format: "%K != %d",
```
ismailgulek, line 125: I would prefer to make this configurable with a property like `hideUnknownMembershipRooms` with a default value of `true`, as i don't like the idea of having such logic in this class.

MatrixSDK/MXSession.h
pixlwave: Ah.... This is exposing an existing private function 🙂

ismailgulek (approved):

Issue #1282, opened 2021-11-03
Realm DB migration fails from version 12 -> version 17


N-Pex:
**Describe the bug**
SDK will crash when migrating from schema < 13 to schema > 16.

**To Reproduce**
1. Install product based on schema 12
2. Update product to newer SDK with Realm schema > 16
3. Crash

**Expected behavior**
Successful migration

I'll submit a PR for this shortly.

PR #1284, opened 2021-11-03
Fix migration from schema <= 12 to schema > 16. Issue #1282

N-Pex:
Signed-off-by: N-Pex <npexdev@gmail.com>

### Pull Request Checklist


pixlwave:
Thanks for the PR @N-Pex and sorry for the long delay in reviewing. This looks good to me, although there is one small tweak we could make:
- It would make more sense to wrap the entire `enumerateObjects` call in the new `if` block, as there is no need to run this step if it's not going to perform any actions.

PR #1286, opened 2021-11-04
Don't allow upgraded rooms that are hidden to become unhidden again.

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/5080

When a room has been upgraded with the device in the background, if a limited sync response comes in for that room a call to `MXRoomSummary`'s `resetRoomStateData` method is made. This recomputes the room state including the room's `create` event which was unhiding the room immediately after it had been hidden.

The PR fixes that and bumps the file store version to force all room state to be recomputed.

### DIFF
--- a/MatrixSDK/Data/MXRoomSummaryUpdater.m
+++ b/MatrixSDK/Data/MXRoomSummaryUpdater.m
@@ -297,7 +297,7 @@
         {
             BOOL isReplacementRoomJoined = replacementRoomSummary.membership == MXMembershipJoin;
                         
-            if (isReplacementRoomJoined && summary.hiddenFromUser == NO)
+            if (isReplacementRoomJoined)
             {
                 summary.hiddenFromUser = YES;
                 updated = YES;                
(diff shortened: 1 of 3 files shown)


ismailgulek (approved):

MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -285,8 +295,13 @@ - (BOOL)checkForTombStoneStateEventAndUpdateRoomSummaryIfNeeded:(MXRoomSummary*)
         
         if (replacementRoomSummary)
         {
-            summary.hiddenFromUser = replacementRoomSummary.membership == MXMembershipJoin;
-            updated = YES;
+            BOOL isReplacementRoomJoined = replacementRoomSummary.membership == MXMembershipJoin;
+                        
+            if (isReplacementRoomJoined && summary.hiddenFromUser == NO)
```
pixlwave, line 300: Now there's a check above this could become
```suggestion
            if (isReplacementRoomJoined)
```

SBiOSoftWhare (approved):

PR #1288, opened 2021-11-08
Fix Reaction counts

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/3884.

Most of the calls to `-[MXFileStore deleteAllData]` occur from within `MXFileStore` and happen before the aggregations have been set on the MXSession:
```
openWithCredentials:onComplete:failure:
|---checkStorageValidity
|   \---loadMetaData		(also called from MXBackgroundStore.init)
|
|---preloadRoomsMessages
```

This PR adds an `MXStoreService` class and adds a `storeService` property to `MXSession` for its store. An `MXStore` can then signal to this service that other stores should be reset before it deletes its own data. Only MXFileStore actually implements this, and attempting to do so from another store type will fail. The service will remember which stores it attempted to reset but weren't available, and will reset them as soon as they're set.

For now the service only encapsulates the MXStore and MXAggregations, however it is generic enough to extend it for other store types in the future.


MatrixSDK/Data/Store/MXStoreHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@objcMembers
+public class MXStoreHelper: NSObject {
```
pixlwave, line 20: Open to better suggestions for the name of this or whether there is a better class for the implementation to live in.

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -382,6 +383,16 @@ - (void)deleteRoom:(NSString *)roomId
...
+    
+    if (self.clearSecondaryStoresCallback)
+    {
+        self.clearSecondaryStoresCallback();
+        [MXStoreHelper secondaryStoresWereClearedFor:credentials.userId];
+    }
+    else
+    {
+        [MXStoreHelper setSecondaryStoresToBeClearedFor:credentials.userId];
+    }
```
pixlwave, line 395: Hmmm just realised that we seem to have multiple instances of the file store at times. So whilst the data needs to be deleted, I'm not sure if this will be helpful with the timing 😕

MatrixSDK/Data/Store/MXStoreService.swift
```diff
@@ -0,0 +1,141 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@objcMembers
+public class MXStoreService: NSObject {
```
SBiOSoftWhare, line 20: We can add a small description

MatrixSDK/Data/Store/MXStore.h
```diff
@@ -31,13 +31,21 @@
...
  Matrix data handled during a `MXSession`.
  */
 @protocol MXStore <NSObject, MXRoomSummaryStore>
 
+#pragma mark - Store Management
+
+/**
+ The store service that is managing this store.
+ */
+@property (nonatomic, weak, nullable) MXStoreService *storeService;
```
SBiOSoftWhare, line 47: It can be interesting to remove this dependency. Probably in another PR. `MXStoreService` will orchestrate all data belong to a user from the outside. 
Maybe we should add callback or local notification on the `MXStore` that `MXStoreService` listen to.

MatrixSDK/Data/Store/MXStoreService.swift
```diff
@@ -0,0 +1,141 @@
...
+        
+        return userStoreTypes.compactMap { StoreType(rawValue: $0) }
+    }
+    
+    /// Update the store types that need to be reset when loaded for the specified user ID.
+    /// - Parameters:
+    ///   - storeTypes: The store types that should be reset.
+    ///   - userId: The user ID to store the types for.
+    private func updateStoresToReset(_ storeTypes: [StoreType], for userId: String) {
+        var allStores = defaults.object(forKey: Constants.storesToResetKey) as? [String: [String]] ?? [:]
```
SBiOSoftWhare, line 136: maybe `allStoresToReset` instead of `allStores allStores`

MatrixSDK/Data/Store/MXStoreService.swift
pixlwave: Oh yep, my bad I missed this!

MatrixSDK/Data/Store/MXStore.h
pixlwave: Agreed.

SBiOSoftWhare (approved):

PR #1289, opened 2021-11-10
Fix room ordering

SBiOSoftWhare:
Part of fix for https://github.com/vector-im/element-ios/issues/5105

### DIFF
--- a/MatrixSDK/Data/MXRoomSummaryUpdater.m
+++ b/MatrixSDK/Data/MXRoomSummaryUpdater.m
@@ -111,17 +111,15 @@
     BOOL updated = NO;
 
     // Accept event which type is in the filter list
-    if (event.eventId && (_lastMessageEventTypesAllowList == nil || [_lastMessageEventTypesAllowList containsObject:event.type]))
+    // Don't accept event related to profile change
+    // TODO: Add a flag if needed to configure
+    // TODO: Only accept membership changes from current user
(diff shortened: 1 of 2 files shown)


MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -113,8 +113,15 @@ - (BOOL)session:(MXSession *)session updateRoomSummary:(MXRoomSummary *)summary
...
-        updated = YES;
+        // Don't accept event related to profile change
+        // TODO: Add a flag if needed to configure this
+        // TODO: Only accept membership changes from current user
+        if (event.eventType == MXEventTypeRoomMember 
+            && !event.isUserProfileChange)
+        {
+            [summary updateLastMessage:[[MXRoomLastMessage alloc] initWithEvent:event]];
+            updated = YES;        
+        }
```
pixlwave, line 124: This currently prevents the last message from updating for the allowed types (such as messages). The logic should probably be flipped like so:
```suggestion
        if (event.eventType == MXEventTypeRoomMember && event.isUserProfileChange)
        {
            return NO;
        }
        
        [summary updateLastMessage:[[MXRoomLastMessage alloc] initWithEvent:event]];
        updated = YES;
```

pixlwave (approved):
Much cleaner 😅

PR #1290, opened 2021-11-11
Fixes vector-im/element-ios/issues/5120 - Fixed fallback key signature validation.

stefanceriu:
Storing full MXKey dictionary so we can use all fields when checking the signature.
Added integration test to exercise currently failing paths.

### DIFF
--- a/MatrixSDK/Crypto/Data/MXKey.m
+++ b/MatrixSDK/Crypto/Data/MXKey.m
@@ -20,6 +20,16 @@ NSString *const kMXKeyCurve25519Type = @"curve25519";
 NSString *const kMXKeySignedCurve25519Type = @"signed_curve25519";
 NSString *const kMXKeyEd25519Type = @"ed25519";
 
+@interface MXKey()
+
+/**
+ * We have to store the original key dictionary because it can contain other fields
+ * that we don't support yet but they would be needed to check signatures
+ */
+@property (nonatomic, strong) NSDictionary *signalableJSONDictionary;
+
+@end
+
 @implementation MXKey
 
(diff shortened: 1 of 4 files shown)


MatrixSDK/Crypto/Data/MXKey.m
```diff
@@ -85,22 +101,24 @@ + (id)modelFromJSON:(NSDictionary *)JSONDictionary
...
     }
     return JSONDictionary;
 }
 
 - (NSDictionary *)signalableJSONDictionary
 {
-    return @{
-             @"key": _value
-             };
+    if (_signalableJSONDictionary == nil) {
```
manuroe, line 117:
```suggestion
    if (_signalableJSONDictionary == nil)
    {
```

MatrixSDKTests/MXCryptoTests.m
```diff
@@ -3007,6 +3009,52 @@ - (void)testDiscardAndRestoreOlmOutboundKey
     }];
 }
 
+#pragma mark - One time / fallback keys
+
+- (void)testFallbackKeySignatures
+{
+    [MXSDKOptions sharedInstance].enableCryptoWhenStartingMXSession = YES;
```
manuroe, line 3016: We need to reset it to `NO` after the MXSession creation to avoid to affect other tests. Yeah, singleton...

manuroe (commented):
The code looks good to me.
@BillCarsonFr, can you double check that keys are aligned now?

MatrixSDKTests/MXCryptoTests.m
stefanceriu: Ha, good one 😅

manuroe (approved):

BillCarsonFr (approved):

PR #1291, opened 2021-11-11
Group highlights/notifications/unread rooms instead of sorting.

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/5105.

Highlight/notification/unread counts were being used to sort rooms instead of grouping them.


ismailgulek:
Will be replaced by another solution.

ShadowJonathan:
What solution?

pixlwave:
This one: https://github.com/matrix-org/matrix-ios-sdk/pull/1292

PR #1292, opened 2021-11-11
Ignore number of highlights, notifications and unreads when sorting rooms

ismailgulek:
Part of vector-im/element-ios#5105

Ignores number of highlights, notifications and unreadCounts. Sorts according to only existence of them.


ismailgulek:
Simulator Screen Shot - iPhone 11 Pro - 2021-11-11 at 19 48 55

pixlwave (approved):
Cool, looks good to me. Sorry I thought the idea was to only do that on the Managed Object otherwise I would've done it :)

PR #1294, opened 2021-11-12
Fix not resuming session

ismailgulek:
Fixes vector-im/element-ios#5127

This PR does two things:
- Allows a session to resume also from `paused` state. Actually this should fix the issue.
- Avoids a catchup request (sync request with a `serverTimeout` value `0`) being made after the session `paused` or `pauseRequested`. This is for the sake of the application state. As we cannot guarantee that a catchup request will get a response shortly, we don't want to make this request in a backgrounded app.


stefanceriu (approved):
Great catch 👍

PR #1295, opened 2021-11-15
Poll support

stefanceriu:
* added MXRoom poll event sending methods
* added poll aggregator with integration tests
* added poll model builder with unit tests
* changed the MacOS testing target to not use optimizations
* changed MXAggregations limit parameters to be `NSIntegers` instead of `NSUIntegers` as the limit can be `-1`


MatrixSDK/Room/Polls/PollBuilder.swift
```diff
@@ -0,0 +1,112 @@
...
+        var answerOptionIdentifiers = [String]()
+        poll.answerOptions = pollStartEventContent.answerOptions.map { answerOption in
+            answerOptionIdentifiers.append(answerOption.uuid)
+            return Poll.AnswerOption(id: answerOption.uuid, text: answerOption.text)
+        }
+        
+        let stopEvent = events.filter { $0.type == kMXEventTypeStringPollEnd }.first
+        if stopEvent != nil {
+            poll.isClosed = true
+        }
```
SBiOSoftWhare, line 37:
```suggestion
        poll.isClosed = stopEvent != nil
```

MatrixSDK/Room/Polls/PollModels.swift
```diff
@@ -0,0 +1,49 @@
...
+        var isCurrentUserSelection: Bool = false
+        
+        init(id: String, text: String) {
+            self.id = id
+            self.text = text
+        }
+    }
+    
+    var text: String!
+    var answerOptions: [AnswerOption]!
```
SBiOSoftWhare, line 40:
```suggestion
    var answerOptions: [AnswerOption] = []
```

MatrixSDK/Room/Polls/PollModels.swift
```diff
@@ -0,0 +1,49 @@
...
+        var isWinner: Bool = false
+        var isCurrentUserSelection: Bool = false
+        
+        init(id: String, text: String) {
+            self.id = id
+            self.text = text
+        }
+    }
+    
+    var text: String!
```
SBiOSoftWhare, line 39: If we are not sure that the variable will be set we should avoid to use unwrap optional. In this case use optional or default value
```suggestion
    var text: String = ""
```

MatrixSDK/Room/Polls/PollModels.swift
```diff
@@ -0,0 +1,49 @@
...
+
+class Poll {
+    enum Kind {
+        case disclosed
+        case undisclosed
+    }
+    
+    class AnswerOption {
+        var id: String!
+        var text: String!
```
SBiOSoftWhare, line 27: Here the properties are set with initializer we should not use unwrap optional or var
```suggestion
        let id: String
        let text: String
```

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+    
+    var delegate: PollAggregatorDelegate?
+    
+    deinit {
+        room.removeListener(eventListener)
+    }
+    
+    init(session: MXSession, room: MXRoom, pollStartEvent: MXEvent) throws {
```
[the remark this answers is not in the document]
stefanceriu: I actually had the same question but it turns out a throwing initializer gives you the best of both worlds as you can use it as `if let aggregator = try? PollAggregator` but also within a try/catch block, and the reason it fails is clear if you're interested in it.

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -0,0 +1,115 @@
...
+
+class PollAggregator {
+    
+    private let session: MXSession
+    private let room: MXRoom
+    private let pollStartEvent: MXEvent
+    private let pollStartEventContent: MXEventContentPollStart
+    private let pollBuilder: PollBuilder
+    
+    private var eventListener: Any!
```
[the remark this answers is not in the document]
stefanceriu: What I really wanted was to have them all as `let non-optional` properties but also be able to construct it without having to pass everything in the initializer, so I settled on these for unwraps that would crash on runtime if any are unset.
In hindsight that wasn't the best choice, I'll figure out another way to construct them properly in the builder and have a clear interface.

MatrixSDK/Room/Polls/PollModels.swift
stefanceriu: Same as above ⏫

MatrixSDK/Room/Polls/PollModels.swift
stefanceriu: Same as above ⏫

MatrixSDK/Room/Polls/PollModels.swift
stefanceriu: Same as above ⏫

MatrixSDK/Room/Polls/PollAggregator.swift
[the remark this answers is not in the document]
stefanceriu: Decided to introduce an immutable protocol that just exposes everything as readonly while the builder uses the real class with all properties variable.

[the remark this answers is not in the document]
SBiOSoftWhare: Yes sure no problem we can improve that in another PR if needed. But as we don't throw errors when the parsing fails, it's probably still safer to avoid unwrap optionals if we are not sure that the property will not be set before used as it will make the app crashing.

MatrixSDK/Room/Polls/PollAggregator.swift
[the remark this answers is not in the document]
SBiOSoftWhare: Yes good point, as long as we don't need ObjC compatibility. It's interesting to throw an error especially if you can have several ones.

SBiOSoftWhare (approved):

PR #1299, opened 2021-11-18
Thread List

ismailgulek:
Part of vector-im/element-ios#5092

Will be based on `develop` branch when https://github.com/matrix-org/matrix-ios-sdk/pull/1285 is merged.


MatrixSDK/Contrib/Swift/MXRestClient.swift
```diff
@@ -1027,6 +1027,32 @@ public extension MXRestClient {
...
     }
     
+    /**
+     Get list of threads for a room.
+     
+     - parameters:
+        - roomId: the id of the room.
+        - from: the token to start getting results from.
+        - direction: `MXTimelineDirectionForwards` or `MXTimelineDirectionBackwards`
+        - limit: (optional, use -1 to not defined this value) the maximum nuber of messages to return.
```
pixlwave, line 1037:
```suggestion
        - limit: (optional, use -1 to not defined this value) the maximum number of messages to return.
```

MatrixSDK/Contrib/Swift/MXRestClient.swift
```diff
@@ -1027,6 +1027,32 @@ public extension MXRestClient {
...
+        - filter: to filter returned events with.
+        - completion: A block object called when the operation completes.
+        - response: Provides a `MXPaginationResponse` object on success.
+     
+     - returns: a MXHTTPOperation instance.
+     */
+    @nonobjc @discardableResult func threads(forRoom roomId: String, from: String, direction: MXTimelineDirection, limit: UInt?, filter: MXRoomEventFilter, completion: @escaping (_ response: MXResponse<MXPaginationResponse>) -> Void) -> MXHTTPOperation {
```
pixlwave, line 1044: Would this be slightly clearer to read(same in MXRestClient.h/m too)? (not important though).
```suggestion
    @nonobjc @discardableResult func threads(forRoom roomId: String, from startToken: String, direction: MXTimelineDirection, limit: UInt?, filter: MXRoomEventFilter, completion: @escaping (_ response: MXResponse<MXPaginationResponse>) -> Void) -> MXHTTPOperation {
```

MatrixSDK/Threads/MXThreadingService.swift
```diff
@@ -100,35 +107,57 @@ public class MXThreadingService: NSObject {
         return result
     }
     
+    /// Get threads in a room
+    /// - Parameter roomId: room identifier
+    /// - Returns: thread list in given room
+    public func threads(inRoom roomId: String) -> [MXThread] {
+        //  sort threads so that the newer is the first
+        return Array(threads.values).filter({ $0.roomId == roomId }).sorted(by: <)
```
pixlwave, line 115: I'm not sure how to achieve this (and it's probably unnecessary early optimisation), but it would be kind of nice if this had a `.lazy` to improve performance in the second `filter` in `participatedThreads`.

MatrixSDK/MXRestClient.m
```diff
@@ -2357,6 +2357,62 @@ - (MXHTTPOperation*)messagesForRoom:(NSString*)roomId
                                  }];
 }
 
+- (MXHTTPOperation*)threadsForRoom:(NSString*)roomId
```
pixlwave, line 2360: Thinking from the perspective of someone else using the library, if this is called with a `nil` filter then is it going to return the same messages as `messagesForRoom`? So should this be adding the `relationTypes` to the filter (creating if needed) for you? That's if I have interpreted this correctly.

MatrixSDK/Threads/MXThread.swift
```diff
@@ -59,16 +55,24 @@ public class MXThread: NSObject {
...
+    public var isParticipated: Bool {
+        guard let session = session else {
+            return false
         }
+        return eventsMap.values.first(where: { $0.sender == session.myUserId }) != nil
+    }
+    
+    /// Root message of the thread
+    public var rootMessage: MXEvent? {
+        return eventsMap[identifier]
```
pixlwave, line 70: Much nicer 🙌

pixlwave (commented):
Just a couple of comments. I'm still not sure on the sort order, but seeing how this chains onto the last PR (and your comment there), probably makes sense to look at that in the next one.

MatrixSDK/Threads/MXThreadingService.swift
ismailgulek: Ah that doesn't seem trivial to me, let's handle it later.

MatrixSDK/Contrib/Swift/MXRestClient.swift
ismailgulek: Method removed.

MatrixSDK/Contrib/Swift/MXRestClient.swift
ismailgulek: Method removed.

MatrixSDK/MXRestClient.m
ismailgulek: Method removed.

MatrixSDK/Threads/MXThread.swift
ismailgulek: 👍

ismailgulek:
Sorry i removed `MXRestClient.threadsForRoom` method for now. I'll add it in a later PR, when we really use it.

pixlwave (approved):
In which case, looks good to me 👍

PR #1301, opened 2021-11-23
Add redacted root events into threads

ismailgulek:
Part of vector-im/element-ios#5119

Will be based on `develop` after other thread PRs merged.


MatrixSDK/Threads/MXThread.swift
```diff
@@ -139,6 +139,9 @@ extension MXEvent: Comparable {
     
     public static func < (lhs: MXEvent, rhs: MXEvent) -> Bool {
         //  event will be 'smaller' than an other event if it's newer
+        if lhs.originServerTs != NSNotFound && rhs.originServerTs != NSNotFound {
+            return lhs.compareOriginServerTs(rhs) == .orderedAscending
+        }
         return lhs.age < rhs.age
```
pixlwave, line 145: Should this now be reversed as timestamp ascending is the same as age descending?
```suggestion
        return lhs.age > rhs.age
```

ismailgulek: It was correct but i agree it was confusing, code updated and documented to be clearer hopefully.

pixlwave (approved):

PR #1303, opened 2021-11-25
#5181 Phone is abnormally warm, and battery is drained faster than usual

gileluard:
fix https://github.com/vector-im/element-ios/issues/5181


manuroe (approved):
LGTM. This is a good explanation of the increasing time :)

PR #1309, opened 2021-12-01
Fix for the in-call screen freezing on a new PSTN call

langleyd:
Resolves https://github.com/vector-im/element-ios/issues/5223

Incoming PSTN calls can now have rooms automatically created and joined server-side.

If the incoming voip call is from a number you have not seen before(or if you have since left the room) and the app is in the terminated state, when you receive the voip push you may not have created a room for the given roomID yet.

In the same way as we exposed `MXSession.getOrCreateUser` for use in calling for such cases a user had not been seen, I've followed the same pattern for rooms, creating the room if does not exist.

Previously `MXCall.room` was null causing an invalid initialization of the call flow.

### DIFF
--- a/MatrixSDK/MXSession.h
+++ b/MatrixSDK/MXSession.h
@@ -923,7 +923,17 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
 
  @return the MXRoom instance.
  */
-- (MXRoom  *)roomWithRoomId:(NSString*)roomId;
+- (MXRoom *)roomWithRoomId:(NSString*)roomId;
+
+/**
+ Get the MXRoom instance of a room.
+ Create it if does not exist yet.
+ 
+ @param roomId The id to the user.
+ 
+ @return the MXRoom instance.
+ */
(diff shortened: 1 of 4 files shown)


MatrixSDK/MXSession.h
```diff
@@ -923,7 +923,17 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
 
  @return the MXRoom instance.
  */
-- (MXRoom  *)roomWithRoomId:(NSString*)roomId;
+- (MXRoom *)roomWithRoomId:(NSString*)roomId;
+
+/**
+ Get the MXRoom instance of a room.
+ Create it if does not exist yet.
```
ismailgulek, line 930: Can we just improve this doc to point out the room won't be created in Matrix layer? It might be misinterpreted by SDK users.

ismailgulek, line 930:
```suggestion
 Create it if does not exist yet. The room will be created locally if needed, won't have any effect on the home server.
```

ismailgulek (approved):

PR #1317, opened 2021-12-07
Thread Notifications

ismailgulek:
Part of vector-im/element-ios#5096


MatrixSDK/Threads/MXThread.swift
```diff
@@ -29,7 +29,14 @@ public class MXThread: NSObject {
     /// Identifier of the room that the thread is in.
     public let roomId: String
     
+    private var _liveTimeline: MXEventTimeline?
```
SBiOSoftWhare, line 32: Why using `_` prefix? seems that `liveTimeline` can be used here.

MatrixSDK/Threads/MXThread.swift
```diff
@@ -49,14 +56,39 @@ public class MXThread: NSObject {
...
-            return
-        }
+    /// Add an event to the thread
+    /// - Parameter event: Event
+    /// - Returns: true if handled, false otherwise
+    @discardableResult
+    internal func addEvent(_ event: MXEvent) -> Bool {
         eventsMap[event.eventId] = event
+        updateNotificationsCount()
+        if event.sender == session?.myUserId {
```
SBiOSoftWhare, line 66: If `event.sender` is nil and `session` is nil? It will be marked as read. Sure we are missing nullability on ObjC :/

MatrixSDK/Threads/MXThread.swift
ismailgulek: You right, redacted.

MatrixSDK/Threads/MXThread.swift
ismailgulek: I've added checks for that.

SBiOSoftWhare (approved):

PR #1318, opened 2021-12-07
Update MXAnalyticsDelegate and MXTaskProfile

pixlwave:
As part of https://github.com/vector-im/element-ios/issues/5035, this PR makes 2 changes.

- `MXAnalyticsDelegate` now provides specific methods per event reported as the new [matrix-analytics-events](https://github.com/matrix-org/matrix-analytics-events) schema should be followed more strictly.
- `MXTaskProfile` has been updated to use a single `MXTaskProfileName` enum instead of separate `name` and `category` strings in order to switch on the task name.

### DIFF
--- a/MatrixSDK/Utils/Profiling/MXTaskProfile.h
+++ b/MatrixSDK/Utils/Profiling/MXTaskProfile.h
@@ -18,13 +18,33 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NSString *const MXTaskProfileCategory NS_TYPED_EXTENSIBLE_ENUM;
+// Timing stats relative to app startup.
+static MXTaskProfileCategory const MXTaskProfileCategoryStartup = @"startup";
+// Metrics related to the initial sync request
+static MXTaskProfileCategory const MXTaskProfileCategoryInitialSync = @"initialSync";
+
+typedef NSString *const MXTaskProfileName NS_TYPED_EXTENSIBLE_ENUM;
+// Duration of the initial /sync request
+static MXTaskProfileName const MXTaskProfileNameStartupInitialSync = @"initialSync";
+// Duration of the first /sync when resuming the app
+static MXTaskProfileName const MXTaskProfileNameStartupIncrementalSync = @"incrementalSync";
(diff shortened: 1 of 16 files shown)


MatrixSDK/Utils/Profiling/MXTaskProfile.h
```diff
@@ -18,13 +18,28 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NSString *const MXTaskProfileName NS_TYPED_EXTENSIBLE_ENUM;
```
SBiOSoftWhare, line 21: Another possibility:
```
extern const struct MXTaskProfileName {
    const struct Startup {
        NSString* __unsafe_unretained _nonnull_ Category; // Category name
        NSString* __unsafe_unretained _nonnull_ InitialSync;
        NSString* __unsafe_unretained _nonnull_ IncrementalSync;
    } Startup;    
    const struct PushNotifications {
        NSString* __unsafe_unretained _nonnull_ Category; // Category name
        NSString* __unsafe_unretained _nonnull_ OpenEvent;
    } PushNotifications;
} MXTaskProfileName;
```

SBiOSoftWhare, line 21: For readability and as `MXTaskProfileName` can grow, I think it worth to create an `MXTaskProfileName.h` and `.m`

pixlwave: In terms of using the above, I've had a quick look and I get as far as:

```objc
- (MXTaskProfile *)startMeasuringTaskWithName:(struct MXTaskProfileName)name;
```
```objc
MXTaskProfile *taskProfile = [MXSDKOptions.sharedInstance.profiler startMeasuringTaskWithName:MXTaskProfileName.Startup.InitialSync];
```
However then it complains 
> Sending 'NSString *__unsafe_unretained const __unsafe_unretained' to parameter of incompatible type 'struct MXTaskProfileName'

which I assume means that `startMeasuringTaskWithName` would then need to accept a plain string again?

PR #1320, opened 2021-12-10
Add missing `#import "TargetConditionals.h"` in `MXCallKitAdapter`

ferologics:
Add missing `#import "TargetConditionals.h"` in `MXCallKitAdapter`


danpe:
@stefanceriu what about this one? :) 
For us it was mandatory to be able to build

PR #1328, opened 2021-12-16
invite to space in room landing

gileluard:
Needed for https://github.com/vector-im/element-ios/issues/5225


niquewoodhouse:
@gileluard for the missing permissions:

image

I don't feel _that_ strongly about this so happy to go with the proposal detailed for testing and see what feedback we get.
My only concerns were:

- The disabled states for buttons on inviting:
  - I don't think it's consistent with the other buttons (eg on the left two screens I can tap a non-disabled FAB to get an alert, while in the other screens the options are disabled, and I don't know why (I can assume I don't have permission but not 100% sure).

gileluard:
@niquewoodhouse 

We have here 2 possibilities:
- We don't disable menu buttons. Downside: the user will tap on the Fab then the menu to see that he doesn't have permission to do that
- We disable the menu buttons. In this case the description of the menu item is changed for `You do not have permission...`. Downside: It may not be obvious enough

Just let me know what you prefer or if you have a better idea.

niquewoodhouse:
> * We don't disable menu buttons. Downside: the user will tap on the Fab then the menu to see that he doesn't have permission to do that

Out of these two options, this one makes the most sense, as it's consistent with the other logic in the app, which makes sense.

My original idea was if you don't have permission to invite to the space, it just directs the user straight to the invite to room option (as it saves them an interaction) but that's not how I think android has done it, so it won't be consistent.

gileluard:
> My original idea was if you don't have permission to invite to the space, it just directs the user straight to the invite to room option (as it saves them an interaction) but that's not how I think android has done it, so it won't be consistent.

I can do that but looks similar as disabling a option as the user won't know why in some cases the menu is displayed and other cases he won't.

pixlwave (approved):

pixlwave:
@gileluard Could this fix be updated to target `develop` seeing as it affects other SDK users as well?

PR #1331, opened 2021-12-23
Location sharing

stefanceriu:
Various location sharing support changes + code improvements.

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -805,7 +805,7 @@ NSInteger const kMXRoomAlreadyJoinedErrorCode = 9001;
         // This is a simple text message
         msgContent = @{
             kMXMessageTypeKey: kMXMessageTypeText,
-            @"body": text
+            kMXMessageBodyKey: text
         };
     }
     else
@@ -813,7 +813,7 @@ NSInteger const kMXRoomAlreadyJoinedErrorCode = 9001;
         // Send the HTML formatted string
         msgContent = @{
             kMXMessageTypeKey: kMXMessageTypeText,
-            @"body": text,
+            kMXMessageBodyKey: text,
             @"formatted_body": formattedText,
             @"format": kMXRoomMessageFormatHTML
         };
@@ -845,7 +845,7 @@ NSInteger const kMXRoomAlreadyJoinedErrorCode = 9001;
         // This is a simple text message
         msgContent = @{
             kMXMessageTypeKey: kMXMessageTypeEmote,
-            @"body": emoteBody
+            kMXMessageBodyKey: emoteBody
         };
     }
     else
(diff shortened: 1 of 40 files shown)


MatrixSDK/Data/MXRoom.m
```diff
@@ -2380,6 +2389,40 @@ - (MXHTTPOperation *)sendPollEndForEvent:(MXEvent *)pollStartEvent
...
+    NSMutableDictionary *content = [NSMutableDictionary dictionary];
+    content[kMXMessageTypeKey] = kMXMessageTypeLocation;
+    
+    NSDictionary *locationContent = [[MXEventContentLocation alloc] initWithLatitude:latitude
+                                                                           longitude:longitude
+                                                                         description:description].JSONDictionary;
+    
+    content[kMXMessageContentKeyExtensibleLocationMSC3488] = locationContent;
+    
+    NSString *geoURI = locationContent[kMXMessageContentKeyExtensibleLocationURI];
```
SBiOSoftWhare, line 2410: Does it make sense to create a `geoURI` property or method on `MXEventContentLocation` instead?

MatrixSDK/JSONModels/MXEvent.m
```diff
@@ -472,12 +486,30 @@ - (BOOL)isReplyEvent
...
 - (BOOL)isVoiceMessage
 {
-    NSString *msgtype = self.content[@"msgtype"];
+    NSString *msgtype = self.content[kMXMessageTypeKey];
     return [msgtype isEqualToString:kMXMessageTypeAudio] && (self.content[kMXMessageContentKeyVoiceMessage] ||
                                                              self.content[kMXMessageContentKeyVoiceMessageMSC2516] ||
                                                              self.content[kMXMessageContentKeyVoiceMessageMSC3245]);
 }
 
+- (MXEventContentLocation *)location
```
SBiOSoftWhare, line 495: Normally we are more adding `- (BOOL)isLocation` kind of method on `MXEvent` rather than building another kind of event from the current one.
Otherwise is it possible to handle:
```
    NSDictionary *locationContent = self.content[kMXMessageContentKeyExtensibleLocationMSC3488];
     if (locationContent == nil) {
         locationContent = self.content[kMXMessageContentKeyExtensibleLocation];
     }
```
directly in `MXEventContentLocation modelFromJSON:`?

MatrixSDK/Data/MXRoom.m
stefanceriu: Yep, good idea 👍

MatrixSDK/JSONModels/MXEvent.m
stefanceriu: I was unsure if I should pass the whole event payload in there but you're right, it does make things cleaner 👍

SBiOSoftWhare (approved):

PR #1335, opened 2022-01-12
Fix DTMF

langleyd:
Fixes https://github.com/vector-im/element-ios/issues/5375
with https://github.com/vector-im/element-ios/pull/5376

I decided to match the implementation of android:
- We don't make use of the additional parameters and it complicates the logic a bit, so simplified the signature.
- android loops though the senders to get the dtmf sender which seems slightly more robust
- Android just hardcodes the duration and gap, copied the values.
- The input is NSTimeInterval which is seconds rather than ms. This differs from the android webrtc api which passes as ms.


ismailgulek (approved):

ismailgulek:
We could add a changelog.

PR #1337, opened 2022-01-13
Poll editing

stefanceriu:
Enabled poll start event editing and added `hasBeenEdited` on the PollAggregator.

### DIFF
--- a/MatrixSDK/JSONModels/MXEvent.h
+++ b/MatrixSDK/JSONModels/MXEvent.h
@@ -211,6 +211,7 @@ FOUNDATION_EXPORT NSString *const kMXEventRelationRelatesToKey;
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeAnnotation;    // Reactions
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeReference;     // Reply
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeReplace;       // Edition
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyNewContent;   // Edited content key
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeThread;        // Thread
 
 /**
--- a/MatrixSDK/JSONModels/MXEvent.h
+++ b/MatrixSDK/JSONModels/MXEvent.h
@@ -223,15 +223,20 @@ FOUNDATION_EXPORT NSString *const kMXEventLocalEventIdPrefix;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyVoiceMessage;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyVoiceMessageMSC2516;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyVoiceMessageMSC3245;
+
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleAudio;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleAudioMSC1767;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleAudioDuration;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleAudioWaveform;
 
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleText;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleTextMSC1767;
 
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleTimestamp;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleTimestampMSC3488;
 
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFile;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFileMSC1767;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFileSize;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFileName;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFileURL;
--- a/MatrixSDK/JSONModels/MXEvent.h
+++ b/MatrixSDK/JSONModels/MXEvent.h
@@ -239,8 +239,14 @@ FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensibleFileMimeType;
 
 // Polls
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollStart;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollStartMSC3381;
+
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollResponse;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollResponseMSC3381;
+
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollEnd;
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollEndMSC3381;
+
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollQuestion;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollAnswers;
 FOUNDATION_EXPORT NSString *const kMXMessageContentKeyExtensiblePollAnswerId;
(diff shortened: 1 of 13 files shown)


MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -38,12 +38,16 @@ public class PollAggregator {
...
-    private let pollStartEvent: MXEvent
-    private let pollStartEventContent: MXEventContentPollStart
+    private let pollStartEventId: String
     private let pollBuilder: PollBuilder
     
-    private var eventListener: Any!
+    private var pollStartEventContent: MXEventContentPollStart!
+    
+    private var referenceEventsListener: Any!
+    private var editEventsListener: Any!
```
pixlwave, line 47: As `referenceEventsListener` is set in init, can it be `: Any`?
Side note: I'd also wonder if `editEventsListener` could actually `Any?` but presumably as `eventListener` was force unwrapped there's more to this elsewhere that is unchanged in this PR.

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -54,40 +58,65 @@ public class PollAggregator {
...
+            } catch {
+                self.delegate?.pollAggregator(self, didFailWithError: PollAggregatorError.invalidPollStartEvent)
+            }
+        }
+        
+        try buildPollStartContent()
+    }
+    
+    private func buildPollStartContent() throws {
+        guard let event = self.session.store.event(withEventId: self.pollStartEventId, inRoom: room.roomId),
```
pixlwave, line 91: How about a few less `self`s? 😉
```suggestion
        guard let event = session.store.event(withEventId: pollStartEventId, inRoom: room.roomId),
```

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -54,40 +58,65 @@ public class PollAggregator {
...
+              let pollStartEventContent = MXEventContentPollStart(fromJSON: event.content) else {
             throw PollAggregatorError.invalidPollStartEvent
         }
         
         self.pollStartEventContent = pollStartEventContent
         
-        pollBuilder = PollBuilder()
+        self.hasBeenEdited = (event.unsignedData.relations?.replace != nil)
```
pixlwave, line 98:
```suggestion
        hasBeenEdited = (event.unsignedData.relations?.replace != nil)
```

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -54,40 +58,65 @@ public class PollAggregator {
...
         
-        poll = pollBuilder.build(pollStartEventContent: self.pollStartEventContent, events: self.events, currentUserIdentifier: session.myUserId)
+        self.poll = self.pollBuilder.build(pollStartEventContent: pollStartEventContent,
```
pixlwave, line 100: I'll stop adding these everywhere 🙃
```suggestion
        poll = pollBuilder.build(pollStartEventContent: pollStartEventContent,
```

MatrixSDK/JSONModels/MXEvent.m
```diff
@@ -668,20 +673,16 @@ - (MXEvent*)editedEventFromReplacementEvent:(MXEvent*)replaceEvent
...
-    else if (event.content[kMXMessageBodyKey] && newContentDict && [newContentDict[kMXMessageTypeKey] isEqualToString:event.content[kMXMessageTypeKey]])
+    else if (newContentDict)
     {
         editedEventDict = [event.JSONDictionary mutableCopy];
-        NSMutableDictionary *editedEventContentDict = [editedEventDict[@"content"] mutableCopy];
-        editedEventContentDict[kMXMessageBodyKey] = newContentDict[kMXMessageBodyKey];
-        editedEventContentDict[@"formatted_body"] = newContentDict[@"formatted_body"];
-        editedEventContentDict[@"format"] = newContentDict[@"format"];
-        editedEventDict[@"content"] = editedEventContentDict;
+        editedEventDict[@"content"] = newContentDict;
```
pixlwave, line 679: If I edit an event that is a reply, I wonder if this change would wipe out the reply relationship on the original? There might be other examples that should be preserved too.

MatrixSDK/JSONModels/MXEvent.m
```diff
@@ -649,6 +650,10 @@ - (MXEvent*)prune
 
 - (MXEvent*)editedEventFromReplacementEvent:(MXEvent*)replaceEvent
 {
+    if ([self.eventId isEqualToString:replaceEvent.eventId]) {
```
pixlwave, line 653: Does that actually happen? I thought it wouldn't be possible to send an event that replaces itself as the server generates the event ID after you've sent it.

pixlwave (commented):
Some comments added in line.

MatrixSDK/JSONModels/MXEvent.h
```diff
@@ -211,6 +211,7 @@ FOUNDATION_EXPORT NSString *const kMXEventRelationRelatesToKey;
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeAnnotation;    // Reactions
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeReference;     // Reply
 FOUNDATION_EXPORT NSString *const MXEventRelationTypeReplace;       // Edition
+FOUNDATION_EXPORT NSString *const kMXMessageContentKeyNewContent;   // Edited content key
```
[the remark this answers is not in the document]
stefanceriu: Not sure to be honest, relationTypes are the only ones left on `MXEvent`, everything else is `kMXEvent`. I only placed it near them because it's directly related to `m.replace`

MatrixSDK/JSONModels/MXEvent.m
```diff
@@ -476,7 +477,7 @@ - (BOOL)isMediaAttachment
 
 - (BOOL)isEditEvent
 {
-    return self.eventType == MXEventTypeRoomMessage && [self.relatesTo.relationType isEqualToString:MXEventRelationTypeReplace];
+    return [self.relatesTo.relationType isEqualToString:MXEventRelationTypeReplace];
```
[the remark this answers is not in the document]
stefanceriu: And if that's the case shouldn't `isEditEvent` return true for them as well?

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -54,40 +58,65 @@ public class PollAggregator {
     public var delegate: PollAggregatorDelegate?
     
     deinit {
-        room.removeListener(eventListener)
+        room.removeListener(referenceEventsListener)
+        session.aggregations.removeListener(editEventsListener as Any)
```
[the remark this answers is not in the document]
stefanceriu: Without it, apparently `Coercion of implicitly unwrappable value of type 'Any?' to 'Any' does not unwrap optional`
It's either cast to Any, provider a default value or force unwrap.

MatrixSDK/Room/Polls/PollAggregator.swift
stefanceriu: Yiis 👍

MatrixSDK/Room/Polls/PollAggregator.swift
stefanceriu: `referenceEventsListener` isn't set in init but what I can do is make them both optional and check in deinit, which would also take care of the `as Any` cast. I honestly can't remember why `referenceEventsListener` was forced unwrapped.

MatrixSDK/JSONModels/MXEvent.m
stefanceriu: I should actually remove this, it got left behind from when I was trying to refactor some things.

MatrixSDK/JSONModels/MXEvent.m
stefanceriu: Good catch, I'll switch it to just simply check for poll start key content as well. 👍

MatrixSDK/JSONModels/MXEvent.m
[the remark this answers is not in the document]
pixlwave: I think it be worth enquiring about this one with everyone else, to see if there was a reason for condition :)

MatrixSDK/JSONModels/MXEvent.h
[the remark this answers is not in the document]
pixlwave: Gotcha.

MatrixSDK/JSONModels/MXEvent.m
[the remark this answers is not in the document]
stefanceriu: Going to leave it like in the PR as per Manu's comment here https://matrix.to/#/!NMoyFpQxIVYaQIfQHG:matrix.org/$WLeCqIZKuXPNuHb3oe_ydJksfNiUzCjfbMiMNVdZ8j4?via=matrix.org&via=element.io&via=one.ems.host

pixlwave (approved):

pixlwave (approved):

PR #1341, opened 2022-01-21
🐛 Hangup call with MXCallHangupReasonInviteTimeout reason when invite expires

ferologics:
The iOS client MatrixSDK is using a 30 second timer that is initiated whenever a call invite is sent. Once it fires the state is set to ended, but only locally, and no `.callHangup` with inviteTimeout reason is sent. The PR corrects this to be in line with the [`m.call.hangup` event specification](https://spec.matrix.org/v1.1/client-server-api/#mcallhangup):
> `invite_timeout` for when the other party did not answer in time

### DIFF
--- a/MatrixSDK/VoIP/MXCall.m
+++ b/MatrixSDK/VoIP/MXCall.m
@@ -1672,25 +1672,12 @@ NSString *const kMXCallSupportsTransferringStatusDidChange = @"kMXCallSupportsTr
     
     if (inviteExpirationTimer)
     {
+        [inviteExpirationTimer invalidate];
         inviteExpirationTimer = nil;
+    }
 
-        if (!_isIncoming)
-        {
-            // Terminate the call at the stack level we initiated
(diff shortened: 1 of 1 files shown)


danpe:
@stefanceriu I believe we can also merge this as it matches the matrix spec?

MatrixSDK/VoIP/MXCall.m
```diff
@@ -1672,25 +1672,12 @@ - (void)expireCallInvite
...
+    }
 
-        if (!_isIncoming)
-        {
-            // Terminate the call at the stack level we initiated
-            [callStackCall end];
-        }
-
-        // Send the notif that the call expired to the app
-        [self setState:MXCallStateInviteExpired reason:nil];
```
ismailgulek, line 1684: We need to keep this state change before senfing hangup.

ismailgulek (changes requested):

PR #1345, opened 2022-01-25
Fix Room Summaries Storage Crash

ismailgulek:
Fixes vector-im/element-ios#5412

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -2882,11 +2882,6 @@ typedef void (^MXOnResumeDone)(void);
             [listener removeSpiedRoom:room];
         }
 
-        // Clean the store
-        [self.store deleteRoom:roomId];
-        [self.store.roomSummaryStore removeSummaryOfRoom:roomId];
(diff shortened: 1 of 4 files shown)


MatrixSDK/MXSession.m
```diff
@@ -2898,6 +2893,10 @@ - (void)removeRoom:(NSString *)roomId
                                                                      kMXSessionNotificationRoomIdKey: roomId
                                                                      }];
     }
+
+    // Clean the store
+    [self.aggregations resetDataInRoom:roomId];
+    [self.store deleteRoom:roomId];
```
manuroe, line 2899: Is it intentional to not call `[self.store.roomSummaryStore removeSummaryOfRoom:roomId];` anymore?
Is the room really removed from the store (and room lists)?

manuroe (commented):
It looks good to me. I just have one question.

MatrixSDK/MXSession.m
ismailgulek: Yes, it’s already called in `MXFileStore`.

manuroe (approved):

PR #1358, opened 2022-02-03
Defense against encryption state issues

BillCarsonFr:
Add some defensive coding when deciding to send encrypted or not in the room.
The decision is based only in the crypto store state (existence of a valid algorithm), instead of the room state.
The crypto store is more stable, only updated with supported alg and never reverted.

### Pull Request Checklist

### DIFF
--- a/MatrixSDK/Crypto/MXCrypto.m
+++ b/MatrixSDK/Crypto/MXCrypto.m
@@ -1881,7 +1881,7 @@ NSTimeInterval kMXCryptoMinForceSessionPeriod = 3600.0; // one hour
 - (BOOL)isRoomEncrypted:(NSString *)roomId
 {
 #ifdef MX_CRYPTO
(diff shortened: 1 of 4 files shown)


MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -1878,6 +1878,15 @@ - (BOOL)isBlacklistUnverifiedDevicesInRoom:(NSString *)roomId
 #endif
 }
 
+- (BOOL)isRoomEncrypted:(NSString *)roomId
+{
+#ifdef MX_CRYPTO
+    return [_store algorithmForRoom:roomId] != NULL;
```
manuroe, line 1884:
```suggestion
    return [_store algorithmForRoom:roomId] != nil;
```

manuroe (approved):
Thanks for it. LGTM.
I just have a suggestion to be more objc.

PR #1359, opened 2022-02-04
Fixes for Background Sync Cache Processing

ismailgulek:
Makes a couple of things:
- Avoids `event/(null)` requests and immediately fails them.
- Adds some error logs around `event/(null)` requests.
- Avoids session state back and forth when there is no cached sync responses.
- Adds another bg sync cache processing round in the session if new data received when processing an old one.

Part of vector-im/element-ios#5426

### DIFF
--- a/MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
+++ b/MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
@@ -183,6 +183,13 @@ public class MXSyncResponseFileStore: NSObject {
         metadata.outdatedSyncResponseIds.removeAll(where: { $0 == id })
         saveMetaData(metadata)
     }
+
+    private func deleteSyncResponseIds(ids: [String]) {
+        var metadata = readMetaData()
+        metadata.syncResponseIds.removeAll(where: { ids.contains($0) })
+        metadata.outdatedSyncResponseIds.removeAll(where: { ids.contains($0) })
+        saveMetaData(metadata)
+    }
 }
 
 //  MARK: - MXSyncResponseStore
@@ -218,6 +225,13 @@ extension MXSyncResponseFileStore: MXSyncResponseStore {
     public func deleteSyncResponse(withId id: String) {
         saveSyncResponse(path: syncResponsePath(withId: id), syncResponse: nil)
         deleteSyncResponseId(id: id)
+    }
+
+    public func deleteSyncResponses(withIds ids: [String]) {
+        for id in ids {
+            saveSyncResponse(path: syncResponsePath(withId: id), syncResponse: nil)
+        }
+        deleteSyncResponseIds(ids: ids)
     }
     
     public var syncResponseIds: [String] {
--- a/MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
+++ b/MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
@@ -178,10 +178,7 @@ public class MXSyncResponseFileStore: NSObject {
     }
     
     private func deleteSyncResponseId(id: String) {
-        var metadata = readMetaData()
-        metadata.syncResponseIds.removeAll(where: { $0 == id })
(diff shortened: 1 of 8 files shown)


MatrixSDK/MXSession.m
```diff
@@ -2902,6 +2936,18 @@ - (MXHTTPOperation*)eventWithEventId:(NSString*)eventId
                              success:(void (^)(MXEvent *event))success
                              failure:(void (^)(NSError *error))failure
 {
+    if (eventId == nil)
+    {
+        if (failure)
+        {
+            MXError *error = [[MXError alloc] initWithErrorCode:kMXErrCodeStringNotFound
```
manuroe, line 2943: Can we dump the call stack (`[NSThread callStackSymbols]`) to find the culprit more easily?

manuroe (commented):
LGTM. Only one comment.
 I added @Anderas as a reviewer because he is looking at a similar issue

MatrixSDK/MXSession.m
ismailgulek: Yeah i thought about it too, but later i’ve decided to add individual logs for every caller. Do you think we still need it?

Anderas:
> LGTM. Only one comment. I added @Anderas as a reviewer because he is looking at a similar issue

I tested this branch and the issue with cancelling HTTP request and subsequently pausing the session still occurs so I think it is unrelated. I will raise separate PR

MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
```diff
@@ -183,6 +183,13 @@ public class MXSyncResponseFileStore: NSObject {
         metadata.outdatedSyncResponseIds.removeAll(where: { $0 == id })
         saveMetaData(metadata)
     }
+
+    private func deleteSyncResponseIds(ids: [String]) {
```
Anderas, line 187: Could the `deleteSyncResponseId` (delete single ID), now internally call `deleteSyncResponseIds(ids: [id])` to avoid code duplication?

MatrixSDK/MXSession.m
```diff
@@ -926,18 +927,35 @@ - (NSString *)syncFilterId
 
 - (BOOL)isPauseable
 {
-    return _state == MXSessionStateSyncInProgress
-        || _state == MXSessionStateRunning
-        || _state == MXSessionStateBackgroundSyncInProgress
-        || _state == MXSessionStatePauseRequested;
+    switch (_state)
```
Anderas, line 930: Out of curiosity (beyond the scope of this PR), what are the conventions for using ivars directly as opposed to synthesised getters? (`_state` vs `self.state`)?

MatrixSDK/MXSession.m
```diff
@@ -926,18 +927,35 @@ - (NSString *)syncFilterId
...
 {
-    return !eventStreamRequest ||
-        (_state == MXSessionStateBackgroundSyncInProgress
-        || _state == MXSessionStatePauseRequested
-        || _state == MXSessionStatePaused);
+    if (!eventStreamRequest)
+    {
+        return YES;
+    }
+    switch (_state)
```
Anderas, line 950: To understand this bit, a session can only be resumed if it hasn't previously failed, right? In that case a session needs to first be reconnected?

Anderas (approved):

MatrixSDK/Background/Store/SyncResponseFileStore/MXSyncResponseFileStore.swift
ismailgulek: Yes surely

MatrixSDK/MXSession.m
ismailgulek: Well, depends on the failure. There are certain cases that session needs to be restarted (like `storeDataReady` and `initialSyncFailed`), for other failures we'll try to pause the session after them.

MatrixSDK/MXSession.m
ismailgulek: Tbh we don't have much conventions about them, for the cases we implemented synthesized getter or setters we're trying to ensure they're used instead of ivars. Otherwise there are mixed usage of them.

PR #1360, opened 2022-02-04
Do not pause MXSession if a sync request is cancelled

Anderas:
Fixes https://github.com/vector-im/element-ios/issues/5509

A sync request could be cancelled whilst the app is in foreground (e.g. a screen that triggers the sync is deallocated), but this cancellation should not be pausing `MXSession`. Other types of failure correctly re-trigger the sync after some timeout.

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1544,18 +1544,7 @@ typedef void (^MXOnResumeDone)(void);
         int32_t code = (int32_t)error.code;
 
         if ([error.domain isEqualToString:NSURLErrorDomain]
-            && code == kCFURLErrorCancelled)
-        {
-            MXLogDebug(@"[MXSession] The connection has been cancelled in state: %@", [MXTools readableSessionState:_state]);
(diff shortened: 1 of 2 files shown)


ismailgulek (approved):

MatrixSDK/MXSession.m
```diff
@@ -1544,18 +1544,7 @@ - (void)handleServerSyncError:(NSError*)error forRequestWithServerTimeout:(NSUIn
         int32_t code = (int32_t)error.code;
 
         if ([error.domain isEqualToString:NSURLErrorDomain]
-            && code == kCFURLErrorCancelled)
```
ismailgulek, line 1547: When we remove this if check, session goes into `homeServerNotReachable` state when we pull-to-refresh, let's just keep the check and do nothing (we could just keep the log line) inside the block.

Anderas: Makes sense, added the log back

ismailgulek (approved):

PR #1361, opened 2022-02-08
Exclude all files and directories from iCloud backup

Anderas:
Fixes https://github.com/vector-im/element-ios/issues/5498

As listed in [documentation](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup), the iOS is automatically backing up a number of folders to iCloud / iTunes, including `<AppHome>/Documents/` and `<AppHome>/Library/Application Support/`. Additionally the Matrix SDK stores files in `AppGroup` containers (potentially shared between app and an extension), which are also automatically backed up to iCloud. These folders include data for rooms, spaces etc.

To opt-out of backups completely, each url needs to be explicitely marked with `isExcludedFromBackupKey` (this is applied recursively). I converted all `createDirectory` calls with a custom method, which automatically does this when creating a new directory.

This does not yet address smaller files (e.g. User Defaults preferences only created on save) or any future folders, which might be created without the backup exclusion. To solve for this second problem I added two functions `excludeAllUserDirectoriesFromBackup` and `excludeAllAppGroupDirectoriesFromBackup`, which need to be called by the main application (e.g. in `AppDelegate`) and exclude all top-level directories.

Arguably this second approach might make it redundant to explicitly specify backup choice when creating a new folder, however the blanker exclusion is only called at specific moments (e.g. app start), which may be before a particular folder is created. This would expose the folder to backups until the next app launch.

### DIFF
--- a/MatrixSDK/Utils/Categories/FileManager+Backup.swift
+++ b/MatrixSDK/Utils/Categories/FileManager+Backup.swift
@@ -1,0 +1,87 @@
+import Foundation
+
+public extension FileManager {
+    /**
+     Create a directory at a url, including intermediate directories, which is excluded from iCloud / manual backups, no matter where it is located.
+     
+     Note: some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     see [details](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup).
+     */
+    @objc func createDirectoryExcludedFromBackup(at url: URL) throws {
+        try createDirectory(at: url, withIntermediateDirectories: true, attributes: nil)
+        try excludeFromBackup(url: url)
+    }
+    
+    /**
+     Create a directory at a path, including intermediate directories, which is excluded from iCloud / manual backups, no matter where it is located.
+     
+     Note: some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     see [details](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup).
+     */
+    @objc func createDirectoryExcludedFromBackup(atPath path: String) throws {
+        try createDirectoryExcludedFromBackup(at: URL(fileURLWithPath: path))
+    }
+    
+    /**
+     Exclude a given url from iCloud / manual backups, no matter where it is located.
+     
+     Note: some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     see [details](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup).
+     */
+    @objc func excludeFromBackup(url: URL) throws {
+        try (url as NSURL).setResourceValue(true, forKey: .isExcludedFromBackupKey)
+    }
+    
+    /**
+     Exclude all user directories from iCloud / manual backups.
+     
+     User directories include `<AppHome>/Documents`, `<AppHome>/Library`, `<AppHome>/Library/Application Support` etc.
+     Some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     */
+    func excludeAllUserDirectoriesFromBackup() {
+        // Only `Documents` and `Library` directories need to be specified explicitly. Other directories are either excluded
+        // automatically (`tmp`, `Library/Caches`), or are nested within one of the excluded ones (e.g. `Library/Application Support`)
+        let urls = [
+            urls(for: .documentDirectory, in: .userDomainMask),
+            urls(for: .libraryDirectory, in: .userDomainMask)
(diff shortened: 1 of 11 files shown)


MatrixSDK/Utils/Categories/FileManager+Backup.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+public extension FileManager {
+    /**
+     Create a directory at a url, including intermediate directories, which is excluded from iCloud / manual backups, no matter where it is located.
+     
+     Note: some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     see [details](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup).
+     */
+    @objc func createDirectoryExcludedFromBackup(at url: URL) throws {
```
Anderas, line 10: I considered an alternative api along the lines of `createDirectory(at:withIntermediate:attributes:isExcluded:)` but we just pass the same parameters everywhere anyway, so it felt cumbersome and pointless.

MatrixSDK/Utils/Categories/FileManager+Backup.swift
```diff
@@ -0,0 +1,103 @@
...
+        try createDirectoryExcludedFromBackup(at: URL(fileURLWithPath: path))
+    }
+    
+    /**
+     Exclude a given url from iCloud / manual backups, no matter where it is located.
+     
+     Note: some directories are excluded automatically if they are nested within `<AppHome>/Library/Caches/` or `<AppHome>/tmp/`
+     see [details](https://developer.apple.com/documentation/foundation/optimizing_app_data_for_icloud_backup).
+     */
+    @objc func excludeFromBackup(url: URL) throws {
```
stefanceriu, line 47: I would rename this to something like `excludeURLFromBackup(_ url: URL)` to be more in line with the others.

MatrixSDK/Utils/Categories/FileManager+Backup.swift
Anderas: This api distinction already exists within `FileManager`'s API (`(at url: URL)` vs `(atPath path: String)`). Reading [swift conventions](https://www.swift.org/documentation/api-design-guidelines/#parameter-names) I assume that `(at url: URL)` is preferable, but then `(at path: String)` could not be easily distinguished / would be overloading. Since `URL` is preferable type over `String` path, the url-api is given preferential treatment. I don't mind aligning them with your suggestion, just wanted to point out that this was a deliberate choice given the existing `FileManager` api.

Anderas:
> This all looks very reasonable to me, the only thing I'm worried about is this line from the documentation: `Set the key each time you save the file because certain common file operations clear the key.`. I don't think we have control over all the saves.

Yea this is true, though the way I understand it, it would mostly be an issue if we were setting the attribute on specific files/folders that are often updated and saved. Instead we are either marking top-level folders that we do not directly create (Documents, Library), or use the `createDirectoryExcludedFromBackup` to always pair mutation/creation with setting the attribute. And if that fails, we'll be calling the blanket exclusion from app delegate.

> We have quite a lot of folder within the application group container. Do we really need all of those shared with the other processes? 🤔

Very good question. I don't fully understand the integration with extensions yet, and would probably be a larger change, but worth taking that as the next logical step.

> Also, have you checked how the integration tests behave with these changes? (MXBackgroundSyncServiceTests, MXSessionTests, MXCoreDataRoomListDataManagerUnitTests etc.)

I have some issues running tests in general, even on `develop`, so need to debug that first.

MatrixSDK/Utils/Categories/FileManager+Backup.swift
stefanceriu: If it were my choice I would make them symmetrical but no strong feelings really.

Anderas: Yea after I changed the `excludeItemFromBackup` method, it seems preferable to stick with `(at url: URL)` everywhere for consistency sake.

Anderas:
> > Also, have you checked how the integration tests behave with these changes? (MXBackgroundSyncServiceTests, MXSessionTests, MXCoreDataRoomListDataManagerUnitTests etc.)
> 
> I have some issues running tests in general, even on `develop`, so need to debug that first.

So the integration tests (particularly `MXSessionTests`) seem quite flaky on `develop`, so in a future PR I can try to fix some of them, but I observed no difference in the failures between this PR and develop, so supposedly no regressions.

SBiOSoftWhare (approved):
Good job, this is well documented. Btw you can probably use `MXLog.error` instead of `MXLog.debug` for try/catch errors.

stefanceriu (approved):
Looks good, let's roll with it 👍

Anderas:
> Good job, this is well documented. Btw you can probably use `MXLog.error` instead of `MXLog.debug` for try/catch errors.

Ah yes, I thought I had. Thanks.

PR #1362, opened 2022-02-09
Fix add phone number validation through custom URL

aringenbach:
Fixes issue https://github.com/vector-im/element-ios/issues/3562

HttpClient was immediately released at the end of the block resulting in request cancelation.

### DIFF
--- a/MatrixSDK/ThreePidAdd/MX3PidAddManager.m
+++ b/MatrixSDK/ThreePidAdd/MX3PidAddManager.m
@@ -28,6 +28,11 @@ NSString *const MX3PidAddManagerErrorDomain = @"org.matrix.sdk.MX3PidAddManagerE
     BOOL doesServerSupportSeparateAddAndBind;
 }
 
+/**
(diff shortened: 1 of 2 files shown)


MatrixSDK/ThreePidAdd/MX3PidAddManager.m
```diff
@@ -28,6 +28,11 @@ @interface MX3PidAddManager()
     BOOL doesServerSupportSeparateAddAndBind;
 }
 
+/**
+ HTTP client dedicated to sending MSISDN token to custom URLs.
+ */
+@property (nonatomic, strong) MXHTTPClient *customUrlHttpClient;
```
stefanceriu, line 34: `customUrl` sounds very generic to me, we should call it something more in line with what it's actually used for e.g. `msisdnSubmissionHttpClient` ¯\_(ツ)_/¯

stefanceriu (changes requested):

stefanceriu (approved):
👍

PR #1363, opened 2022-02-09
Core Data Fix Parameterized Fetch Requests

ismailgulek:
Fix vector-im/element-ios#5519

> This property can be set with [managedObjectResultType](https://developer.apple.com/documentation/coredata/nsfetchrequestresulttype/1506452-managedobjectresulttype) and thereby implement a partial faulting (whereby only some of the properties are populated) of the returned objects, as well as the [dictionaryResultType](https://developer.apple.com/documentation/coredata/nsfetchrequestresulttype/1506237-dictionaryresulttype) to define what properties are included in the resulting [NSDictionary](https://developer.apple.com/documentation/foundation/nsdictionary).

Documentation for `NSFetchRequest.propertiesToFetch` says above, but `managedObjectResultType` it doesn't work properly on some devices, seems mostly on iOS 12.

Will also fix https://github.com/vector-im/element-ios/issues/5528

### DIFF
--- a/MatrixSDK/Data/RoomList/CoreData/MXCoreDataRoomListDataFetcher.swift
+++ b/MatrixSDK/Data/RoomList/CoreData/MXCoreDataRoomListDataFetcher.swift
@@ -444,10 +444,10 @@ private class RoomSummaryForTotalCounts: NSObject, MXRoomSummaryProtocol {
     /// Initializer with a dictionary obtained by a fetch request, with result type `.dictionaryResultType`. Only parses some properties.
     /// - Parameter dictionary: Dictionary object representing an `MXRoomSummaryMO` instance
     init(withDictionary dictionary: [String: Any]) {
-        dataTypes = MXRoomSummaryDataTypes(rawValue: dictionary[kPropertyNameDataTypesInt] as? Int ?? 0)
-        sentStatus = MXRoomSummarySentStatus(rawValue: dictionary[kPropertyNameSentStatusInt] as? UInt ?? 0) ?? .ok
-        notificationCount = dictionary[kPropertyNameNotificationCount] as? UInt ?? 0
-        highlightCount = dictionary[kPropertyNameHighlightCount] as? UInt ?? 0
+        dataTypes = MXRoomSummaryDataTypes(rawValue: Int(dictionary[kPropertyNameDataTypesInt] as? Int64 ?? 0))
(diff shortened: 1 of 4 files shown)


MatrixSDK/Data/RoomList/CoreData/MXCoreDataRoomListDataFetcher.swift
```diff
@@ -380,3 +389,70 @@ extension MXCoreDataRoomListDataFetcher: NSFetchedResultsControllerDelegate {
...
+    var spaceChildInfo: MXSpaceChildInfo?
+    var parentSpaceIds: Set<String> = []
+
+    /// Initializer with a dictionary obtained by a fetch request, with result type `.dictionaryResultType`. Only parses some properties.
+    /// - Parameter dictionary: Dictionary object representing an `MXRoomSummaryMO` instance
+    init(withDictionary dictionary: [String: Any]) {
+        dataTypes = MXRoomSummaryDataTypes(rawValue: dictionary[kPropertyNameDataTypesInt] as? Int ?? 0)
+        sentStatus = MXRoomSummarySentStatus(rawValue: dictionary[kPropertyNameSentStatusInt] as? UInt ?? 0) ?? .ok
+        notificationCount = dictionary[kPropertyNameNotificationCount] as? UInt ?? 0
+        highlightCount = dictionary[kPropertyNameHighlightCount] as? UInt ?? 0
```
pixlwave, line 450: All looks good to me. My only thought is whether these dictionary values should be cast to the same Int types that is used in the model definition?
e.g. `Int64` for `s_dataTypesInt` and `Int16` for `sentStatusInt`, `notificationCount` and `highlightCount`?

ismailgulek: Does not matter much in this case, updated anyway.

MatrixSDK/Data/RoomList/CoreData/MXCoreDataRoomListDataFetcher.swift
```diff
@@ -94,14 +99,18 @@ internal class MXCoreDataRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
             properties.append(property)
         }
         request.propertiesToFetch = properties
+        request.resultType = .dictionaryResultType
```
pixlwave, line 102: Thinking about it, might also be worth adding a comment here to say that it fixes issues that were seen on iOS 12 for future reference.

pixlwave (approved):
LGTM 👍

PR #1369, opened 2022-02-15
Fix e2ee regression introduced by #1358

manuroe:
Closes https://github.com/vector-im/element-ios/issues/5564.

The source of truth for the encryption of a room is back to `MXRoomSummary.isEncrypted`, which is a rollback of #1358.
The additional change of this PR is that on every send, there is a check if this value is aligned with the one in the crypto store.
There is also some error logs if something tries to reset `MXRoomSummary.isEncrypted`.


pixlwave (approved):
LGTM 👍

pixlwave (approved):
So much for my side-by-side review of the old PR and this one 🤦‍♂️
Works great with that last commit.

PR #1371, opened 2022-02-16
🐛 Fix `directRooms` array argument is not an NSArray exception

ferologics:
This exception would occur when event.content has exactly one direct room and MXJSONModelSetDictionary converts it to an NSString and not an NSArray.

Signed-off-by: Frantisek Hetes [f.hetes@gmail.com](mailto:f.hetes@gmail.com)


stefanceriu (changes requested):
Looks good to me but I think we need to do this whole dance right after the `MXJSONModelSetDictionary` and before the comparisons in order for them to work properly.

PR #1373, opened 2022-02-21
Fix room state for peeking

phlniji:
During a bug investigation, we found that the FileStore may be corrupted with peeking feature (when previewing a room), but not joining the room after previewing it.

In result, we had a « ghost » room in the room list, with some content, but not all the content of this room (ie: no room summary, but messages were visible, and we can pickup the previous messages when scrolling), instead of having nothing.

So I added a condition on Unknown membership rooms to avoid to add them to the file store.


stefanceriu (approved):
lgtm 👍

stefanceriu (approved):
lgtm 👍

PR #1375, opened 2022-02-22
Threads: Replace `/context` api

ismailgulek:
Part of vector-im/element-ios#5629

### DIFF
--- a/MatrixSDK/Data/EventTimeline/Thread/MXThreadEventTimeline.swift
+++ b/MatrixSDK/Data/EventTimeline/Thread/MXThreadEventTimeline.swift
@@ -16,6 +16,10 @@
 
 import Foundation
 
+enum MXThreadEventTimelineError: Error {
+    case initialEventNotFound
+}
+
 @objcMembers
 public class MXThreadEventTimeline: NSObject, MXEventTimeline {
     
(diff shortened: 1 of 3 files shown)


MatrixSDK/Data/EventTimeline/Thread/MXThreadEventTimeline.swift
```diff
@@ -152,58 +156,55 @@ public class MXThreadEventTimeline: NSObject, MXEventTimeline {
...
                 }
-            case .failure(let error):
-                MXLog.error("[MXThreadEventTimeline][\(self.timelineId)] resetPaginationAroundInitialEvent failed: \(error)")
-                failure(error)
+                dispatchGroup.leave()
+            }
+            result.mutate(to: newRequest)
+        }
+
+        paginationBlock()
```
pixlwave, line 194: I'm curious as to why use a block here seeing as it isn't stored and is called immediately after being created? Otherwise LGTM 👍

pixlwave (approved):

MatrixSDK/Data/EventTimeline/Thread/MXThreadEventTimeline.swift
ismailgulek: Because it's recursively getting next page, until the specified event found, so block used like a function here.

pixlwave: Oh, ignore me! I completely missed that call inside was recursive on itself and not on the method we are already in 🤦‍♂️

PR #1381, opened 2022-02-24
MXThreadingService: Add thread creation delegate method

ismailgulek:
Improve `MXThreadingServiceDelegate` by adding a new method for thread creation, with a `direction` property.

### DIFF
--- a/MatrixSDK/Threads/MXThreadingService.swift
+++ b/MatrixSDK/Threads/MXThreadingService.swift
@@ -37,7 +37,13 @@ extension MXThreadingServiceError: CustomNSError {
 
 @objc
 public protocol MXThreadingServiceDelegate: AnyObject {
-    func threadingServiceDidUpdateThreads(_ service: MXThreadingService)
+    /// Delegate method to be called when thread are updated in any way.
+    @objc optional func threadingServiceDidUpdateThreads(_ service: MXThreadingService)
+
+    /// Delegate method to be called when a new local thread is created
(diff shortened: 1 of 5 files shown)


MatrixSDK/Threads/MXThreadingService.swift
```diff
@@ -37,7 +37,13 @@ extension MXThreadingServiceError: CustomNSError {
 
 @objc
 public protocol MXThreadingServiceDelegate: AnyObject {
-    func threadingServiceDidUpdateThreads(_ service: MXThreadingService)
+    /// Delegate method to be called when thread are updated in any way.
+    @objc optional func threadingServiceDidUpdateThreads(_ service: MXThreadingService)
+
+    /// Delegate method to be called when a new local thread is created
+    @objc optional func threadingService(_ service: MXThreadingService,
```
Anderas, line 44: I can't quite wrap my head around the concept of `direction` for a newly created thread. Does `new` imply `empty`? And if so, how is `direction` used?

Anderas (approved):

MatrixSDK/Threads/MXThreadingService.swift
ismailgulek: It's actually the same logic with a message, either you get it live, or can see a new event coming when back paginating. A thread is also the same, either it's created by a live event, or it's _recreated_ by some old events when back paginating.

PR #1386, opened 2022-03-03
SDK: UISI AutoReporting

langleyd:
Accompanies app PR https://github.com/vector-im/element-ios/pull/5743

To address https://github.com/vector-im/element-ios/issues/5017


MatrixSDK/EventStream/MXEventStreamService.swift
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+/// Manages the adding, removing MXLiveEventListeners and dispatching of events to those listeners.
+@objcMembers public class MXEventStreamService: NSObject {
+    
+    var listeners = [MXLiveEventListener]()
```
gileluard, line 22: If `MXLiveEventListener` subclasses `NSObject` we could use then `NSHashTable<ObjectType>.weakObjectsHashTable()` and that could be even safer.

gileluard (commented):
LGTM. Just would like to have @manuroe approval for the breaking change and also you can check the potential approval.

MatrixSDK/Crypto/MXCrypto.h
```diff
@@ -264,12 +264,12 @@ extern NSString *const MXDeviceListDidUpdateUsersDevicesNotification;
...
 /**
- Handle a room key event.
+ Handle toDevice event
  
- @param event the room key event.
+ @param event the `toDevice` event.
  @param onComplete the block called when the operation completes.
  */
-- (void)handleRoomKeyEvent:(MXEvent*)event onComplete:(void (^)(void))onComplete;
+- (void)handleToDeviceEvent:(MXEvent*)event onComplete:(void (^)(void))onComplete;
```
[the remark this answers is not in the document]
langleyd: Just for a bit of context. The function becomes a bit more general, so not just handling `RoomKeyEvent` but handles toDeviceEvents generally. The code to handle `RoomKeyEvent` is unchanged. The naming also made some of the code read better like [here](https://github.com/matrix-org/matrix-ios-sdk/pull/1386/files#diff-c68a0a04ecb089a9fb727895d9bd9e85899d3c26f2fb321783b4ce30474d2129R1896). It also matches the a[ndroid implementation](https://github.com/vector-im/element-android/blob/ce4ad8819a920b022cfcd0bab2a4e1badee7ee75/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt#L803) now.

gileluard (approved):
LGTM 👍 Only thing is the breaking change in `MXCrypto`

MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -817,6 +818,24 @@ - (void)handleDeviceListsChanges:(MXDeviceListResponse*)deviceLists
...
+        {
+            [self handleRoomKeyEvent:event onComplete:onComplete];
+            break;
+        }
+            
+        default:
+            onComplete();
+            break;
+    }
+    [self.mxSession.eventStreamService dispatchOnLiveToDeviceWithEvent:event];
```
manuroe, line 835: Do we really need to have this dependency on MXSession here?
Can this block stay in  MXSession [here](https://github.com/matrix-org/matrix-ios-sdk/pull/1386/files#diff-480363b404ebe0778edd253e1f353db5b0476b367fc5f986440f7814ec79912eL1897)?

I cannot see this method used outside the SDK. The related app [PR](https://github.com/vector-im/element-ios/pull/5743) do not use it.

MatrixSDK/Crypto/MXCrypto.h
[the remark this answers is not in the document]
manuroe: I am not against the change of this API. Plus, it seems it is used only internally.
But I wonder if we really this method. See my review [comment](https://github.com/matrix-org/matrix-ios-sdk/pull/1386#discussion_r834523321).

MatrixSDK/Crypto/MXCrypto.m
langleyd: `mxSession` is already a hard dependency in crypto, i just re-used it.
Android handles `toDeviceEvents` in crypto like this and it made the caller site clearer in some instances as we were passing a `toDevice` variable.
But I really don't feel strong about it, so happy to move it back, Will create a new PR.

PR #1388, opened 2022-03-04
Hide suggested rooms that a user is already part of

gileluard:
fix https://github.com/vector-im/element-ios/issues/5276

### DIFF
--- a/MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
+++ b/MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
@@ -20,7 +20,7 @@ import Foundation
 internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
     
     internal let fetchOptions: MXRoomListDataFetchOptions
-    private let session: MXSession?
+    private weak var session: MXSession?
     private let spaceService: MXSpaceService
     private let cache: MXSuggestedRoomListDataCache
     
@@ -52,7 +52,7 @@ internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
     }
     
     internal init(fetchOptions: MXRoomListDataFetchOptions,
-                  session: MXSession?,
+                  session: MXSession,
                   spaceService: MXSpaceService,
                   cache: MXSuggestedRoomListDataCache = .shared) {
         self.fetchOptions = fetchOptions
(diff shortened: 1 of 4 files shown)


MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
```diff
@@ -20,9 +20,12 @@ import Foundation
 internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
     
     internal let fetchOptions: MXRoomListDataFetchOptions
+    private let session: MXSession?
```
ismailgulek, line 23: Can we make it a weak variable?

MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
```diff
@@ -48,9 +52,11 @@ internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
     }
     
     internal init(fetchOptions: MXRoomListDataFetchOptions,
+                  session: MXSession?,
```
ismailgulek, line 55: It doesn't make much sense to me to initialize with an optional session here.

MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
```diff
@@ -88,10 +94,16 @@ internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
             }
             self.refresh()
         })
+        sessionDidSyncObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name.mxSessionDidSync, object: nil, queue: OperationQueue.main) { [weak self] notification in
```
ismailgulek, line 97:
```suggestion
        sessionDidSyncObserver = NotificationCenter.default.addObserver(forName: .mxSessionDidSync, object: nil, queue: OperationQueue.main) { [weak self] notification in
```

MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
```diff
@@ -88,10 +94,16 @@ internal class MXSuggestedRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
...
+            self?.updateData()
+        }
     }
     
     private func removeDataObservers(for space: MXSpace?) {
         space?.room?.removeListener(spaceEventsListener)
+        if let observer = sessionDidSyncObserver {
```
ismailgulek, line 104: If we're removing the listener in this method, we should add a new one in `addDataObservers(for space:`method. Otherwise, we would lose it when updating the space.

gileluard: Not sure to follow you on this one. `sessionDidSyncObserver` is already set in `addDataObservers(for space: MXSpace?)`

ismailgulek (approved):

MatrixSDK/Data/RoomList/Common/MXSuggestedRoomListDataFetcher.swift
ismailgulek: Ah sorry, i saw it wrong.

PR #1389, opened 2022-03-04
Threads server support check

ismailgulek:
Fix vector-im/element-ios#5744

### DIFF
--- a/MatrixSDK/Data/Store/MXNoStore/MXNoStore.m
+++ b/MatrixSDK/Data/Store/MXNoStore/MXNoStore.m
@@ -433,6 +433,14 @@
 {
 }
 
+- (MXMatrixVersions *)supportedMatrixVersions
+{
+    return nil;
+}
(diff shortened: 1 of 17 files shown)


MatrixSDK/Data/Store/MXNoStore/MXNoStore.m
```diff
@@ -433,6 +433,14 @@ - (void)storeHomeserverCapabilities:(MXCapabilities *)homeserverCapabilities
 {
 }
 
+- (MXMatrixVersions *)supportedMatrixVersions
+{
+    return nil;
+}
```
gileluard, line 439:
```suggestion
}

```

gileluard (approved):
LGTM (feel free to accept or not my suggestion).

MatrixSDK/Data/Store/MXNoStore/MXNoStore.m
ismailgulek: Thanks, i won't commit this, as i would like to keep those two methods close.

PR #1393, opened 2022-03-08
Thread Notifications Fixes

ismailgulek:
Part of vector-im/element-ios#5749.

Decrypts events fetched from server before adding to the timeline.


stefanceriu (approved):
👍

PR #1394, opened 2022-03-08
Update property for in-thread replies

ismailgulek:
Fix vector-im/element-ios#5704

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -795,7 +795,7 @@ NSInteger const kMXRoomAlreadyJoinedErrorCode = 9001;
             if (isReply)
             {
                 //  this will be a real in-thread reply
-                mutableContent[kMXEventRelationRelatesToKey][@"m.in_reply_to"][@"m.render_in"] = @[MXEventRelationTypeThread];
+                mutableContent[kMXEventRelationRelatesToKey][@"io.element.display_reply_fallback"] = @(YES);
             }
             else
             {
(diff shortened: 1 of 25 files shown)


MatrixSDK/Data/MXRoom.m
```diff
@@ -795,7 +795,7 @@ - (MXHTTPOperation*)_sendEventOfType:(MXEventTypeString)eventTypeString
             if (isReply)
             {
                 //  this will be a real in-thread reply
-                mutableContent[kMXEventRelationRelatesToKey][@"m.in_reply_to"][@"m.render_in"] = @[MXEventRelationTypeThread];
+                mutableContent[kMXEventRelationRelatesToKey][@"io.element.display_reply_fallback"] = @(YES);
```
Anderas, line 798: Couldn't / shouldn't this use the `kJSONKeyDisplayReplyFallback` constant instead?

Also I am curious why is the event an `io.element...` instead of `m...` when in the matrixSDK? Is it meant as unsable replacement whilst in development?

Anderas (approved):

MatrixSDK/Data/MXRoom.m
ismailgulek: > Couldn't / shouldn't this use the kJSONKeyDisplayReplyFallback constant instead?

Thanks, i've converted the constants to public and used them.

> Also I am curious why is the event an io.element... instead of m... when in the matrixSDK? Is it meant as unsable replacement whilst in development?

Yes, it'll be updated as `m...` in future.

PR #1395, opened 2022-03-08
`MXRoomEventFilter`: New property names for relation types and senders

ismailgulek:
Fix vector-im/element-ios#5705

This is just updating some property names right now, to be more aligned with stable values, which i'll update later in another PR.

### DIFF
--- a/MatrixSDK/Data/Filters/MXRoomEventFilter.h
+++ b/MatrixSDK/Data/Filters/MXRoomEventFilter.h
@@ -66,14 +66,16 @@
 @property (nonatomic) NSArray<NSString*> *notSenders;
 
 /**
- A list of relation types which must be bundled with the event to include it. If this list is absent then no filtering is done on relation types.
+ A list of relation types to include. An event `A` is included in the filter only if there exists another event `B`,
+ which relates to `A` with a relation type included in the list.
+ If this list is absent then no filtering is done on relation types.
  */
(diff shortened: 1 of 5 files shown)


MatrixSDK/Data/Filters/MXRoomEventFilter.h
```diff
@@ -66,14 +66,16 @@
...
+ A list of relation types to include. An event `A` is included in the filter only if there exists another event `B`,
+ which relates to `A` with a relation type included in the list.
+ If this list is absent then no filtering is done on relation types.
  */
-@property (nonatomic) NSArray<NSString*> *relationTypes;
+@property (nonatomic) NSArray<NSString*> *relatedByTypes;
 
 /**
- A list of senders of relations
+ A list of senders to include. An event `A` is included in the filter only if there exists another event `B` which relates to `A` and has a `sender` included in the list.
```
pixlwave, line 76: Might be useful to add whether this sender is checked against `A` or `B` in this doc.

pixlwave (approved):
LGTM, 1 small comment.

MatrixSDK/Data/Filters/MXRoomEventFilter.h
ismailgulek: Thanks for the suggestion, added.

PR #1399, opened 2022-03-10
Threads: Stable Values

ismailgulek:
Fixes vector-im/element-ios#5791.

This will cause all thread events to move to the main timeline of a room. Will be based on `develop` after the PR merged.


aringenbach (approved):
LGTM

PR #1402, opened 2022-03-13
Increase CLIENT_TIMEOUT_MS for initial /sync

CarlAmbroselli:
Increase the client timeout, failed to do an initial sync for me on iOS simulator (1020 rooms) when using the SDK in a custom app.

### Pull Request Checklist


stefanceriu (commented):
Unfortunately this won't fix your problem. Load balancer timeouts will trigger before the 240 mark for example. 
Because /sync v2 is generally slow the indended behavior for the clients is to retry sync requests until the backend responds with a proper result.
On the synapse side, sync results will be cached for a small period of time to ensure that subsequent client requests receive the data in the end.

CarlAmbroselli:
Thanks for that info, that's good to know!

PR #1404, opened 2022-03-14
Prevent skipping an item between each pagination batch

Anderas:
Fixes [#5819](https://github.com/vector-im/element-ios/issues/5819)

Each time we backpaginate on a timeline, the `MXEventsEnumeratorOnArray` will continuously call `nextEvent` to reach a desired count of events. The problem is that this is performed in a `while` loop where checking the current count of events happens *after* `nextEvent` has been called, meaning that each time we call `nextBatch` we decrement the `paginationPosition` by one more than we should, and thus skipping the display of one event between each pagination.


manuroe (approved):
This is an excellent catch!
I still cannot realise that we had this bug for years.

PR #1408, opened 2022-03-16
Distinguish between original and edit message when preventing replay attacks

Anderas:
Fixes https://github.com/vector-im/element-ios/issues/5835

This solution is not perfect and bloats the `decrypt` API even further, but is the most straightforward way to fix the issue where an edit event is first decrypted and then the original event with the same content (because it has been aggregated) is being decrypted and fails, resulting in the message being dropped from the timeline.

In this solution we use the `isEditEvent` flag as yet another component of `messageIndexKey` to determine whether this is likely a replay attack or not.

### DIFF
--- a/MatrixSDK.xcodeproj/project.pbxproj
+++ b/MatrixSDK.xcodeproj/project.pbxproj
@@ -5509,9 +5509,9 @@
 			buildConfigurationList = 32C6F94019DD814400EA4E9C /* Build configuration list for PBXNativeTarget "MatrixSDK-iOS" */;
 			buildPhases = (
 				385E87177134AE6EBD88F314 /* [CP] Check Pods Manifest.lock */,
+				32C6F92A19DD814400EA4E9C /* Headers */,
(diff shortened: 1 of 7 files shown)


MatrixSDK.xcodeproj/project.pbxproj
```diff
@@ -5509,9 +5509,9 @@
 			buildConfigurationList = 32C6F94019DD814400EA4E9C /* Build configuration list for PBXNativeTarget "MatrixSDK-iOS" */;
 			buildPhases = (
 				385E87177134AE6EBD88F314 /* [CP] Check Pods Manifest.lock */,
+				32C6F92A19DD814400EA4E9C /* Headers */,
```
Anderas, line 5512: Xcode started complaining about the order of these, but I am actually surprised this ever worked, The headers should be before the sources.

manuroe (approved):
LGTM. I hope we do not decrypt some messages more than twice ;) but it will be another real bug.
Thanks for the good test coverage.

PR #1412, opened 2022-03-17
Fix thread events ordering

ismailgulek:
Fixes vector-im/element-ios#5840
Fixes vector-im/element-ios#5832 (only empty screen case).


stefanceriu (approved):
👍

PR #1414, opened 2022-03-18
Fix MXRoom retain cycles.

pixlwave:
As part of https://github.com/vector-im/element-ios/issues/5805 this fixes some memory issues discovered.

In particular `MXRoomOperation`'s block would often have a strong reference to the operation that owned it, causing them to hang around even after being removed from the `orderedOperations` array.

Additionally, `MXRoomAccountDataUpdater` and `MXRoomSummaryUpdater` now use weak map tables to avoid strong references that hold on to `MXRoom` instances forever.

### DIFF
--- a/MatrixSDK/MXSession.h
+++ b/MatrixSDK/MXSession.h
@@ -1084,12 +1084,16 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
 /**
  Delegate for updating room summaries.
  By default, it is the one returned by [MXRoomSummaryUpdater roomSummaryUpdaterForSession:].
+ 
+ This property is strong as the only other reference to the delegate is a weak ref in the updaterPerSession table.
  */
 @property id<MXRoomSummaryUpdating> roomSummaryUpdateDelegate;
 
 /**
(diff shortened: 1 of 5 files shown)


aringenbach (approved):

MatrixSDK/MXSession.h
```diff
@@ -1084,12 +1084,16 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
 /**
  Delegate for updating room summaries.
  By default, it is the one returned by [MXRoomSummaryUpdater roomSummaryUpdaterForSession:].
+ 
+ This property is strong as the only other reference to the delegate is a weak ref in the updaterPerSession table.
  */
 @property id<MXRoomSummaryUpdating> roomSummaryUpdateDelegate;
```
aringenbach, line 1090: I wonder if these two should really be called delegates though :)

pixlwave: It's a very good point to be fair. I'm not sure it would be worth making source breaking changes to the SDK for this though. I doubt the account data updater is used much by others but the room summary updater could well be 😕

PR #1426, opened 2022-03-28
Return room identifier and known servers when resolving alias

Anderas:
Relates to https://github.com/vector-im/element-ios/issues/4858

The [directory/room/<room_alias>](https://spec.matrix.org/v1.2/client-server-api/#get_matrixclientv3directoryroomroomalias) endpoint used to convert room alias to a room identifier returns a list of servers that are aware of the room. In the existing SDK implementation this list of servers is never stored (we only grab the identifier), meaning that if we later want to join this room from another home server, it is no longer possible without the knowledge of the servers.

To solve this on the SDK side, store both `room_id` as well as list of `servers` in a new `MXRoomAliasResolution` object which can be fetched via `resolveRoomAlias` endpoint, replacing the simpler `roomIdForAlias`

### DIFF
--- a/MatrixSDKTests/MXRestClientTests.m
+++ b/MatrixSDKTests/MXRestClientTests.m
@@ -383,6 +383,8 @@
                     XCTAssertNotNil(resolution);
                     XCTAssertNotEqual(resolution.roomId.length, 0);
                     XCTAssertEqualObjects(resolution.roomId, roomId, @"Mapping from room alias to room ID is wrong");
+                    XCTAssertNotEqual(resolution.servers.count, 0);
+                    XCTAssertNotEqual(resolution.servers[0].length, 0);
                     
                     // Test with a valid alias which already exists
                     [bobRestClient2 addRoomAlias:roomId alias:correctAlias success:^{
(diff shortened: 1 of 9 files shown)


MatrixSDKTests/MXRestClientTests.m
```diff
@@ -377,11 +378,11 @@ - (void)testRoomAddAlias
...
                 
-                [bobRestClient2 roomIDForRoomAlias:correctAlias success:^(NSString *roomId2) {
+                [bobRestClient2 resolveRoomAlias:correctAlias success:^(MXRoomAliasResolution *resolution) {
                     
-                    XCTAssertNotNil(roomId2);
-                    XCTAssertNotEqual(roomId2.length, 0);
-                    XCTAssertEqualObjects(roomId2, roomId, @"Mapping from room alias to room ID is wrong");
+                    XCTAssertNotNil(resolution);
+                    XCTAssertNotEqual(resolution.roomId.length, 0);
+                    XCTAssertEqualObjects(resolution.roomId, roomId, @"Mapping from room alias to room ID is wrong");
```
aringenbach, line 386: I wonder if we could test something relevant on the servers list 🤔

aringenbach (approved):
LGTM 👍

MatrixSDKTests/MXRestClientTests.m
Anderas: Oh yea! I was going to do that and then it slipped my mind, thanks!

PR #1428, opened 2022-03-31
[Location Sharing]: Pin drop location sharing

MaximeEvrard42:
https://github.com/vector-im/element-ios/issues/5858

### DIFF
--- a/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
+++ b/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
@@ -72,11 +72,15 @@
     }
     
     NSDictionary *assetDictionary = JSONDictionary[kMXMessageContentKeyExtensibleAssetMSC3488];
+    if (assetDictionary == nil)
+    {
+        assetDictionary = JSONDictionary[kMXMessageContentKeyExtensibleAsset];
+    }
     if (assetDictionary)
     {
         assetType = [[[MXEventAssetTypeMapper alloc] init] eventAssetTypeFrom:assetDictionary[kMXMessageContentKeyExtensibleAssetType]];
     } else {
-        // Should behave like m.self if assetType is null
+        // Should behave like m.self if assetType is nil
         assetType = MXEventAssetTypeUser;
     }
     
--- a/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
+++ b/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
@@ -17,6 +17,14 @@
 #import "MXEventContentLocation.h"
 #import "MXEvent.h"
 
+#import "MatrixSDKSwiftHeader.h"
+
+@interface MXEventContentLocation()
+
+@property (nonatomic) MXEventAssetTypeMapper *eventAssetTypeMapper;
+
+@end
+
 @implementation MXEventContentLocation
 
 - (instancetype)initWithAssetType:(MXEventAssetType)assetType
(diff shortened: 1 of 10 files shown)


MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
```diff
@@ -63,16 +72,12 @@ + (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
...
-    {
-        assetDictionary = JSONDictionary[kMXMessageContentKeyExtensibleAsset];
-    }
-    
     if (assetDictionary)
     {
-        if (![assetDictionary[kMXMessageContentKeyExtensibleAssetType] isEqualToString:kMXMessageContentKeyExtensibleAssetTypeUser]) {
-            assetType = MXEventAssetTypeGeneric;
-        }
+        assetType = [[[MXEventAssetTypeMapper alloc] init] getMXEventAssetTypeFrom: assetDictionary[kMXMessageContentKeyExtensibleAssetType]];
```
pixlwave, line 77:
```suggestion
        assetType = [[[MXEventAssetTypeMapper alloc] init] getMXEventAssetTypeFrom:assetDictionary[kMXMessageContentKeyExtensibleAssetType]];
```

MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
```diff
@@ -63,16 +72,12 @@ + (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
     }
     
     NSDictionary *assetDictionary = JSONDictionary[kMXMessageContentKeyExtensibleAssetMSC3488];
-    if (assetDictionary == nil)
-    {
-        assetDictionary = JSONDictionary[kMXMessageContentKeyExtensibleAsset];
-    }
```
pixlwave, line 69: I'm probably missing context, but isn't this removing support for the stable event type if the dictionary contained `m.asset` instead of `org.matrix.msc3488.asset`?

MatrixSDK/Utils/MXEventAssetTypeMapper.swift
```diff
@@ -0,0 +1,54 @@
...
+import Foundation
+
+/// MXEventAssetTypeMapper enables to get the corresponding MXEventAssetType from event key String and the other way around.
+@objcMembers
+public class MXEventAssetTypeMapper: NSObject {
+    
+    // MARK: - Public
+    
+    /// Return event key String associate to the MXEventAssetType given
+    public func getEventKey(from eventAssetType: MXEventAssetType) -> String {
```
pixlwave, line 26: As this is simply mapping one value to another, it probably doesn't need the `get` :)
```suggestion
    public func eventKey(for eventAssetType: MXEventAssetType) -> String {
```

MatrixSDK/Utils/MXEventAssetTypeMapper.swift
```diff
@@ -0,0 +1,54 @@
...
+        case .generic:
+            eventKey = ""
+        @unknown default:
+            eventKey = ""
+        }
+        return eventKey
+    }
+    
+    /// Return MXEventAssetType associate to the event key String given
+    public func getMXEventAssetType(from eventKey: String) -> MXEventAssetType {
```
pixlwave, line 42:
```suggestion
    public func eventAssetType(for eventKey: String) -> MXEventAssetType {
```

pixlwave (commented):
Looks good, will be great to be able to send more location types 😎 I've added a couple of comments inline.

MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
MaximeEvrard42: I don't think so, and according to MSC https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/location/proposals/3488-location.md#proposal when m.asset is missing, it should be treated like "m.self" and it was not done

pixlwave (approved):
Changes look good to me 👍

PR #1432, opened 2022-04-04
Add assertion failure macros

Anderas:
Add macros and `MXLog` methods to log assertions that "crash" the app in DEBUG mode. This is a useful tool during development: wherease merely logging to console is easy to miss, an assertion forces the developer to debug a given issue whilst the context is still in place.


MatrixSDK/Data/RoomList/CoreData/MXCoreDataRoomListDataFetcher.swift
```diff
@@ -114,10 +113,8 @@ internal class MXCoreDataRoomListDataFetcher: NSObject, MXRoomListDataFetcher {
         return result
     }
     
-    internal init(session: MXSession?,
-                  fetchOptions: MXRoomListDataFetchOptions,
+    internal init(fetchOptions: MXRoomListDataFetchOptions,
```
Anderas, line 116: Unrelated change, the `session` is never used so can be removed from init parameters

pixlwave (approved):
LGTM 👍

PR #1437, opened 2022-04-08
Notify other devices of verification acceptance

langleyd:
Fixes https://github.com/vector-im/element-ios/issues/4470

When a client receives a verification request acceptance it needs to notify other devices that it was accepted so they can dismiss their requests.

### Todo
- [ ] check if the same issue exists on android and fix. https://github.com/vector-im/element-android/issues/5724


gileluard (approved):
LGTM

PR #1444, opened 2022-04-21
Store `sharedHistory` flag for inbound Megolm sessions

Anderas:
Relates to https://github.com/vector-im/element-ios/issues/4947 which is an implementation of https://github.com/matrix-org/matrix-doc/pull/3061

This PR stores a `sharedHistory` flag on all `MXOlmInboundGroupSession`s, which reflects the current history visibility of a room. In a future PR this flag will be used to determine which sessions can be shared with other users when invited.

The changes in this PR:
- add `sharedHistory` flag to `MXOlmInboundGroupSession` and other related objects, so that it can be both persisted as well as backed up to the server
- add `roomId` to `MXRealmOlmInboundGroupSession` so that inbound sessions can be fetched by `roomId` in the future (i.e. "share keys per room"). Otherwise we would need to fetch them all, deserialize the data and only then filter in memory.
- extact a method out `MXMegolmEncryption.m` which checks whether a session should be reset and add another check based on room history visibility change


MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -579,13 +579,18 @@ public enum MXBackgroundSyncServiceError: Error {
...
         }
         
-        olmDevice.addInboundGroupSession(sessionId,
-                                         sessionKey: sessionKey,
-                                         roomId: roomId,
-                                         senderKey: senderKey,
-                                         forwardingCurve25519KeyChain: forwardingKeyChain,
-                                         keysClaimed: keysClaimed,
-                                         exportFormat: exportFormat)
+        MXRoomState.load(from: store, withRoomId: roomId, matrixSession: nil) { [weak self] state in
```
Anderas, line 582: This feels a bit awkward but I did not find a better way to get the room state whilst in background sync

MatrixSDK/Data/MXRoom.m
```diff
@@ -1839,6 +1839,11 @@ - (MXHTTPOperation *)setHistoryVisibility:(MXRoomHistoryVisibility)historyVisibi
     return [mxSession.matrixRestClient setRoomHistoryVisibility:self.roomId historyVisibility:historyVisibility success:success failure:failure];
 }
 
+- (MXRoomHistoryVisibility)historyVisibility
+{
+    return liveTimeline.state.historyVisibility;
```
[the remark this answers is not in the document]
ismailgulek: I believe we can store this info in room summary.

MatrixSDK/Background/MXBackgroundSyncService.swift
ismailgulek: We can avoid this and async call if we have `historyVisibility` on room summary

MatrixSDK/Crypto/Data/MXMegolmSessionData.h
```diff
@@ -54,6 +54,14 @@
  */
 @property NSString *sessionKey;
 
+/**
+ Flag indicating whether the history of this room is considered as shared.
+ 
+ This is typically the case if room's `historyVisibility` is set to `world_readable` or `joined`.
```
ismailgulek, line 60:
```suggestion
 This is typically the case if room's `historyVisibility` is set to `world_readable` or `shared`.
```

MatrixSDK/Crypto/Data/MXOlmInboundGroupSession.h
```diff
@@ -66,6 +66,14 @@
  */
 @property (nonatomic) NSDictionary<NSString*, NSString*> *keysClaimed;
 
+/**
+ Flag indicating whether the history of this room is considered as shared.
+ 
+ This is typically the case if room's `historyVisibility` is set to `world_readable` or `joined`.
```
ismailgulek, line 72:
```suggestion
 This is typically the case if room's `historyVisibility` is set to `world_readable` or `shared`.
```

MatrixSDK/Crypto/MXOlmDevice.h
```diff
@@ -230,15 +230,18 @@ Determine if an incoming messages is a prekey message matching an existing sessi
  @param forwardingCurve25519KeyChain devices which forwarded this session to us (normally empty)
  @param keysClaimed Other keys the sender claims.
  @param exportFormat YES if the megolm keys are in export format (ie, they lack an ed25519 signature).
+ @param sharedHistory YES if the session was created whilst room's history was set to visible (i.e. `world_readable` or `joined`
```
ismailgulek, line 233:
```suggestion
 @param sharedHistory YES if the session was created whilst room's history was set to visible (i.e. `world_readable` or `shared`)
```

MatrixSDKTests/Crypto/Data/MXMegolmSessionDataTests.swift
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+@testable import MatrixSDK
+
+class MXMegolmSessionDataTests: XCTestCase {
```
ismailgulek, line 20: This can be unit tests?

MatrixSDKTests/Crypto/Data/MXOlmInboundGroupSessionTests.swift
```diff
@@ -0,0 +1,56 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+@testable import MatrixSDK
+
+class MXOlmInboundGroupSessionTests: XCTestCase {
```
ismailgulek, line 20: This can be unit tests?

ismailgulek (changes requested):

MatrixSDK/Crypto/Data/MXMegolmSessionData.h
Anderas: 🤦 thx

MatrixSDKTests/Crypto/Data/MXMegolmSessionDataTests.swift
Anderas: Makes sense, didn't know about the naming convention

MatrixSDK/Background/MXBackgroundSyncService.swift
Anderas: Yea that does make it better, thanks

MatrixSDK/Data/MXRoom.m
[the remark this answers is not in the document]
Anderas: Initially wasn't too keen on adding it to the Core Data model, but in the long run its probably more consistent that way with the rest of the state events, thx.

MatrixSDK/Data/RoomSummaryStore/CoreData/MXCoreDataRoomSummaryStore.xcdatamodeld/.xccurrentversion
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>_XCCurrentVersionName</key>
+	<string>MXRoomSummaryCoreDataStore2.xcdatamodel</string>
```
Anderas, line 6: I do not know why we haven't been versioning the `MXRoomSummaryCoreDataStore` despite plenty of updates to the model. When tested without a new version the app runs fine without crashes, even if an existing property is renamed, suggesting Core Data is doing a lot under the hood to guess mapping old data without a known model to the current one, either way this is an undocumented behaviour. For that reason it is still better to declare a new version each time the model is changed.

ismailgulek (changes requested):
Just one comment about test plan file, otherwise LGTM 👍

MatrixSDK/Data/RoomSummaryStore/CoreData/MXCoreDataRoomSummaryStore.xcdatamodeld/.xccurrentversion
ismailgulek: Yeah you right. Thanks for adding the first new version.

MatrixSDKTests/TestPlans/UnitTests.xctestplan
```diff
@@ -47,8 +47,14 @@
         "MXHTTPAdditionalHeadersUnitTests",
         "MXJSONModelUnitTests",
         "MXKeyProviderUnitTests",
+        "MXLoggerUnitTests\/testDeleteLogFiles",
```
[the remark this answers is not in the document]
Anderas: Makes sense, thx

ismailgulek (approved):

PR #1449, opened 2022-04-28
Change `sharedHistory` flag to unstable variant

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/4947 which is an implementation of https://github.com/matrix-org/matrix-doc/pull/3061

Handle several edge cases when key sharing:

- use unstable parameter name of `shared_history` when exporting and importing keys. This makes it compatible with the existing web implementation (e.g. backup keys on the web and restore on iOS => preserve `sharedHistory` flag)
- add integration tests for key sharing across several devices
- ensure that keys are only shared if the roomId of the session matches the room a user is invited into

### DIFF
--- a/MatrixSDKTests/MXCryptoShareTests.m
+++ b/MatrixSDKTests/MXCryptoShareTests.m
@@ -21,6 +21,7 @@
 
 #import "MXCrypto_Private.h"
 #import "MXCryptoStore.h"
+#import "MXMemoryStore.h"
 
 #import <OHHTTPStubs/HTTPStubs.h>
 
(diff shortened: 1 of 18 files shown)


MatrixSDKTests/MXCryptoShareTests.m
```diff
@@ -571,6 +572,228 @@ - (void)testNotBetterSharedSession
...
+                                 {
+                                    
+                                    // Initially Alice2 has no keys
+                                    XCTAssertEqual([self numberOfKeysInSession:aliceSession2], 0);
+                        
+                                    // Make each Alice device trust each other
+                                    [aliceSession1.crypto setDeviceVerification:MXDeviceVerified forDevice:aliceSession2.myDeviceId ofUser:aliceSession1.myUserId success:^{
+                                        [aliceSession2.crypto setDeviceVerification:MXDeviceVerified forDevice:aliceSession1.myDeviceId ofUser:aliceSession1.myUserId success:^{
+                                            
+                                            // Alice2 pagingates in the room to get the keys forwarded to her
```
aringenbach, line 707:
```suggestion
                                            // Alice2 paginates in the room to get the keys forwarded to her
```

aringenbach (approved):
Code LGTM (would be awesome to write new tests in Swift if possible though)

Anderas:
> Code LGTM (would be awesome to write new tests in Swift if possible though)

Yea in this case I wanted to write the integration tests consistently with the rest of the `CryptoSharingTests`, but should start switching over to Swift there as well.

PR #1462, opened 2022-05-12
Replace NSParameterAssert with MXLogFailure and early return

Anderas:
Replace `NSParameterAssert`, which is only cought in develop and not indicated in rageshakes, by `MXLogFailure` (which is also caught in debug mode, but adds logs to rageshakes) and an early exit from the method. This is to address an issue where inconsistent state (previously asserted by the `NSParameterAssert`) is nevertheless ignored by the method in production and potentially causing further inconsistencies.

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1928,7 +1928,12 @@ typedef void (^MXOnResumeDone)(void);
 
 - (void)handleBackgroundSyncCacheIfRequiredWithCompletion:(void (^)(void))completion
 {
-    NSParameterAssert(_state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused);
+    BOOL isInValidState = _state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused;
+    if (!isInValidState) {
+        MXLogFailure(@"[MXSession] state is not valid to handle background sync cache, investigate why the method was called");
+        completion();
+        return;
+    }
 
     if (!self.hasAnyBackgroundCachedSyncResponses)
     {


MatrixSDK/MXSession.m
```diff
@@ -1928,7 +1928,12 @@ - (BOOL)hasAnyBackgroundCachedSyncResponses
 
 - (void)handleBackgroundSyncCacheIfRequiredWithCompletion:(void (^)(void))completion
 {
-    NSParameterAssert(_state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused);
+    BOOL isInValidState = _state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused;
+    if (!isInValidState) {
+        MXLogFailure(@"[MXSession] state is not valid to handle background sync cache, investigate why the method was called");
+        completion();
```
ismailgulek, line 1934: `completion` can be nil, in theory.

MatrixSDK/MXSession.m
```diff
@@ -1928,7 +1928,12 @@ - (BOOL)hasAnyBackgroundCachedSyncResponses
 
 - (void)handleBackgroundSyncCacheIfRequiredWithCompletion:(void (^)(void))completion
 {
-    NSParameterAssert(_state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused);
+    BOOL isInValidState = _state == MXSessionStateStoreDataReady || _state == MXSessionStatePaused;
+    if (!isInValidState) {
+        MXLogFailure(@"[MXSession] state is not valid to handle background sync cache, investigate why the method was called");
```
ismailgulek, line 1933: It may be a good idea to log current `state` also.

MatrixSDK/MXSession.m
Anderas: Yea forgot about that, thanks

ismailgulek (approved):

PR #1463, opened 2022-05-13
vector-im/element-ios/issues/6149 - Add `io.element.video` room type.

stefanceriu:


Anderas (approved):
~Is there some ticket / spec with more context regarding this type?~ Ah right its in the title https://github.com/vector-im/element-ios/issues/6149

PR #1467, opened 2022-05-19
Shared history flag in crypto module

Anderas:
Crypto module should internally respect build flag when deciding if a room has shared history or not.


pixlwave (approved):
LGTM 👍

PR #1475, opened 2022-05-24
Recreate summaries when detected missing

ismailgulek:
Fixes vector-im/element-ios#5924

Also maybe a fix for vector-im/element-ios#6006


RamadanAlharahsheh:
Is this the same issue when you have the room but the summary is nil? 
And does this fix the mentioned issue? 
Thanx in advance.

SBiOSoftWhare (approved):

PR #1476, opened 2022-05-25
Add more logs when encrypting messages

Anderas:
Relates to https://github.com/matrix-org/element-ios-rageshakes/issues/18461

There are rare reports of rooms forgetting that they are encrypted, and it is plausable that this is caused by some state corruption. Given some existing measures that restore summary encryption if crypto store contains an algorithm, this state loss would need to be more comprehensive. Either way there isn't currently enough logs to pinpoint more precisely why encryption is not possible. This PR merely adds a few more logs to the encryption flow.

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -548,6 +548,8 @@ NSInteger const kMXRoomInvalidInviteSenderErrorCode = 9002;
                             success:(void (^)(NSString *eventId))success
                             failure:(void (^)(NSError *error))failure
 {
+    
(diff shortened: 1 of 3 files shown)


pixlwave (approved):
LGTM, although probably needs a changelog?

BillCarsonFr (approved):
LGTM

Anderas:
> LGTM Do we have a log when the event is finaly sent (with eventId printed?) would help for investigation. Also maybe adding the roomId in the new log lines would be helpfull

Yes we log sending of the message in the HTTP client, which includes the room id as well as temporary local id. Once the message is sent, we log the transformation of the local id into permament id. The logs are a bit scattered around amid other stuff but all the info is there. Adding one more roomId at the start to make this easier.

PR #1478, opened 2022-05-27
Add extra logs when saving and loading room state

Anderas:
Relates to repeatedly reported issues:
- https://github.com/matrix-org/element-ios-rageshakes/issues/18751
- https://github.com/matrix-org/element-ios-rageshakes/issues/18740

This does not fix the state corruption issue but uses newer KeyedUnarchiver/Archiver API which provides errors when something fails. Additionally add logs when we arent able to load state or the number of state events we are persisting.


gileluard (approved):
LGTM

PR #1482, opened 2022-05-31
Cannot join rooms with large amount of via servers #6224

gileluard:
fix https://github.com/vector-im/element-ios/issues/6224

### DIFF
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -93,9 +93,6 @@ NSString *const kMXCredentialsNewRefreshTokenDataKey = @"refresh_token_data";
  */
 #define PREEMPT_REFRESH_EXPIRATION_INTERVAL 60000
 
-/// Maximum number of caracters for a URL
-static const NSInteger kUrlMaxLength = 2048;
-
 /**
  Authentication flow: register or login
  */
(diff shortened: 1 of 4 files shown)


MatrixSDK/MXRestClient.m
```diff
@@ -2416,14 +2419,21 @@ - (MXHTTPOperation*)joinRoom:(NSString*)roomIdOrAlias
         for (NSString *viaServer in viaServers)
         {
             NSString *value = [MXTools encodeURIComponent:viaServer];
+            NSString *parameter = [NSString stringWithFormat:@"server_name=%@", value];
```
Anderas, line 2422: Optional improvement, but it would be nice to extract this now into `MXTools` or somewhere else where the code can be tested. The code is not super complicated, but does include some magical constants (2048, +1) and it would be good to validate with a test.

MatrixSDK/Utils/MXTools.swift
```diff
@@ -26,4 +28,20 @@ public extension MXTools {
         return state.description
     }
     
+    @objc
+    static func urlString(base: String, queryParameters: [String]) -> String {
+        var urlString = base
+        var hasQueryParameters = urlString.firstIndex(of: "?") != nil
```
Anderas, line 34: `urlString.firstIndex(of: "?") != nil` can also be `urlString.contains("?")`

Anderas (approved):

PR #1483, opened 2022-05-31
Reload room state if detected empty

Anderas:
Aims to solve state loss reported (among others) in:
- https://github.com/matrix-org/element-ios-rageshakes/issues/18865
- https://github.com/matrix-org/element-ios-rageshakes/issues/18751
- https://github.com/matrix-org/element-ios-rageshakes/issues/18740

# Problem

After [previously](https://github.com/matrix-org/matrix-ios-sdk/pull/1478) adding some extra logs, the following is likely a reason for state corruption:

1. Some rooms report having no `state` file present on disk. This should be the case only when there is no user logged in, or when we are clearning cache / initial syncing. After clearing of the cache all rooms should be restored to their full glory, but based on the logs this does not actually happen. The issue has been reported on particularly large accounts, so the assumption is that during the very long initial sync (minutes), some room state fails to be loaded, either due to some implementation bug, or because the app is killed etc
2. When the state file is missing, `MXRoomState` will load 0 state events from disk and later write 0 state events to disk. At this point `MXFileStore` has a file but no state in it. Note that the user may not see any issue at this point, because data is also loaded from `MXRoomSummary` which is managed by Core Data, and is only updated when new state events arrive. In other words, even if state file is missing, the room may be still correctly showing a name at this point.
3. As time goes on, new state events arrive to the device and are correctly stored on disk, but the previously missing state is never recovered. Additionally, only now will `MXRoomSummary` be updated, and valid name be replaced with "X & Y others", because there is indeed no state with name present. 

In short, the root cause is missing state file at t0, which only because apparent in the form of "X and Y others" name at t1, when new state events have arrived (minutes, hours or days later).

# Solution
Ideally we would want to identify and resolve the issue where state file is never recreated after clearing of cache, initial sync or log in. This is however still too complex, as the bug may be in `MXSession` managing sync data, or `MXFileStore` suffering from a race condition, or any other issue.

Instead we will monitor when a state file is missing during the loading of `MXRoomState`, and if no state events are returned from the store, we will trigger a (previously unused) HTTP request to fetch the entire room state and persist it on disk. When this happens, and new state events are returned from the API, name of the room will be automatically correct, because it is updated on any new state events.

### DIFF
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -112,13 +112,34 @@
                matrixSession:(MXSession *)matrixSession
                   onComplete:(void (^)(MXRoomState *roomState))onComplete
 {
+    NSString *loadEventId = [NSUUID UUID].UUIDString;
+    MXLogDebug(@"[MXRoomState] loadRoomStateFromStore(%@): Loading state for room %@", loadEventId, roomId)
+    
     MXRoomState *roomState = [[MXRoomState alloc] initWithRoomId:roomId andMatrixSession:matrixSession andDirection:YES];
     if (roomState)
     {
         [store stateOfRoom:roomId success:^(NSArray<MXEvent *> * _Nonnull stateEvents) {
-            [roomState handleStateEvents:stateEvents];
+            if (!stateEvents.count) {
+                MXLogWarning(@"[MXRoomState] loadRoomStateFromStore(%@): No state events stored, loading from api", loadEventId);
+                
+                [matrixSession.matrixRestClient stateOfRoom:roomId success:^(NSArray *JSONData) {
+                    NSArray<MXEvent *> *events = [MXEvent modelsFromJSON:JSONData];
+                    MXLogDebug(@"[MXRoomState] loadRoomStateFromStore(%@): Loaded %lu events from api", loadEventId, events.count);
+                    
+                    [roomState handleStateEvents:events];
+                    onComplete(roomState);
+                } failure:^(NSError *error) {
+                    MXLogError(@"[MXRoomState] loadRoomStateFromStore(%@): Failed to load any events from api", loadEventId);
+                    
+                    onComplete(roomState);
+                }];
+            } else {
+                MXLogDebug(@"[MXRoomState] loadRoomStateFromStore(%@): Initializing with %lu state events", loadEventId, stateEvents.count);
+                
+                [roomState handleStateEvents:stateEvents];
+                onComplete(roomState);
+            }
 
-            onComplete(roomState);
         } failure:nil];
     }
 }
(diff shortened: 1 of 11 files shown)


MatrixSDK/Data/MXRoomState.m
```diff
@@ -112,13 +112,34 @@ + (void)loadRoomStateFromStore:(id<MXStore>)store
                matrixSession:(MXSession *)matrixSession
                   onComplete:(void (^)(MXRoomState *roomState))onComplete
 {
+    NSString *loadEventId = [NSUUID UUID].UUIDString;
```
Anderas, line 115: This is a bit of an experiment, adding a unique ID to all logs that are logically related but may not follow right after each other in time (e.g. making API calls). In this case if multiple room states are being loaded, the ID will disambiguate them. If this approach turns out to be useful, we could abstract this identification into a better API.

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -2251,9 +2251,8 @@ - (id)loadObjectOfClasses:(NSSet<Class> *)classes fromFile:(NSString *)file
         }
         
         id object = [NSKeyedUnarchiver unarchivedObjectOfClasses:classes fromData:data error:&error];
-        if (object)
+        if (object && !error)
         {
-            MXLogDebug(@"[MXFileStore] Loaded object from class");
```
Anderas, line 2256: This particular log is not very useful I think, so removing

MatrixSDK/MXRestClient.h
```diff
@@ -1444,7 +1444,7 @@ NS_REFINED_FOR_SWIFT;
  @return a MXHTTPOperation instance.
  */
 - (MXHTTPOperation*)stateOfRoom:(NSString*)roomId
-                        success:(void (^)(NSDictionary *JSONData))success
+                        success:(void (^)(NSArray *JSONData))success
```
Anderas, line 1447: This type was wrong. AFNetworking will return responseObject of type `id`, so we cannot generally assume it is always a dictionary. In this particular case the API actually returns an array. We never had issues with this, because this API was not being used by the app until now.

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -29,7 +29,7 @@
 #import "MatrixSDKSwiftHeader.h"
 #import "MXFileRoomSummaryStore.h"
 
-static NSUInteger const kMXFileVersion = 79;
+static NSUInteger const kMXFileVersion = 80;
```
Anderas, line 32: As I understand this, incrementing file store version will delete all data and trigger initial sync. This will ensure all state files with already corrupted data is removed and we start monitoring for empty room state from this point onwards. Are there any further consequences for when this gets released to production users?

MatrixSDK/Data/MXRoomState.m
pixlwave: Sounds helpful to me. I wonder if having `eventId` in the name could cause some confusion until reading through the code though?

Anderas: Yea was wondering about this too. Renamed for now, and I'll think about the abstraction next.

manuroe (approved):
The PR looks good to me. This is a great catch!
Thanks also for starting unit tests based on a stubbed version of MXRestClient.

PR #1484, opened 2022-05-31
Mark MXRestClient init as required for mocking.

pixlwave:
In order to add a protocol that includes the client's `init`, this PR marks it as `required`. Part of 
https://github.com/vector-im/element-ios/issues/6179


MaximeEvrard42 (approved):
LGTM !

PR #1486, opened 2022-06-01
Remove unwanted parts from replies new_content body/formatted_body

aringenbach:
This fixes an issue mentioned in https://github.com/vector-im/element-ios/issues/3517 
It removes body parts that shouldn't be added to the **new_content** body/formatted_body but only to the compatibility body/formatted_body. 

It also enables the ability to edit replies that have lost their mx-reply fallback (preserving the same behaviour as it already has: clients that handle rich replies will display everything properly, clients that don't will only be able to display the reply body)

Here's a screenshot of EleWeb displaying edited reply from an iOS client before and after this update. 

image

### DIFF
--- a/MatrixSDK/Aggregations/MXAggregatedEditsUpdater.m
+++ b/MatrixSDK/Aggregations/MXAggregatedEditsUpdater.m
@@ -80,8 +80,8 @@
         return nil;
     }
     
-    NSString *finalText;
-    NSString *finalFormattedText;
+    NSString *compatibilityText;
+    NSString *compatibilityFormattedText;
     
     if (event.isReplyEvent)
     {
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
Looks good to me 👍  What would be nice is to add a test to check that this works properly in all cases.

aringenbach:
> Looks good to me 👍 What would be nice is to add a test to check that this works properly in all cases.

Sounds good but unfortunately MXAggregatedEditsTests is one of these test cases that are currently not run at all (due to being broken from what I see - the edition of a standard message just doesn't happen on the localhost test) So I guess the first step would be to fix this - which doesn't look trivial - and this small fix is already kinda out of my scope of current work 😅

PR #1487, opened 2022-06-01
Location sharing: Authorize only one live beacon info per member and per room

SBiOSoftWhare:
https://github.com/vector-im/element-ios/issues/6100


ismailgulek (approved):
LGTM

PR #1491, opened 2022-06-03
Location sharing: Fix geo URI parsing with altitude component

SBiOSoftWhare:
Fix https://github.com/vector-im/element-ios/issues/6247


MaximeEvrard42 (approved):
LGTM !

PR #1492, opened 2022-06-07
Accept key backups as usable if they're signed with the master cross-signing key

bradtgmurray:
Accept key backups if they're signed by the master cross-signing key.

Used matrix-js-sdk as a reference which accepts it as usable if either the signing device is verified or it's the cross-signing key. https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/crypto/backup.ts#L426-L433

Signed-off-by: Brad Murray <brad@beeper.com>

### Pull Request Checklist

### DIFF
--- a/MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
+++ b/MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
@@ -1139,18 +1139,18 @@ static Class DefaultAlgorithmClass;
     for (NSString *keyId in mySigs)
     {
         // XXX: is this how we're supposed to get the device id?
-        NSString *deviceId;
+        NSString *deviceIdOrCrossSigningKey;
         NSArray<NSString *> *components = [keyId componentsSeparatedByString:@":"];
         if (components.count == 2)
         {
-            deviceId = components[1];
+            deviceIdOrCrossSigningKey = components[1];
         }
 
-        if (deviceId)
+        if (deviceIdOrCrossSigningKey)
         {
             BOOL valid = NO;
 
-            MXDeviceInfo *device = [self->crypto.deviceList storedDevice:myUserId deviceId:deviceId];
+            MXDeviceInfo *device = [self->crypto.deviceList storedDevice:myUserId deviceId:deviceIdOrCrossSigningKey];
             if (device)
             {
                 NSError *error;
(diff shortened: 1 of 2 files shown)


Anderas (commented):
Hi Brad, thanks for the contribution. I made one comment regarding ensuring we do indeed have the master key when trusting a backup, and it would be also great to add a test case if possible for this, for instance in `MXCryptoBackupTests`

MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
```diff
@@ -1154,6 +1154,8 @@ - (MXKeyBackupVersionTrust *)trustForKeyBackupVersionFromCryptoQueue:(MXKeyBacku
                 }
                 else
                 {
+                    keyBackupVersionTrust.usable = YES;
```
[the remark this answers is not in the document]
bradtgmurray: Only a month and a half later I've finally addressed this feedback, thanks @Anderas !

MatrixSDKTests/MXCryptoBackupTests.m
```diff
@@ -515,9 +515,9 @@ - (void)testTrustForKeyBackupVersion
 // - Alice and Bob have messages in a room
 // - Alice has cross-signing enabled
 // - Alice creates a backup
+// - Modify the backup to remove all signatures other than the MSK so we know that the MSK alone is enough for the device to be trusted
 // - Check the returned MXKeyBackupVersion is trusted
-// -> It must be trusted by 2 entities
-// -> Trusted by her device
+// -> It must be trusted by 1 entities
 // -> Trusted by her MSK
 - (void)testCrossSigningMSKTrustForKeyBackupVersion
```
bradtgmurray, line 522: Repurposed an existing test. Based on the name it seems to have been wanting to test MSKTrust, but because the valid device signature was marking the device usable it was hiding the bug that I'm fixing with this PR. Now the only signature on the backup is the MSK.

bradtgmurray:
Strange, these tests seem to pass locally, any ideas?

MatrixSDKTests/MXCryptoBackupTests.m
Anderas: The new test looks good, however it might be better to create it as copy and leave `testCrossSigningMSKTrustForKeyBackupVersion` as is, otherwise we loose the test case for getting a device signature (`XCTAssertEqual(keyBackupVersionTrust.signatures.count, 2)`)

MatrixSDKTests/MXBaseKeyBackupTests.m
```diff
@@ -509,26 +509,28 @@ - (void)testCrossSigningMSKTrustForKeyBackupVersion
...
+                    keyBackupVersion.authData = [authData JSONDictionary];
+                    
                     // - Check the returned MXKeyBackupVersion is trusted
                     [aliceSession.crypto.backup trustForKeyBackupVersion:keyBackupVersion onComplete:^(MXKeyBackupVersionTrust *keyBackupVersionTrust) {
-
-                        // -> It must be trusted by 2 entities
+                        // -> It must be trusted by 1 entity
                         XCTAssertNotNil(keyBackupVersionTrust);
                         XCTAssertTrue(keyBackupVersionTrust.usable);
-                        XCTAssertEqual(keyBackupVersionTrust.signatures.count, 2);
```
Anderas, line 519: The new test looks good, however it might be better to create it as copy and leave testCrossSigningMSKTrustForKeyBackupVersion as is, otherwise we loose the test case for getting a device signature (XCTAssertEqual(keyBackupVersionTrust.signatures.count, 2))

MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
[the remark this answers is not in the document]
Anderas: Thanks for making the change. I see that we will now be adding a signature object even if the master key verification did not succeed, but it will be set as invalid (plus this copies js / android behaviour) so it sounds good to me

Anderas:
> Strange, these tests seem to pass locally, any ideas?

Unfortunatelly crypto + integration tests are very flaky, we are in the process of enabling + fixing them. I'd say make sure that they pass on your local branch, and optionally you can rerun the GitHub tests a few times. We cannot just turn them off because locally they are not as flaky and thus hard to spot the misbehaving ones.

PR #1494, opened 2022-06-08
Fixes vector-im/element-ios/issues/6227 - Prevent invalid room names …

stefanceriu:
…on member count underflows.


pixlwave (approved):
LGTM 👍

PR #1496, opened 2022-06-09
Subclass MXCrypto to enable work-in-progress Rust sdk

Anderas:
To avoid long-running feature branch when implementing [Matrix Rust SDK](https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk-crypto) aka Element-R, I'd like to start merging small changes into `develop` that are well isolated and cannot be shipped to production by accident. In particular all newly added code is hidden both behind a build flag as well as behind `DEBUG` compiler macro, so the new code cannot be accidentally used and shipped to production builds.

This particular PR sets up the skeleton for new rust-based implementation by overriding `MXCrypto` and all of its methods. This has both benefits and risks:
- no need to do any risky refactor across the rest of the application (for example by definining shared protocol)
- existing integration tests can still run unaltered, and validate that future Rust implementation works as expected
- potential risk: compiler does not ensure that each `MXCrypto`'s method is overriden, so if new public API was added to `MXCrypto`, this would not be automatically picked up by `MXCryptoV2`. This is a small risk, as the code is not meant for production yet, and when productionizing, more safety measures will be put in place.

Finally note this PR does not yet add any actual Rust library, merely sets the stage up.

### DIFF
--- a/MatrixSDK.xcodeproj/project.pbxproj
+++ b/MatrixSDK.xcodeproj/project.pbxproj
@@ -1791,6 +1791,8 @@
 		ED44F01528180EAB00452A5D /* MXSharedHistoryKeyManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED44F01328180EAB00452A5D /* MXSharedHistoryKeyManager.swift */; };
 		ED44F01A28180F4000452A5D /* MXSharedHistoryKeyManagerUnitTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED44F01728180F1C00452A5D /* MXSharedHistoryKeyManagerUnitTests.swift */; };
 		ED44F01B28180F4000452A5D /* MXSharedHistoryKeyManagerUnitTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED44F01728180F1C00452A5D /* MXSharedHistoryKeyManagerUnitTests.swift */; };
+		ED47CB6D28523995004FD755 /* MXCryptoV2.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED47CB6C28523995004FD755 /* MXCryptoV2.swift */; };
+		ED47CB6E28523995004FD755 /* MXCryptoV2.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED47CB6C28523995004FD755 /* MXCryptoV2.swift */; };
 		ED51943928462D130006EEC6 /* MXRoomStateUnitTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED51943828462D130006EEC6 /* MXRoomStateUnitTests.swift */; };
 		ED51943A28462D130006EEC6 /* MXRoomStateUnitTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED51943828462D130006EEC6 /* MXRoomStateUnitTests.swift */; };
 		ED51943C284630090006EEC6 /* MXRestClientStub.m in Sources */ = {isa = PBXBuildFile; fileRef = ED51943B284630090006EEC6 /* MXRestClientStub.m */; };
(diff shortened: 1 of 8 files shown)


MatrixSDK.xcodeproj/project.pbxproj
```diff
@@ -7237,6 +7243,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.10;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
+				OTHER_SWIFT_FLAGS = "-D DEBUG";
```
Anderas, line 7246: Note the addition of this flag, otherwise `DEBUG` macro in Swift is incompatible with the one in objective-c

MatrixSDK/MXSDKOptions.h
```diff
@@ -200,10 +200,22 @@ NS_ASSUME_NONNULL_BEGIN
  Enable sharing of session keys for an immediate historical context (e.g. last 10-20 messages)
  when inviting a new user to a room with shared history.
  
- @remark YES by default.
+ @remark NO by default.
```
Anderas, line 203: Unrelated change, docs were stating incorrect default value

pixlwave (approved):
Looks good to me and the use of the build setting and compiler macro sounds sane. Two minor comments:
- Should we be concerned about other library users if they also have the `DEBUG` compiler macro set? I think the answer is no, but seems worth asking in case.
- The doc comments in the Swift file should probably use `///` instead of `/** */`

Anderas:
>     * Should we be concerned about other library users if they also have the `DEBUG` compiler macro set? I think the answer is no, but seems worth asking in case.

Could you elaborate on what the worry is pls? To be clear, having `DEBUG` configuration by itself is not enough to use the new Crypto, one also has to set the build flag to true, the flag is just prevent submitting anything to production.

MatrixSDK/Data/MXRoom.m
```diff
@@ -3643,7 +3643,7 @@ - (void)validateEncryptionStateConsistency
     if (!crypto)
     {
 #ifdef MX_CRYPTO
-        MXLogFailure(@"[MXRoom] checkEncryptionState: Crypto module is not present");
+        MXLogError(@"[MXRoom] checkEncryptionState: Crypto module is not present");
```
Anderas, line 3646: There are some tests where these conditions actually occur. I do not yet know whether this is expected or not, either way the assertion will stop the tests. Temporarily switching to logging errors instead of failures and will investigate the failing tests individually.

pixlwave:
> Could you elaborate on what the worry is pls? To be clear, having `DEBUG` configuration by itself is not enough to use the new Crypto, one also has to set the build flag to true, the flag is just prevent submitting anything to production.

It was only if using `DEBUG` meant further PRs might introduce small changes that were behind the `DEBUG` configuration but not also the build flag. Seemed unlikely but just asking in case :)

manuroe (approved):
Looks smart to me :D

Anderas:
> > Could you elaborate on what the worry is pls? To be clear, having `DEBUG` configuration by itself is not enough to use the new Crypto, one also has to set the build flag to true, the flag is just prevent submitting anything to production.
> 
> It was only if using `DEBUG` meant further PRs might introduce small changes that were behind the `DEBUG` configuration but not also the build flag. Seemed unlikely but asked in case. :)

I see. Yea I do not currently anticipate such change, but will have to be careful about this.

I would actually like to get rid of the feature flag and replace it with compiler macro altogether (e.g. `MX_CRYPTO_V2`). The trouble here is that it needs to be usable from both obj-c and Swift and macros in each aren't currently mutally interchangable. The only reason DEBUG works this way is because it is manually set in build settings of the project. I may try to find another solution for this in the next PR, where the creation of the `MXCryptoV2` happens in some swift extension of `MXCrypto.m`, which means they could refer to the same Swift-only compiler macro.

PR #1500, opened 2022-06-14
MXBackgroundSyncService - Expose separate method for fetching a parti…

stefanceriu:
…cular room's read marker event without causing extra syncs.

### DIFF
--- a/MatrixSDK/Background/MXBackgroundSyncService.swift
+++ b/MatrixSDK/Background/MXBackgroundSyncService.swift
@@ -97,16 +97,18 @@ public enum MXBackgroundSyncServiceError: Error {
     /// - Parameters:
     ///   - eventId: The event identifier for the desired event
     ///   - roomId: The room identifier for the desired event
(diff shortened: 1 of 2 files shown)


MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -100,13 +100,14 @@ public enum MXBackgroundSyncServiceError: Error {
     ///   - completion: Completion block to be called. Always called in main thread.
     public func event(withEventId eventId: String,
                       inRoom roomId: String,
+                      allowSync: Bool = true,
```
ismailgulek, line 103: Can we just document this?

ismailgulek (approved):
One inline comment, otherwise LGTM!

MatrixSDK/Background/MXBackgroundSyncService.swift
stefanceriu: Yup, sorry about that 👍

PR #1501, opened 2022-06-15
Integrate MatrixSDKCrypto

Anderas:
Builds on top of experimental [`MXCryptoV2`](https://github.com/matrix-org/matrix-ios-sdk/pull/1496) and integrates independently published `MatrixSDKCrypto` CocoaPod, published as offical CocoaPod

To avoid bloating up the size of production builds as well as making them safer, the new pod will only be loaded in DEBUG builds and is currently compiled only for iOS, as some of the internal libraries in Rust Crypto cannot be compiled for macos at the moment. This will be resolved in the future but does not have to be a blocker at this point.

Finally note that the instantiation of `OlmMachine` in this PR means the integration of the pod works as expected.


MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -67,6 +68,17 @@ public class MXCryptoV2: MXCrypto {
         return nil
     }
     
+    private let machine: OlmMachine
+    public override init() {
+        do {
+            machine = try OlmMachine(userId: "", deviceId: "", path: "", passphrase: nil)
```
Anderas, line 74: `OlmMachine` is a symbol defined in `MatrixSDKCrypto` pod

manuroe (commented):
We need an `.api` changelog file to communicate on the iOS 10 drop. Then, it will be all good for me.
Exciting time!

Anderas:
> We need an `.api` changelog file to communicate on the iOS 10 drop. Then, it will be all good for me. Exciting time!

Is this in addition or instead of the existing `pr-1501.misc`?

manuroe (approved):

PR #1502, opened 2022-06-16
DevX: Add Codecov

ismailgulek:
Part of vector-im/element-ios#6306

Used the same config file with https://github.com/vector-im/element-x-ios/blob/develop/codecov.yml, as i've no shining ideas to append atm.

If i'm correct Codecov will be able to comment on subsequent PRs after the base coverage is detected when this is merged. We should be able to see the difference also at https://app.codecov.io/gh/matrix-org/matrix-ios-sdk/pulls

### DIFF
--- a/.github/workflows/ci-unit-tests.yml
+++ b/.github/workflows/ci-unit-tests.yml
@@ -56,3 +56,6 @@ jobs:
         with:
           name: MatrixSDK-macOS.xcresult
           path: build/test/MatrixSDK-macOS.xcresult/
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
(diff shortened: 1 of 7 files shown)


stefanceriu (changes requested):
Nice but I'm not sure this is going to generate the intended results. ElementX does 2 extra things:
1) Enables coverage data gathering on the main scheme https://github.com/vector-im/element-x-ios/blob/develop/ElementX/SupportingFiles/target.yml#L23
and 
2) Exports the resulting coverage data in a format that Codecov understands (cobertura) https://github.com/vector-im/element-x-ios/blob/develop/fastlane/Fastfile#L93

stefanceriu (approved):
👍

PR #1503, opened 2022-06-16
Track non-fatal issues via analytics

Anderas:
Use existing `MXAnalyticsDelegate` to track any non-fatal issues, i.e. anything logged via `MXLog.failure` or `MXLogFailure`.

The SDK itself is not responsible for tracking the issues, it merely sends relevant information via delegate. If a client does not set the delegate, or the user has not consenting to sending crash and analytics data, no issues will be tracked.


gileluard (approved):
LGTM 👍

PR #1504, opened 2022-06-17
Server Offline Activity Indicator

gileluard:
Needed for https://github.com/vector-im/element-ios/issues/5607

Added `MXNetworkMonitor` to monitor network reachability

### DIFF
--- a/MatrixSDK/Network/MXNetworkMonitor.swift
+++ b/MatrixSDK/Network/MXNetworkMonitor.swift
@@ -1,0 +1,85 @@
+// 
+// Copyright 2022 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


MatrixSDK/Network/MXNetworkMonitor.swift
```diff
@@ -0,0 +1,88 @@
...
+                MXLog.debug("[MXNetworkMonitor] is offline")
+            }
+            
+            DispatchQueue.main.async {
+                NotificationCenter.default.post(name: Self.reachabilityDidChange, object: self)
+            }
+        }
+        
+        // Needed by the SDK
+        AFNetworkReachabilityManager.shared().startMonitoring()
```
Anderas, line 78: Could you expand on why this is needed? Is it because a lot of existing SDK code relies on `AFNetworkReachabilityManager` for network reachability and it would be too much work to refactor at this point?

Anderas: I dont mind either way, it just isn't clear to me why `MXNetworkMonitor` needs to start both `NWPathMonitor` and `AFNetworkReachabilityManager`

gileluard:
Updated EI code to use exclusively `AFNetworkReachabilityManager`.

PR #1507, opened 2022-06-27
Enforce application/json mime types on bug report post requests.

stefanceriu:
This change will attempt fixing random 500 errors coming from the rageshake service when submitting bug requests.


pixlwave (approved):
LGTM, 🤞

stefanceriu:
Turns out this is unnecessary.

PR #1508, opened 2022-06-27
Handle empty pagination end token on timeline end reached

aringenbach:
Fixes https://github.com/vector-im/element-ios/issues/6347

Required since this change: https://github.com/matrix-org/synapse/pull/12903


stefanceriu (approved):
lgtm 👍

PR #1510, opened 2022-06-29
Integrate Rust-based OlmMachine to encrypt / decrypt messages

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6357

This PR is the first integration of Rust-based [OlmMachine](https://github.com/matrix-org/matrix-rust-sdk/blob/main/bindings/matrix-sdk-crypto-ffi/src/machine.rs#L73) which will eventually replace all existing Crypto code in the iOS SDK. This PR specifically takes advantage of existing `MXCryptoV2` class and does the following:

- ensure communicating users establish an olm chanel, exchange keys etc
- encrypt messages before sending to remote user
- decrypt local and incoming messages from a remote user

As a reminder, all of the Rust SDK is only available in `DEBUG` mode after toggling a feature flag so very far from production ready. Specifically some issues this PR does not yet address with encrypt / decrypt:

- ~the very first sent message fails to decrypt, suggesting we have not exchanged keys in time / properly~ Solved via TaskQueues
- ~the time to encrypt is quite slow, most likely because network requests of key exchange are blocking the encryption~ network tasks now running in parallel
- there are some unit tests added here, but the entire `CryptoMachine` class is thus far untested. I have not yet decided on a suitable testing strategy for the code that cannot be easily mocked out (`OlmMachine`). We have an existing suite of crypto integration tests which will eventually prove the overal correctness but I would like to add more specific unit tests in future PRs as well (particularly when serializing / deserializing from different formats)


MatrixSDKTests/TestPlans/UnitTests.xctestplan
```diff
@@ -38,6 +38,7 @@
         "MXBeaconInfoUnitTests",
         "MXCoreDataRoomListDataManagerUnitTests",
         "MXCredentialsUnitTests",
+        "MXCryptoRequestsUnitTests",
```
ismailgulek, line 41: We could also add this to `UnitTestsWithSanitizers`

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
```diff
@@ -0,0 +1,206 @@
...
+            deviceId: deviceId,
+            path: url.path,
+            passphrase: nil
+        )
+        
+        setLogger(logger: self)
+    }
+    
+    static func storeURL(for userId: String) throws -> URL {
+        guard let sharedContainerURL = FileManager.default.applicationGroupContainerURL() else {
```
ismailgulek, line 63: It makes sense to provide another directory for storage even if `MXSDKOptions.sharedInstance().applicationGroupIdentifier` is not set. I believe we won't need `invalidStorage` error in this case.

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
```diff
@@ -0,0 +1,206 @@
...
+            keyCounts: keyCounts,
+            unusedFallbackKeys: unusedFallbackKeys
+        )
+        
+        if let result = MXTools.deserialiseJSONString(result) as? [String: Any], !result.isEmpty {
+            log(error: "Result processing not implemented \(result)")
+        }
+    }
+    
+    func ensureOlmChanel(roomId: String, users: [String]) async throws {
```
ismailgulek, line 97: Maybe rename to `ensureOlmChannel`?

ismailgulek (approved):
Couple of inline comments, otherwise LGTM!

MatrixSDK/JSONModels/MXJSONModels.m
```diff
@@ -1063,11 +1063,21 @@ + (id)modelFromJSON:(NSDictionary *)JSONDictionary
 
 #pragma mark - Crypto
 
+@interface MXKeysUploadResponse ()
+
+/**
+ The original JSON used to create the response model
+ */
+@property (nonatomic, strong) NSDictionary *responseJSON;
```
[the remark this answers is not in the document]
Anderas: In practice this should not occur (recieved HTTP requests are instantly deserialized in a way that I would prefer not to use the `MXJSONModels` at all), but it is probably safer and clearer to change 👍

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
```diff
@@ -0,0 +1,206 @@
+//
+//  CryptoBridge.swift
+//  MatrixSDK
+//
+//  Created by Element on 05/05/2022.
+//
```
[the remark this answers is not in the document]
Anderas: Thx, yea I created these (and previous files) as part of Riot project => changed

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
Anderas: Good point, the existing `MXFileStore` and `RealmCryptoStore` use different alternative folders, so here I am just using the user's documents directory as an alternative

MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -101,33 +158,119 @@ public class MXCryptoV2: MXCrypto {
...
+        guard isRoomEncrypted(roomId) else {
+            MXLog.error("[MXCryptoV2] encryptEventContent: attempting to encrypt event in room without encryption")
+            return nil
+        }
+        
+        MXLog.debug("[MXCryptoV2] encryptEventContent: Encrypting content")
+        
+        Task {
+            do {
+                let users = try await getRoomUserIds(for: room)
```
BillCarsonFr, line 184: I always wonder if you should fail sending completly if this fails. 
Is it using cache? or always doing a request?

MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -101,33 +158,119 @@ public class MXCryptoV2: MXCrypto {
...
+        guard let event = event else {
+            MXLog.debug("[MXCryptoV2] Missing event")
+            return nil
+        }
+        do {
+            return try machine.decryptEvent(event)
+        } catch {
+            MXLog.error("[MXCryptoV2] decryptEvent: \(error)")
+            let result = MXEventDecryptionResult()
+            result.error = error
```
BillCarsonFr, line 219: we will need a way to map this to existing errors (will help to be consistent with existing posthog events)

MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -101,33 +158,119 @@ public class MXCryptoV2: MXCrypto {
...
+        let results = events?.compactMap {
+            decryptEvent($0, inTimeline: timeline)
+        }
+        onComplete?(results)
     }
     
     public override func ensureEncryption(inRoom roomId: String!, success: (() -> Void)!, failure: ((Error?) -> Void)!) -> MXHTTPOperation! {
-        warnNotImplemented()
-        return nil
+        guard let roomId = roomId, let room = session?.room(withRoomId: roomId) else {
```
BillCarsonFr, line 232: I think we need some synchronization here.
I believe android is using a lock per roomID to ensure that only one of those is going on at a time per room

MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -101,33 +158,119 @@ public class MXCryptoV2: MXCrypto {
...
+    public override func handle(_ syncResponse: MXSyncResponse!) {
+        do {
+            try machine.handleSyncResponse(
+                toDevice: syncResponse.toDevice,
+                deviceLists: syncResponse.deviceLists,
+                deviceOneTimeKeysCounts: syncResponse.deviceOneTimeKeysCount ?? [:],
+                unusedFallbackKeys: syncResponse.unusedFallbackKeys
+            )
+        } catch {
+            MXLog.error("[MXCryptoV2] handleSyncResponse: \(error)")
```
BillCarsonFr, line 270: not sure we can swallow this. The all sync should fail

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
```diff
@@ -0,0 +1,221 @@
...
+        if let result = MXTools.deserialiseJSONString(result) as? [String: Any], !result.isEmpty {
+            log(error: "Result processing not implemented \(result)")
+        }
+    }
+    
+    func ensureOlmChannel(roomId: String, users: [String]) async throws {
+        try await getMissingSessions(users: users)
+        try await shareRoomKey(roomId: roomId, users: users)
+        
+        let requests = try machine.shareRoomKey(roomId: roomId, users: users)
```
BillCarsonFr, line 116: why 2 shareRoomKey calls?

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
```diff
@@ -0,0 +1,221 @@
...
+            unusedFallbackKeys: unusedFallbackKeys
+        )
+        
+        if let result = MXTools.deserialiseJSONString(result) as? [String: Any], !result.isEmpty {
+            log(error: "Result processing not implemented \(result)")
+        }
+    }
+    
+    func ensureOlmChannel(roomId: String, users: [String]) async throws {
+        try await getMissingSessions(users: users)
```
BillCarsonFr, line 113: need some synchronization? Android has a keyClaimLock

BillCarsonFr (commented):
LGTM, some comments around synchronization. EA has some locks on some operations

MatrixSDK/Crypto/MXCryptoV2.swift
Anderas: Adding this to my TODOs. Given that I've merged Sentry now, these will be primarily tracked there, and after implementing further [error details](https://github.com/matrix-org/matrix-rust-sdk/issues/734), we can observe the tracked data and decide how to map to posthog

MatrixSDK/Crypto/MXCryptoV2.swift
Anderas: Yea, added new mechanism to synchronize `async` tasks

MatrixSDK/Crypto/V2/MXCryptoMachine.swift
Anderas: Added

MatrixSDK/Utils/MXTaskQueue.swift
```diff
@@ -0,0 +1,72 @@
...
+/// before starting the next block of async tasks.
+///
+/// Swift concurrency treats each `await` as a potential suspension point meaning there is no guarantee that a group of related `await` tasks
+/// will be completed in full before another group is started. This is also the reason why `Actor`s are designed for
+/// [re-entrancy](https://github.com/apple/swift-evolution/blob/main/proposals/0306-actors.md#actor-reentrancy).
+///
+/// The solution to this problem is using serial task queues where work is scheduled, but only executed once all of the previously
+/// scheduled tasks have completed. This is an analogous mechanism to using serial `DispatchQueue`s.
+@available(iOS 13.0.0, macOS 10.15.0, *)
+public actor MXTaskQueue {
```
Anderas, line 31: @ismailgulek I had to add this whole new concept to synchronize a group of `async` tasks, could you pls re-review?

pixlwave (approved):
Reviewing b8ca763 through 5889499:

Looks good to me 👍 - shame there isn't such a mechanism already built it.

BillCarsonFr (approved):

MatrixSDKTests/Utils/MXTaskQueueUnitTests.swift
```diff
@@ -0,0 +1,333 @@
...
+    
+    private func doSomeHeavyWork(timeInterval: TimeInterval = 0.1) async {
+        do {
+            try await Task.sleep(nanoseconds: UInt64(timeInterval * 1e9))
+        } catch {
+            XCTFail("Error sleeping \(error)")
+        }
+    }
+    
+    private func XCTAssertAllOperationsPerformed(_ taskIds: [String], file: StaticString = #file, line: UInt = #line) async {
```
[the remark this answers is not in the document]
Anderas: Indeed, thanks

PR #1511, opened 2022-06-29
Fix MXCall answer not being sent to server in some cases

langleyd:
Resolves https://github.com/vector-im/element-ios/issues/6359

### DIFF
--- a/MatrixSDK/VoIP/MXCall.m
+++ b/MatrixSDK/VoIP/MXCall.m
@@ -161,7 +161,7 @@ NSString *const kMXCallSupportsTransferringStatusDidChange = @"kMXCallSupportsTr
 
         callStackCall.delegate = self;
 
-        callStackCallDispatchQueue = dispatch_queue_create("callStackCallDispatchQueue", DISPATCH_QUEUE_SERIAL);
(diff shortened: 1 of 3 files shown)


langleyd:
This is a draft PR as the fix needs to be verified on a particular deployment first.

aringenbach:
The issue has been narrowed down to the call to WebRTC that sets the remoteDescription. This call is actually never made when the issue occurs. 
MXCall notably uses the main operationQueue in order to handle calls which means that something else might have been interfering. Moving all calls operation to a dedicated operationQueue has fixed the issue for the affected users.

MatrixSDK/VoIP/MXCall.m
```diff
@@ -1124,10 +1129,14 @@ - (void)handleCallInvite:(MXEvent *)event
     [callStackCallOperationQueue addOperationWithBlock:^{
         MXStrongifyAndReturnIfNil(self);
         
+        MXLogDebug(@"[MXCall][%@] processing block", self.callId)
```
ismailgulek, line 1132: Can we make this comment more explanatory?

ismailgulek (approved):
One inline comment, otherwise LGTM!

ismailgulek (approved):

PR #1513, opened 2022-07-01
Detect multiple valid SSSS keys

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/4569

An issue has been reported aprox a year ago with multiple valid SSSS keys and the iOS picking up the wrong one when authenticating (i.e. asking only for passkey, not passphrase). The issue has not been reported since then and could not be reproduced, so it isn't obvious whether it still occurs.

Based on the [initial report](https://github.com/vector-im/element-ios/issues/4569) the iOS did have a valid SSSS key and `m.secret_storage.default_key` was correctly pointing to it (otherwise the authentication would not ask for passphrase instead of passkey), but the client had reportedly more than one valid key. This means that the client has either failed to re-point `m.secret_storage.default_key` to the more recent key, or failed to delete / invalidate the previous key. Note that SSSS keys are never fully deleted, but remain in the local / server database, albeit with nil-ed content.

As it isn't possible to reproduce this issue, I decided to add some extra logs to help with detecting this issue in the future. Namely assuming that at most one valid SSSS key should be present at any given point (client can also have zero keys), the client will log an error if it detects more than one.

### DIFF
--- a/MatrixSDK/Crypto/Recovery/MXRecoveryService.m
+++ b/MatrixSDK/Crypto/Recovery/MXRecoveryService.m
@@ -87,7 +87,7 @@ NSString *const MXRecoveryServiceErrorDomain = @"org.matrix.sdk.recoveryService"
     MXSecretStorageKeyContent *keyContent = [_secretStorage keyWithKeyId:self.recoveryId];
     if (!keyContent)
     {
-        // No recovery at all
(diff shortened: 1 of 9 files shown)


aringenbach (approved):
LGTM 👍

BillCarsonFr (approved):
LGTM

PR #1515, opened 2022-07-06
Delight edit layout experiment

gileluard:
needed for https://github.com/vector-im/element-ios/issues/6079, https://github.com/vector-im/element-ios/issues/6406, https://github.com/vector-im/element-ios/issues/6407, https://github.com/vector-im/element-ios/issues/6409


MatrixSDK/Data/RoomList/Common/MXRecentsRoomListDataFetcher.swift
```diff
@@ -0,0 +1,158 @@
...
+        for event in events {
+            if let type = event["type"] as? String,
+                type == kMXAccountDataTypeBreadcrumbs,
+                let content = event["content"] as? [AnyHashable: [String]] {
+                breadcrumbs = content
+            }
+        }
+        
+        guard var recentRoomIds = breadcrumbs?[kMXAccountDataTypeRecentRoomsKey] else {
+            MXLog.error("[MXRecentsRoomListDataFetcher] cannot retrieve recent rooms")
```
pixlwave, line 97: I wonder if this (and the one above should be `warn` instead of `error` from the perspective of whether its the kind of thing we want to report to Sentry? Presumably every new account would do this.

MatrixSDK/Data/MXRoomSummaryDataTypes.h
```diff
@@ -26,7 +26,10 @@ typedef NS_OPTIONS(NSInteger, MXRoomSummaryDataTypes)
     MXRoomSummaryDataTypesServerNotice = 1 << 4,
     MXRoomSummaryDataTypesHidden = 1 << 5,
     MXRoomSummaryDataTypesSpace = 1 << 6,
-    MXRoomSummaryDataTypesConferenceUser = 1 << 7
+    MXRoomSummaryDataTypesConferenceUser = 1 << 7,
+    MXRoomSummaryDataTypesNotDirect = 1 << 8,
+    MXRoomSummaryDataTypesUnread = 1 << 9,
+    MXRoomSummaryDataTypesRead = 1 << 10,
```
pixlwave, line 32: It feels a bit odd to have `Direct`/`NotDirect` & `Unread`/`Read` in here when the omission of one implies the other, but I haven't had a look at the Element PR yet to see the full context.

pixlwave (approved):
Looks good to me, some minor comments inline. What this PR is missing however is a changelog :)

MatrixSDK/Data/MXRoomSummaryDataTypes.h
pixlwave: Ok, I see the use now. Logically my brain would have gone for adding `Direct` to `notDataTypes` instead of `NotDirect` but the way this works in the `RecentsListService` is much more readable 🙂.

gileluard: I've removed unnecessary `MXRoomSummaryDataTypesNotDirect` and `MXRoomSummaryDataTypesRead ` types

pixlwave (approved):
LGTM 😎

PR #1516, opened 2022-07-08
Integration tests should wait until the room is ready

Anderas:
Some of the flaky integration tests are failing because they manually start a session and try to get a room immediately. The problem with that approach is not only that the room may not be created at that point, but even if it is, not all state may be properly updated in the room summary.

The (partial) solution to this is to start sessions and then listen to global notifications that announce room creation or sync completeion. It is not 100% reliable, and requires a bit of manual delay, but it seems to be more reliable than the current setup.

Additionally I decided to enable the entire suite of `MXCryptoTests` and disable in code those that do not pass. This makes it more obvious when reading the test code and automatically adds any new tests to the test plan.


MatrixSDKTests/MXCryptoTests.m
```diff
@@ -3422,6 +3428,65 @@ - (void)testIsRoomSharingHistory
...
+/**
+ Manually restart the session and wait until a given room has finished syncing all state
+ 
+ Note: there is a lot of state update and sync going on when the session is started,
+ and integration tests often assume given state before it has finished updating. To solve
+ that this helper method makes the best guesses by observing global notifications
+ and adding small delays to ensure all updates have really completed.
+ */
+- (void)restartSession:(MXSession *)session
+    waitingForRoomId:(NSString *)roomId
```
pixlwave, line 3442: The indentation is a little out of on this and the 2 calls to it.

MatrixSDKTests/MXCryptoTests.m
```diff
@@ -3422,6 +3428,65 @@ - (void)testIsRoomSharingHistory
...
+                    // Now wait until this room reports sync completion
+                    observer = [[NSNotificationCenter defaultCenter] addObserverForName:kMXRoomInitialSyncNotification
+                                                                                 object:nil
+                                                                                  queue:[NSOperationQueue mainQueue]
+                                                                             usingBlock:^(NSNotification * notification) {
+                        [[NSNotificationCenter defaultCenter] removeObserver:observer];
+                        
+                        // Even when sync completed, there are actually still a few async updates that happen (i.e. the notification
+                        // fires too early), so have to add some small arbitrary delay.
+                        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
```
pixlwave, line 3472: Kind of annoying to have to stall for a whole second, but can see why it is needed 🙁

MatrixSDKTests/MXCryptoTests.m
```diff
@@ -3422,6 +3428,65 @@ - (void)testIsRoomSharingHistory
...
+             failure:(void (^)(NSError *))failure
+{
+    __block id observer;
+    
+    // First start the session
+    [session start:^{
+        
+        // Wait until we know that the room has actually been created
+        observer = [[NSNotificationCenter defaultCenter] addObserverForName:kMXSessionNewRoomNotification
+                                                          object:nil
```
pixlwave, line 3453: Indentation here also.

pixlwave (approved):
LGTM 👍

pixlwave:
Oh, but missing a changelog, I forgot to add that to my review.

MatrixSDKTests/MXCryptoTests.m
Anderas: Yea it is indeed. I would fix the notification, but then have no test to proove that I did not break something else, so this should be passing first before any changes to production code.

PR #1517, opened 2022-07-11
Log errors with details in analytics

Anderas:
Enabling analytics tracking for `MXLog.error`, where previously only `MXLog.failure` where tracked. To limit the amount of noise that would result from tracking errors with variables in the string we are only tracking those errors that manually specify a separate `details` dictionary. This will be changed in the future to either log all errors by default or convert them all to detail-based events


gileluard (approved):
lgtm 👍

PR #1519, opened 2022-07-12
Fix non-deterministic unit test

Anderas:


MatrixSDKTests/Utils/MXTaskQueueUnitTests.swift
```diff
@@ -301,18 +310,33 @@ class MXTaskQueueUnitTests: XCTestCase {
...
     
+    /// Assert that for every task a given operation occurs before another operation
+    private func XCTAssertOperationBefore(_ taskIds: [String], operation: Operation.Kind, before: Operation.Kind, file: StaticString = #file, line: UInt = #line) async {
+        for id in taskIds {
+            let realOrder = await timeline.operationOrder(for: id)
+            guard let index = realOrder.firstIndex(of: operation), let beforeIndex = realOrder.firstIndex(of: before) else {
+                XCTFail("Cannot find given operations", file: file, line: line)
+                return
+            }
+            XCTAssertLessThan(index, beforeIndex, "Operation \(operation) does not happen before \(before)", file: file, line: line)
```
pixlwave, line 335: Naming wise, `beforeIndex` should come after `index` is a bit hard to parse. Maybe more helpful to define something like `first`/`secondOperation` and `first`/`secondIndex` or similar.

pixlwave (approved):
LGTM

PR #1525, opened 2022-07-13
MXRoom: Support reply to beacon info event (PSG-4)

SBiOSoftWhare:
Part of https://github.com/vector-im/element-ios/issues/6423

### DIFF
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -2172,22 +2172,18 @@ NSInteger const kMXRoomInvalidInviteSenderErrorCode = 9002;
         NSString *question = [MXEventContentPollStart modelFromJSON:eventToReply.content].question;
         
         senderMessageBody = question;
-        senderMessageFormattedBody = senderMessageBody;
     }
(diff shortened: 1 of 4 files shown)


MatrixSDK/Data/MXRoom.m
```diff
@@ -2158,46 +2158,37 @@ - (void)getReplyContentBodiesWithEventToReply:(MXEvent*)eventToReply
...
     NSString *senderMessageBody;
     NSString *senderMessageFormattedBody;
     
-    if (eventToReply.location)
+    if (eventToReply.eventType == MXEventTypePollStart)
+    {
+        NSString *question = [MXEventContentPollStart modelFromJSON:eventToReply.content].question;
+        
+        senderMessageBody = question;
+        senderMessageFormattedBody = senderMessageBody;
```
aringenbach, line 2175: If this happen on several other cases, I suppose we could add a specific setter that updates both

aringenbach (approved):
LGTM

MatrixSDK/Data/MXRoom.m
SBiOSoftWhare: Good catch I simplified the code

PR #1526, opened 2022-07-14
Enable more crypto integration tests

Anderas:


pixlwave (approved):
LGTM, sorry I was a bit slow to notice this.

Anderas:
> LGTM, sorry I was a bit slow to notice this.

I don't think 2hrs since raising the PR is slow 😄 . Will actually have to disable a few of the failing ones first

Anderas:
@pixlwave I cheated a little and made some of the tests pass by increasing the wait time from 1 to 5 seconds. It is not ideal, but at this point I'd rather have them pass at all, before making them pass quick, otherwise I won't be able to validate some of the incoming crypto refactors.

PR #1528, opened 2022-07-18
Convert verification request and transaction to protocols

Anderas:
As part of an ongoing work to implement Rust-based [device verification](https://github.com/vector-im/element-ios/issues/6443) I need to prepare the ground by converting existing `MXKeyVerificationRequest`, `MXKeyVerificationTransaction` and `MXSASTransaction` classes into protocols, so that future PRs can implement these protocols with `MatrixCryptoSDK` objects. This is not a full list of classes that need to be hidden behind an interface, only those directly needed for SAS flow.

I have previously taken a [subclassing approach](https://github.com/matrix-org/matrix-ios-sdk/pull/1496) to solve a similar problem, but this was because refactoring existing code for `MXCrypto` class was too risky without extensive test coverage and it was best to leave this untouched. Swapping request and transaction objects for protocols should be safer, and now validated by some newly enabled [integration tests](https://github.com/matrix-org/matrix-ios-sdk/pull/1526).

Regarding naming, I considered both alternatives where protocol adopts existing name (e.g. `MXKeyVerificationRequest`) and subclasses extend this (e.g. Default, V2 ...), as well as naming the protocol `MXKeyVerificationRequestProtocol` and leaving the implementation names intact. Reviewing the SDK I did not find many examples of adding `Protocol` to protocol definitions so I opted for the first naming variant to preserve consistency. This choice also means that only objective-c files need to perform renames, whereas swift files remain unchanged.


MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
```diff
@@ -63,13 +63,13 @@ @interface MXKeyVerificationManager ()
     dispatch_queue_t cryptoQueue;
 
     // All running transactions
-    MXUsersDevicesMap<MXKeyVerificationTransaction*> *transactions;
+    MXUsersDevicesMap<MXDefaultKeyVerificationTransaction *> *transactions;
```
Anderas, line 66: `MXKeyVerificationManager` is accessing some of the internal methods of request and transaction which are defined in `_Private.h` header files and not exposed in the protocol. For this reason it is still using implementation types rather than protocol types

stefanceriu (approved):
Looks good to me 👍

PR #1529, opened 2022-07-18
Fix formatted_body content for unformatted events

aringenbach:
Fixes https://github.com/vector-im/element-ios/issues/6446

The formatted_body content is now properly omitted if there is no actual formatting within the message (same behaviour as Web).

### DIFF
--- a/MatrixSDK/Aggregations/MXAggregatedEditsUpdater.m
+++ b/MatrixSDK/Aggregations/MXAggregatedEditsUpdater.m
@@ -103,7 +103,7 @@
 
             // This enables editing replies that don't provide a fallback mx-reply body.
             compatibilityText = [NSString stringWithFormat:@"* %@", text];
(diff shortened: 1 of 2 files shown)


MatrixSDK/Aggregations/MXAggregatedEditsUpdater.m
```diff
@@ -109,7 +109,10 @@ - (MXHTTPOperation*)replaceTextMessageEvent:(MXEvent*)event
     else
     {
         compatibilityText = [NSString stringWithFormat:@"* %@", text];
-        compatibilityFormattedText = [NSString stringWithFormat:@"* %@", formattedText];
+        if (formattedText)
```
stefanceriu, line 112: Is it worth doing this check on the `Fail to parse reply event` branch and also checking for empty strings?

PR #1531, opened 2022-07-18
Crypto user and device identity

Anderas:
Add mechanisms to acquire `MXDeviceInfo`, `MXCrossSigningInfo` and other related objects using `MatrixSDKCrypto` and bootstrap cross-signing keys. 

This is another pre-requisite for having all of the required pieces to implement device verification.

The (interrelated) changes in this PR:
- Implement device and cross signing info sources that will return `MatrixSDK` objects from `MatrixSDKCrypto` definitions
- create two wrappers that allow the creation of `MXDeviceInfo` and `MXCrossSIgningInfo` from swift code even if some of the API is not exposed to swift, but is in `_Private.h` header files
- Add `MXCrossSigningV2` (a subclass of `MXCrossSigning`) which overrides all functionality, similar to [`MXCryptoV2`](https://github.com/matrix-org/matrix-ios-sdk/pull/1496)
- Hook up cross signing with existing infrastructure to upload cross signing keys when new user logs in
- Add `MXNamedLog` to avoid verbose and repeating logs, this log will always print logs in the "[Name] function: message" format
- ~Run iOS tests instead of macos tests, to run unit tests containing `MatrixSDKCrypto` objects (currently only compiled for iOS), and without loosing anything mac-os specific~ Leaving tests on macos target, will try to recompile MatrixSDKCrypto for all platforms asap


MatrixSDKTests/TestPlans/UnitTests.xctestplan
```diff
@@ -13,15 +13,15 @@
       "targets" : [
         {
           "containerPath" : "container:MatrixSDK.xcodeproj",
-          "identifier" : "B14EF1C72397E90400758AF0",
-          "name" : "MatrixSDK-macOS"
+          "identifier" : "32C6F92C19DD814400EA4E9C",
+          "name" : "MatrixSDK-iOS"
```
Anderas, line 17: I may be missing some context here why we ran unit tests on macos instead of iOS. If it is important to keep as it was, I can instead add another test target with iOS-specific tests only. I will also try to package the MatrixCryptoSDK to run on macos so that this issue can be avoided altogether

MatrixSDK/Utils/MXLog.swift
```diff
@@ -194,3 +194,40 @@ private var logger: SwiftyBeaver.Type = {
         return "\(message) - \(details)"
     }
 }
+
+/// Convenience wrapper around `MXLog` which formats all logs as "[Name] function: <message>"
+struct MXNamedLog {
```
SBiOSoftWhare, line 199: Interesting! If we want global usage of same log naming we can also make a `MXLogModule` like:
```
public enum LogModule: String {
  case webService
  case parsing
  case database
  case geolocation
}

extension LogModule {
  
  // MARK: Public

    func debug(_ message: String, function: String = #function) {
         MXLog.debug(formattedMessage(message, function: function))
     }
  
  // MARK: Private
    private func formattedMessage(_ message: String, function: String) -> String {
         return "[\(self.rawValue))] \(function): \(message)"
     }
}

```
And then use it like: `MXLogModule.geolocation.debug("Test")`

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -245,10 +344,25 @@ extension MXCryptoMachine: MXCryptoEventEncrypting {
...
+extension MXCryptoMachine: MXCryptoCrossSigning {
+    func crossSigningStatus() -> CrossSigningStatus {
+        return machine.crossSigningStatus()
+    }
+    
+    func bootstrapCrossSigning(authParams: [AnyHashable: Any]) async throws {
+        let result = try machine.bootstrapCrossSigning()
+        try await requests.uploadSigningKeys(request: result.uploadSigningKeysRequest, authParams: authParams)
+        try await requests.uploadSignatures(request: result.signatureRequest)
+    }
```
SBiOSoftWhare, line 363: You can parallelize if possible:
```
         async let uploadSigningKeys = requests.uploadSigningKeys(request: result.uploadSigningKeysRequest, authParams: authParams)
         async let uploadSignatures = requests.uploadSignatures(request: result.signatureRequest)
         
         await (uploadSigningKeys, uploadSignatures)
```

SBiOSoftWhare (approved):
LGTM, I just added some small suggestions

MatrixSDK/Utils/MXLog.swift
Anderas: Yea that sounds good, thought perhaps yet another use case, as it assumes a smaller number of stable areas, as your examples above. In my case the context equals any given class, which would be too many areas to keep a track of in an enum. I think both these use cases are complimentary

PR #1532, opened 2022-07-19
Outgoing SAS User Verification Flow

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6443

Implement (outgoing) verification flow using SAS / emojis, incl sending initial request, responding via SAS methods, accepting or rejecting matching emojis and displaying the verification status of user after verification is complete.

## Changes in the PR
- Add `MXKeyVerificationManagerV2` which uses Rust-based crypto machine under the hood
- Expose `OlmMachine` verification methods under `MXCryptoVerification`
- Hide some of the [previously exposed](https://github.com/matrix-org/matrix-ios-sdk/pull/1528) protocol methods inside legacy implementations, as they are not / shoud not be needed by every verification implementation
- Implement Rust-based verification request and sas transactions that can be used with existing views and view models

## Missing bits
- Some methods are still not implemented, which is signalled via added logs. It is not yet clear whether these methods are really needed or can be hidden inside legacy implementations
- Incoming flow is not yet implemented, i.e. another device asking to verify does not currently hook up to the Crypto Machine
- Verification only uses direct room mechanism, and assumes DM with a given user already exists

All of these will be resolved in future PRs

### DIFF
--- a/MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
+++ b/MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
@@ -1,0 +1,238 @@
+//
+//  MXKeyVerificationManagerV2.swift
+//  MatrixSDK
+//
+//  Created by Element on 05/07/2022.
+//
+
+import Foundation
+
+#if DEBUG && os(iOS)
+
+import MatrixSDKCrypto
+
+@available(iOS 13.0.0, *)
+class MXKeyVerificationManagerV2: MXKeyVerificationManager {
+    typealias GetOrCreateDMRoomId = (_ userId: String) async throws -> String
+    
+    override var requestTimeout: TimeInterval {
+        set {
+            log.debug("Not implemented")
+        }
+        get {
+            log.debug("Not implemented")
+            return 1000
+        }
+    }
+    
+    private let verification: MXCryptoVerification
+    private let getOrCreateDMRoomId: GetOrCreateDMRoomId
+    
+    private var requests: [MXKeyVerificationRequestV2]
+    private var transactions: [MXSASTransactionV2]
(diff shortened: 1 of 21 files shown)


MatrixSDK/Crypto/Verification/Requests/MXKeyVerificationRequest.h
```diff
@@ -68,39 +68,28 @@ Accept an incoming key verification request.
...
-
-// Current state
 @property (nonatomic, readonly) MXKeyVerificationRequestState state;
 
-// Original data for this request
-@property (nonatomic, readonly) MXEvent *event;
+/**
+ The cancellation reason, if state is cancelled
+ */
+@property (nonatomic, readonly, nullable) MXTransactionCancelCode *reasonCancelCode;
```
Anderas, line 78: Here and in the other protocols I am either moving required methods from the generic protocol into default implementation (when not required by every implementation) or changing them to be readonly. No methods are added or removed

MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
```diff
@@ -0,0 +1,236 @@
...
+            guard let req = verification.verificationRequest(userId: request.otherUser, flowId: request.requestId) else {
+                continue
+            }
+            request.update(request: req)
+        }
+        
+        for transaction in transactions {
+            guard let verification = verification.verification(userId: transaction.otherUserId, flowId: transaction.transactionId) else {
+                continue
+            }
```
aringenbach, line 57: Perhaps it might be useful to log something if we don't get through these guards ?

aringenbach (approved):
LGTM

PR #1533, opened 2022-07-19
Wait for all verification requests in integration tests

Anderas:
Change existing key verification tests in a way that each side awaits incoming key request rather than assuming the incoming order (and hacking a delay time)


pixlwave (approved):
LGTM. Maybe needs a changelog (but presuming its covered by an existing one).

Anderas:
> LGTM. Maybe needs a changelog (but presuming its covered by an existing one).

Yea this will be an ongoing effort so don't think changelog per PR is useful in this case

PR #1539, opened 2022-08-01
Send an empty dictionary when calling /join to be spec compliant.

pixlwave:
The SDK isn't sending spec-complaint joins, and Synapse is about to stop handling these: https://github.com/matrix-org/synapse/issues/13388.

Fixes https://github.com/vector-im/element-ios/issues/6481.


ismailgulek (approved):
LGTM!

PR #1540, opened 2022-08-02
Rooms without an unread badge show up in "unread" tab

gileluard:
resolves https://github.com/vector-im/element-ios/issues/6511

- Excluded room summaries without notifications from unread list


pixlwave (approved):
LGTM 😎

PR #1545, opened 2022-08-08
MXBeaconAggregations: Add debug log when MXBeaconInfoSummary associated to a stopped beacon info is not found

SBiOSoftWhare:


aringenbach (approved):
LGTM

PR #1547, opened 2022-08-10
Stop using the room ID from the space child state events.

pixlwave:
This property was removed in Synapse 1.65 as it was going against the spec.

For more info see https://github.com/matrix-org/synapse/pull/13365 and https://github.com/vector-im/element-ios/issues/6547.


manuroe (approved):
LGTM

clokep:
> For more info see [matrix-org/synapse#13365](https://github.com/matrix-org/synapse/pull/13365)

And vector-im/element-ios#5144.

PR #1553, opened 2022-08-17
Fix `kickUser` to set `"leave"` membership

neilalexander:
There is no such thing as a `"kick"` membership in the spec. The fact that Synapse aliases `"kick"` to `"leave"` is unspecced behaviour.


stefanceriu (approved):
lgtml 👍  as per https://spec.matrix.org/latest/client-server-api/#mroommember

PR #1557, opened 2022-08-22
Fix setting a log level of .none.

### LINKED ISSUE #1550: Change verbosity of MXLog
I have

```
        var config = MXLogConfiguration()
        config.logLevel = .none
        MXLog.configure(config)
```

before calling any MatrixSDK methods. Still, I'm seeing tons of `MXSyncResponseFileStore`, `MXHTTPClient` and `MXFileStore` log outputs. How does one properly configure this?

pixlwave:
Fixes #1550. There was a bug where setting the log level to `.none` had no effect if a call to `MXLogDebug` (etc) had already been called as the default configuration would already have been applied.


stefanceriu (approved):
👍

PR #1558, opened 2022-08-22
Track all errors in Sentry

Anderas:
Following up on https://github.com/matrix-org/matrix-ios-sdk/pull/1517, which tracked only some of `MXLogError` and `MXLogFailure` in analytics (Sentry for element-ios, assuming user consent), this PR extends this approach across the entire SDK and makes each error / failure trackable. This requires a number of changes:

- add `AnalyticsDestination` to `MXLog` which sends errors to `AnalyticsDelegate`. Previous implementation simply hooked into the existing error methods and was a little hacky. Using destinations is the preffered approach
- error and failure logs now require `StaticString` instead of `String`, meaning no variables can be present in the message. This may sound annoying but will make tracing of parameters more explicit by:
    - not accidentally leaking sensitive data
    - grouping errors by message without variable components
- replace `details` argument with existing `context` argument that can be of any type. This is to both reduce the number of arguments as well as stay consistent with SwiftyBeaver
- extract all variables from logs and pass them in the context, sometimes as a dictionary of values (this is the largest number of changes)


MatrixSDK/Crypto/KeyBackup/MXKeyBackup.m
```diff
@@ -153,7 +153,8 @@ - (void)checkAndStartWithKeyBackupVersion:(nullable MXKeyBackupVersion*)keyBacku
             Class<MXKeyBackupAlgorithm> algorithmClass = AlgorithmClassesByName[keyBackupVersion.algorithm];
             if (algorithmClass == NULL)
             {
-                MXLogError(@"[MXKeyBackup] checkAndStartWithKeyBackupVersion: unknown algorithm: %@", keyBackupVersion.algorithm);
+                NSString *message = [NSString stringWithFormat:@"[MXKeyBackup] checkAndStartWithKeyBackupVersion: unknown algorithm: %@", keyBackupVersion.algorithm];
```
Anderas, line 156: In some cases the variable is one of very few and it is okay making it a part of the message itself (objc has no `StaticString` equivalent, so this works)

MatrixSDK/Crypto/MXCrypto.m
```diff
@@ -653,7 +661,12 @@ - (MXEventDecryptionResult *)decryptEvent2:(MXEvent *)event inTimeline:(NSString
         result = [alg decryptEvent:event inTimeline:timeline];
         if (result.error)
         {
-            MXLogError(@"[MXCrypto] decryptEvent: Error for %@: %@\nEvent: %@", event.eventId, result.error, event.JSONDictionary);
+            NSDictionary *details = @{
```
Anderas, line 664: Annoyingly objc will crash if we ever add `nil` value to a dictionary, meaning it is safer to always default to something rather than risk logs crashing the app

MatrixSDK/Data/MXRoomSummary.m
```diff
@@ -552,7 +554,9 @@ - (void)setIsEncrypted:(BOOL)isEncrypted
     // This should never happen
     if (_isEncrypted && !isEncrypted)
     {
-        MXLogError(@"[MXRoomSummary] setIsEncrypted: Attempt to reset isEncrypted for room %@. Ignote it. Call stack: %@", self.roomId, [NSThread callStackSymbols]);
```
pixlwave, line 555: I'm presuming we don't need to worry about call stacks now that these go to Sentry.

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -993,7 +993,7 @@ - (RoomReceiptsStore*)getOrCreateRoomReceiptsStore:(NSString *)roomId
                         @"roomId": roomId ?: @"",
                         @"exception": exception
                     };
-                    MXLogErrorWithDetails(@"[MXFileStore] Warning: loadReceipts file for room as been corrupted", logDetails);
+//                    MXLogErrorWithDetails(@"[MXFileStore] Warning: loadReceipts file for room as been corrupted", logDetails);
```
pixlwave, line 996: Is this intentionally left commented?

MatrixSDK/Utils/Logs/AnalyticsDestination.swift
```diff
@@ -0,0 +1,56 @@
...
+//  MatrixSDK
+//
+//  Created by Element on 22/08/2022.
+//
+
+import Foundation
+import SwiftyBeaver
+
+/// SwiftyBeaver log destination that sends errors to analytics tracker
+class AnalyticsDestination: BaseDestination {
```
pixlwave, line 12: Maybe this should be `MXAnalyticsDestination`?

MatrixSDK/Utils/Logs/MXLog.h
```diff
@@ -32,18 +32,20 @@
...
+    [MXLogObjcWrapper logFailure:message file:@__FILE__ function:[NSString stringWithFormat:@"%s", __FUNCTION__] line:__LINE__ context:nil]; \
 }
 
-#define MXLogFailureWithDetails(message, dictionary) { \
-    [MXLogObjcWrapper logFailure:message details:dictionary file:@__FILE__ function:[NSString stringWithFormat:@"%s", __FUNCTION__] line:__LINE__]; \
+#define MXLogFailureDetails(message, details) { \
+    [MXLogObjcWrapper logFailure:message file:@__FILE__ function:[NSString stringWithFormat:@"%s", __FUNCTION__] line:__LINE__ context:details]; \
 }
+
+// TODO: failure context
```
pixlwave, line 51: Is that todo different to the `MXLogFailure`s above?

MatrixSDK/Utils/Logs/MXLog.swift
```diff
@@ -108,39 +108,65 @@ private var logger: SwiftyBeaver.Type = {
...
+        function: String,
+        line: Int,
+        context: Any? = nil
+    ) {
+        logger.error(message, file, function, line: line, context: context)
+    }
+    
+    /// Log failure with additional details
+    ///
+    /// A failure is any type of programming error which should never occur in production. In `DEBUG` confuguration
```
pixlwave, line 139:
```suggestion
    /// A failure is any type of programming error which should never occur in production. In `DEBUG` configuration
```

MatrixSDK/Utils/Logs/MXLog.swift
```diff
@@ -183,52 +211,39 @@ private var logger: SwiftyBeaver.Type = {
...
-        return "\(message) - \(details)"
+        
+        let analytics = AnalyticsDestination()
+        logger.addDestination(analytics)
     }
 }
 
 /// Convenience wrapper around `MXLog` which formats all logs as "[Name] function: <message>"
+///
+/// Note: Ideallly the `format` of `ConsoleDestination` is set to track filename and function automatically
```
pixlwave, line 223:
```suggestion
/// Note: Ideally the `format` of `ConsoleDestination` is set to track filename and function automatically
```

MatrixSDK/Crypto/MXCrypto.m
pixlwave: Whereabouts does this crash actually happen? I'm wondering if there would be an opportune place we could map all of the `nil` values in the dictionary into `@"unknown"` before the crash takes place?

Otherwise it kind of feels like it would be an easy place for someone to introduce a crash by either not knowing this pitfall, or by not realising a value is nullable.

pixlwave (commented):
This looks like a great change overall to me 😎

MatrixSDK/Crypto/MXCrypto.m
Anderas: It is the actual construction of dictionary, for example this will crash:
```objc
NSString *value = nil;
NSDictionary *dict = @{
  @"value": value // If I passed nil here directly the compiler would catch it
};
```

So we can't really put any kind of safety wrapper around it. The rest of the codebase has this issue as well but is seems we rarely have unexpected nil values. My justification for allowing this was that most new code we write is in swift and thus we won't be really adding new dictionary contexts in objective-c, but of course I could be wrong

MatrixSDK/Data/MXRoomSummary.m
Anderas: Exactly, though it will be missing from rageshake files ... I think that is okay?

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
Anderas: Oops sorry, thanks

MatrixSDK/Utils/Logs/MXLog.h
Anderas: Missed that, thanks 🤦

MatrixSDK/Data/MXRoomSummary.m
pixlwave: Yep yep, makes sense to me 👍

MatrixSDK/Crypto/MXCrypto.m
pixlwave: TIL that this is standard objc behaviour! Ignore me 🤐

pixlwave (approved):
LGTM 👍

PR #1559, opened 2022-08-23
Allow setting room alias regardless of join rule

aringenbach:
This simply allows to set a room alias on room creation regardless of join rule.


gileluard (approved):
lgtm 👍

PR #1560, opened 2022-08-23
Fix incoming calls sometimes ringing after being answered on another client

aringenbach:
Potential fix for https://github.com/vector-im/element-ios/issues/6614

Allows the MXSession to be paused if app is already in background when the preventPauseCount reaches 0.

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1345,13 +1345,6 @@ typedef void (^MXOnResumeDone)(void);
     if (self.preventPauseCount > 0)
     {
         self.preventPauseCount--;
-#if TARGET_OS_IPHONE
-        if (self.preventPauseCount == 0
-            && self.applicationStateService.applicationState == UIApplicationStateBackground)
-        {
-            [self pause];
-        }
-#endif
     }
 }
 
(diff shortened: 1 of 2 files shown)


MatrixSDK/MXSession.m
```diff
@@ -1338,6 +1345,13 @@ - (void)releasePreventPause
     if (self.preventPauseCount > 0)
     {
         self.preventPauseCount--;
+#if TARGET_OS_IPHONE
+        if (self.preventPauseCount == 0
+            && self.applicationStateService.applicationState == UIApplicationStateBackground)
```
ismailgulek, line 1350: Instead of checking equal to background, can you check against not equal to active?

MatrixSDK/MXSession.m
```diff
@@ -1338,6 +1345,13 @@ - (void)releasePreventPause
     if (self.preventPauseCount > 0)
     {
         self.preventPauseCount--;
+#if TARGET_OS_IPHONE
```
ismailgulek, line 1348: Can you move this into the setter: `setPreventPauseCount`?

ismailgulek (changes requested):

ismailgulek (approved):
LGTM!

PR #1561, opened 2022-08-23
App Layout: force clear cache

gileluard:
part of #6616


stefanceriu (approved):
I didn't know this is the way we do it, makes sense 👍

PR #1563, opened 2022-08-25
Remove event from tracking

Anderas:
Remove encrypted event from being tracked to analytics via Sentry, only track the eventId.


stefanceriu (approved):
lgtm 👍

PR #1565, opened 2022-08-25
Self-verification flow in Crypto V2

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6589

Implement self-verification flow using rust-based OlmMachine. Changes include:

- Split up CryptoMachine verification methods into further, more separated protocols
- handle verification to-device events returned by CryptoMachine
- pass verification handler to requests and transactions

### DIFF
--- a/MatrixSDK/Crypto/Verification/MXKeyVerificationManager.h
+++ b/MatrixSDK/Crypto/Verification/MXKeyVerificationManager.h
@@ -86,12 +86,6 @@ FOUNDATION_EXPORT NSString *const MXKeyVerificationManagerNotificationTransactio
 #pragma mark - Requests
 
 /**
- The timeout for requests.
- Default is 5 min.
- */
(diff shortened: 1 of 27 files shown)


MatrixSDK/Crypto/Verification/MXKeyVerificationManager.h
```diff
@@ -85,12 +85,6 @@ FOUNDATION_EXPORT NSString *const MXKeyVerificationManagerNotificationTransactio
 
 #pragma mark - Requests
 
-/**
- The timeout for requests.
- Default is 5 min.
- */
-@property (nonatomic) NSTimeInterval requestTimeout;
```
Anderas, line 92: `requestTimeout` and `keyVerificationIdFromDMEvent` is not in fact called from outside of the manager, so can be hidden and not overriden by manager v2

PR #1566, opened 2022-08-26
Cache inbound group sessions when decrypting

Anderas:
Improve megolm event decryption behaviour by caching rather than re-fetching inbounds group sessions from the store. In anecdotal comparison on a single device this leads to 5x - 10x performance improvement.

In order to observe the impact of this improvement more precisely, the following changes are made:
- add build flag that randomly selects between new cache behaviour and old fetch behaviour, which will result in a very basic form of A/B test
- add duration tracking to analytics to measure the difference between both (this is performed by Sentry in element-ios)
- add duration tracking for MXSession.initialSync which calls megolm decrypt methods

Note that the new performance tracking method conflicts with existing `MXProfiler` which does the time tracking manually, whereas Sentry handles performance measurements internally. This ambiguity will be resolved in a future PR

Once in Sentry, the following dashboards can be constructed from these events:

Screenshot 2022-08-26 at 16 33 33

### DIFF
--- a/MatrixSDK/Crypto/MXOlmDevice.m
+++ b/MatrixSDK/Crypto/MXOlmDevice.m
@@ -29,6 +29,8 @@
 #import "MXKeyProvider.h"
 #import "MXRawDataKey.h"
 
+NSInteger const kMXInboundGroupSessionCacheSize = 100;
+
 @interface MXOlmDevice () <OLMKitPickleKeyDelegate>
 {
     // The OLMKit utility instance.
@@ -53,8 +55,10 @@
 // The store where crypto data is saved.
 @property (nonatomic, readonly) id<MXCryptoStore> store;
 
-@end
+// Cache to avoid refetching unchanged sessions from the crypto store
+@property (nonatomic, strong) MXLRUCache *inboundGroupSessionCache;
 
+@end
 
 @implementation MXOlmDevice
 @synthesize store;
@@ -98,6 +102,8 @@
 
         _deviceCurve25519Key = olmAccount.identityKeys[@"curve25519"];
         _deviceEd25519Key = olmAccount.identityKeys[@"ed25519"];
+        
+        _inboundGroupSessionCache = [[MXLRUCache alloc] initWithCapacity:kMXInboundGroupSessionCacheSize];
     }
     return self;
 }
@@ -394,7 +400,7 @@
         session.sharedHistory = sharedHistory;
     }
 
-    [store storeInboundGroupSessions:@[session]];
+    [self storeInboundGroupSessions:@[session]];
 
     return YES;
 }
@@ -441,7 +447,7 @@
         [sessions addObject:session];
     }
 
-    [store storeInboundGroupSessions:sessions];
+    [self storeInboundGroupSessions:sessions];
 
     return sessions;
 }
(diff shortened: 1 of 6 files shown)


MatrixSDK/MXSDKOptions.h
```diff
@@ -214,6 +214,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic) BOOL enableCryptoV2;
 
+/**
+ Enable performance optimization where inbound group sessions are cached between decryption of events
+ rather than fetched from the store every time.
+ 
+ @remark The value is set randomly between YES / NO to perform a very basic A/B test
```
manuroe, line 221:
```suggestion
 @remark By default, the value is set randomly between YES / NO to perform a very basic A/B test
```

MatrixSDK/Utils/MXAnalyticsDelegate.h
```diff
@@ -20,6 +20,12 @@
 #import "MXCallHangupEventContent.h"
 #import "MXTaskProfile.h"
 
+/**
+ Callback function to cancel ongoing duration tracking
```
manuroe, line 24:
```suggestion
 Callback function to finish ongoing duration tracking
```

MatrixSDK/Utils/MXAnalyticsDelegate.h
```diff
@@ -46,6 +52,21 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)trackDuration:(NSInteger)milliseconds name:(MXTaskProfileName)name units:(NSUInteger)units;
 
+/**
+ Start tracking the duration of a task and manually cancel when finished using the return handle
```
manuroe, line 56:
```suggestion
 Start tracking the duration of a task and manually stop when finished using the return handle
```

manuroe (commented):
This A/B performance testing is really great!

I was not comfortable with the word `cancel` at first. I thought it was for abortion. This is why I asked for some renaming.

MatrixSDK/Crypto/MXOlmDevice.m
```diff
@@ -521,6 +527,41 @@ - (MXDecryptionResult *)decryptGroupMessage:(NSString *)body
...
+    if (enableCache)
+    {
+        __block MXOlmInboundGroupSession *session;
+        @synchronized (self.inboundGroupSessionCache)
+        {
+            session = (MXOlmInboundGroupSession *)[self.inboundGroupSessionCache get:sessionId];
+            if (!session)
+            {
+                session = [store inboundGroupSessionWithId:sessionId andSenderKey:senderKey];
+                [self.inboundGroupSessionCache put:sessionId object:session];
```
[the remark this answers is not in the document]
Anderas: Oh I thought `session_id` was globally unique, as per [spec](https://spec.matrix.org/v1.3/client-server-api/#mmegolmv1aes-sha2) (see Rationale section) and `sender_key` deprecated. Is that not the case on iOS?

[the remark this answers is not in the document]
manuroe: oh. Thanks for the information. I was not aware about this spec update.

manuroe (approved):

PR #1567, opened 2022-08-30
Add build flag for symmetric backup

Anderas:
Add build flag for symmetric backup, when dynamically constructing backup algoritm class, based on incomming algorithm.

Unrelated: remove event from event tracking, see https://github.com/matrix-org/matrix-ios-sdk/pull/1563


aringenbach (approved):
LGTM

PR #1571, opened 2022-09-05
Remove partialTextMessage support

aringenbach:
Removes the support for MXRoom's partialTextMessage as it has been deprecated for a few months now and since it's usage was the root cause of issue https://github.com/vector-im/element-ios/issues/6670

Part of https://github.com/vector-im/element-ios/pull/6671

### DIFF
--- a/MatrixSDK/Background/MXBackgroundStore.swift
+++ b/MatrixSDK/Background/MXBackgroundStore.swift
@@ -201,15 +201,6 @@ class MXBackgroundStore: NSObject, MXStore {
     func deleteGroup(_ groupId: String) {
     }
 
-    @available(*, deprecated, message: "use storePartialAttributedTextMessage")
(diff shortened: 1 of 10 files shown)


gileluard:
This is a breaking change @ismailgulek or @manuroe shouldn't we just keep it?

aringenbach:
We can safely put this change on hold, as it is not really required for the fix on mention pills

manuroe:
> This is a breaking change @ismailgulek or @manuroe shouldn't we just keep it?

Thanks for asking. It is fine by me to remove this method that was marked as deprecated a long time ago. Plus, having a single method to store this data would avoid bugs in the app like the one linked by @aringenbach.

gileluard (approved):
I think it's the right time to apply this change. If so feel free to merge it.

PR #1575, opened 2022-09-16
Enable group session cache by default

Anderas:
After running a store / cache megolm decryption A/B test for a few weeks we gathered that cache decryption is aprox 10x faster and no new issues have been reported. As a result we can switch to cache variant exclusively, but leave the build flag in for a bit longer as a safety measure.

Screenshot 2022-09-16 at 09 36 58


gileluard (approved):
lgtm

PR #1576, opened 2022-09-19
Include concrete device model in user agent (PSG-754)

Johennes:
Fixes vector-im/element-ios#6742

### Pull Request Checklist

### DIFF
--- a/MatrixSDK/Utils/MXHTTPClient.m
+++ b/MatrixSDK/Utils/MXHTTPClient.m
@@ -26,6 +26,8 @@
 #import "MXHTTPClient_Private.h"
 #import "MXCredentials.h"
 
+#import "MatrixSDKSwiftHeader.h"
+
 #import <AFNetworking/AFNetworking.h>
 
 #pragma mark - Constants definitions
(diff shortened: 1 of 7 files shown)


MatrixSDK/Utils/MXHTTPClient.m
```diff
@@ -649,6 +651,27 @@ - (void)setRequestParametersInJSON:(BOOL)requestParametersInJSON
     {
         httpManager.requestSerializer = [AFHTTPRequestSerializer serializer];
     }
+
+    // Overwrite user agent to include concrete device model (code copied from AlamoFire)
```
Johennes, line 655: The code below is copied from `AFHTTPRequestSerializer`. I decided to keep this in the ObjC part because this way it's an exact copy and possibly easier to maintain than a Swift port.

Podfile.lock
```diff
@@ -14,6 +14,7 @@ PODS:
   - AFNetworking/Serialization (4.0.1)
   - AFNetworking/UIKit (4.0.1):
     - AFNetworking/NSURLSession
+  - DeviceKit (4.7.0)
```
Johennes, line 17: This is the best option I found for converting Apple's device identifiers into human readable strings. It's MIT, relatively small, well-maintained and somewhat popular so I hope this is an ok choice.

Johennes:
Rolling the reviewer assignment dice again because I think Andy is busy with the releases.

ismailgulek (commented):
Changes LGTM but I'd like to check with @manuroe that should this change be at the sdk or at the app side (can be done via `MXSDKOptions.additionalHTTPHeaders`)?

manuroe:
> Changes LGTM but I'd like to check with @manuroe that should this change be at the sdk or at the app side (can be done via `MXSDKOptions.additionalHTTPHeaders`)?

Good point, if possible, we would prefer this customisation only at the app level for apps that require it.

Johennes:
Sounds good. Closing this in favor of https://github.com/vector-im/element-ios/pull/6761.

PR #1578, opened 2022-09-22
Extract key backup engine

Anderas:
As part of implementing key backups with Rust-based CryptoSDK I need to separate out shared and distinct backup functionality in the existing codebase. In the past I have simply duplicated a class (`MXCrypto`, `MXKeyVerificationManager`) and overriden all of its methods, but this approach is not suitable for backups, because a lot more code ends up being shared (state management, enabling/disabling backups, restoring, sending http requests etc).

The approach I have chosen instead is to identify all of the code specific to native / legacy implementation and refactor it as new `MXKeyBackupEngine` protocol. At the moment only `Native` implementation exists, but in a future PR I will add the rust-based variant. It is absent from this PR to make the refactor cleaner and easier to review.

As for the refactor I have copy-pasted entire chunks of code with no or very minimal changes to avoid any regressions. There are some integration tests that cover this functionality, but given the complexity of this code it is best to be extra careful. In reviewing I recommend comparing the deleted and added chunks to confirm there are no behaviour changes.


Anderas:
@aringenbach it seems that codecov has made reviewing this PR quite difficult, in that it annotates almost every line of the refactor. I can disable codecov for the purpose of the review, unless you have some other flow of reviewing (e.g. from local command line)

aringenbach (approved):
@Anderas No worries I reviewed locally instead. 

Code LGTM 👍

PR #1581, opened 2022-09-26
Fix missing audio on consulting call leg during call transfer

### LINKED ISSUE #1580: Consulting call in call transfer has no audio
In a setup with a sip bridge, there is no audio on the consulting call leg within a call transfer

* The element iOS client is active in foreground
* A call from a PSTN number (0797710405) via sip bridge is coming in and is accepted
* This call has good audio
* The user now presses the transfer button, putting the PSTN user on hold (hears hold music)
* The user dials a second PSTN number on the transfer dial pad and connects with “consult first” active
* The second PSTN user gets and accepts the call
* This consulting call has no audio at all

simonwiedmer:
Fixes #1580 

Signed-off by simon.wiedmer@aarenet.com


benparsons:
Raising the `maximumCallGroups` value in CallKit is required when transferring calls. When set to `1` there is no audio on the transferred call. Dev raised the value to `3`, which is also the Apple default.

manuroe:
Hi @simonwiedmer, thanks for your bug fix and the research that lead to it. I close this PR in favor of #1738 which is more generic. Please use the new setting `MXCallKitConfiguration.maximumCallGroups` in your application.

PR #1584, opened 2022-09-27
Simplify Key Backup Engine API

Anderas:
After creating [key backup engine](https://github.com/matrix-org/matrix-ios-sdk/pull/1578) in preparation for rust-based key backups, this PR makes another set of API changes to the engine to separate out refactors (this PR) from rust backups (future PR).

The main changes in this PR include:
- Simplify backup and import methods of the engine
- Explicit dependencies for recovery service rather than accessing them via `self.crypto...`
- Extract secret storage out of `MXCryptoStore` into separate protocol to be usable with Crypto V2


gileluard (approved):
lgtm

PR #1585, opened 2022-09-27
Key backups with Crypto V2

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6768 and https://github.com/vector-im/element-ios/issues/6769

This PR adds key backup functionality using rust-based crypto machine. The changes include:

- Implement backup-related functionality from rust-based `OlmMachine`, hidden behind `MXCryptoBackup` protocol
- Implement new backup engine, similar to the one introduced [previously](https://github.com/matrix-org/matrix-ios-sdk/pull/1578)
- Initialize backup and cross signing upon first keys upload via crypto machine
- Allow discarding room keys via `/discardsession` slash command

### DIFF
--- a/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
+++ b/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
@@ -43,7 +43,7 @@ class MXCryptoMachine {
     enum Error: Swift.Error {
         case invalidStorage
         case invalidEvent
-        case nothingToEncrypt
+        case cannotSerialize
         case missingRoom
         case missingVerificationContent
         case missingVerificationRequest
@@ -60,6 +60,9 @@ class MXCryptoMachine {
     private let syncQueue = MXTaskQueue()
     private var roomQueues = RoomQueues()
     
+    private var hasUploadedKeys = false
+    private var keysUploadCallback: (() -> Void)?
+    
     private let log = MXNamedLog(name: "MXCryptoMachine")
 
     init(userId: String, deviceId: String, restClient: MXRestClient, getRoomAction: @escaping GetRoomAction) throws {
@@ -74,6 +77,16 @@ class MXCryptoMachine {
         self.getRoomAction = getRoomAction
         
         setLogger(logger: self)
+    }
+    
+    func onKeysUpload(callback: @escaping () -> Void) {
+        log.debug("->")
+        
+        if hasUploadedKeys {
+            callback()
+        } else {
+            keysUploadCallback = callback
+        }
     }
     
     private static func storeURL(for userId: String) throws -> URL {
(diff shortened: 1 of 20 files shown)


MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -76,6 +79,16 @@ class MXCryptoMachine {
         setLogger(logger: self)
     }
     
+    func onKeysUpload(callback: @escaping () -> Void) {
+        log.debug("->")
```
stefanceriu, line 83: Interesting looking log 🙃

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -76,6 +79,16 @@ class MXCryptoMachine {
         setLogger(logger: self)
     }
     
+    func onKeysUpload(callback: @escaping () -> Void) {
```
stefanceriu, line 82: Should we name this something more descriptive like `setKeyUploadObserver`?

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -510,6 +546,85 @@ extension MXCryptoMachine: MXCryptoSASVerifying {
...
+extension MXCryptoMachine: MXCryptoBackup {
+    var isBackupEnabled: Bool {
+        return machine.backupEnabled()
+    }
+    
+    var backupKeys: BackupKeys? {
+        do {
+            return try machine.getBackupKeys()
+        } catch {
+            log.error("Failed fetching backup keys", context: error)
```
stefanceriu, line 558: Does this need to be handled better, throw or maybe just crash?

stefanceriu (approved):
I'm obviously not particularl familiar with any of these but apart from some error handling issues, which presumably come from the original implementation, it looks really good to me 👍

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
Anderas: I don't know what legitimate reasons for this to fail are, given that it still returns an optional if we don't have any backup. It is called by too many places to surface the error upward so logs / sentry will have to suffice

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -510,6 +546,85 @@ extension MXCryptoMachine: MXCryptoSASVerifying {
...
+            return nil
+        }
+    }
+    
+    var roomKeyCounts: RoomKeyCounts? {
+        do {
+            return try machine.roomKeyCounts()
+        } catch {
+            log.error("Cannot get room key counts", context: error)
```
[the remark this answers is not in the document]
Anderas: Yea same thing here. Perhaps this throws if the machine is not sufficiently setup or so, but have not managed to reproduce. This is why the domain API on top of it does not expose the error but only logs it.

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
Anderas: Hm, not sure if thats better, because with the observer pattern you typically also need to explictly call `unset` or `remove`, whereas here it is unset internally on first upload. So perhaps I can rename it to `onInitialKeysUpload`

PR #1586, opened 2022-09-28
Key gossiping with Crypto V2

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6773 and https://github.com/vector-im/element-ios/issues/6774

Expose specific decryptionError when room keys are missing (which is displayed as "re-request keys" message on iOS), and implement `requestRoomKey` which will send out `m.room_key_request` message to other verified devices of the same user. Once recieved by the other device, its OlmMachine will automatically send out `m.forwarded_room_key` if necessary.


ismailgulek (approved):
Ah forgot to approve.

PR #1588, opened 2022-09-29
Add `enableNewClientInformationFeature` sdk option (PSG-799)

ismailgulek:


MatrixSDK/ClientInformation/MXClientInformationService.swift
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@objcMembers
+public class MXClientInformationService: NSObject {
+
+    private weak var session: MXSession?
```
paleksandrs, line 22: Any reason `session` is `weak`?

MatrixSDK/ClientInformation/MXClientInformationService.swift
```diff
@@ -0,0 +1,90 @@
...
+        guard let session = session else {
+            return
+        }
+
+        guard let bundleDisplayName = Bundle.main.object(forInfoDictionaryKey: "CFBundleDisplayName") ?? Bundle.main.object(forInfoDictionaryKey: "CFBundleName") as? String,
+              let appVersion = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String else {
+            return
+        }
+
+        let updatedInfo: [AnyHashable: String] = [
```
paleksandrs, line 42: A small suggestion. It would be nice to break down `refresh` method to make it more readable because it contains multiple `guard` statements. For example, the creation of `updatedInfo` could be extracted into a separate method `createAccountData`

MatrixSDK/ClientInformation/MXClientInformationService.swift
```diff
@@ -0,0 +1,90 @@
...
+@objcMembers
+public class MXClientInformationService: NSObject {
+
+    private weak var session: MXSession?
+
+    public init(withSession session: MXSession) {
+        self.session = session
+    }
+
+    public func refresh() {
```
paleksandrs, line 28: Considering that this method is also removing data, maybe some other naming would be more suitable,.e.g. `updateOrRemoveAccountData` ?

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,91 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import XCTest
+import MatrixSDK
+
+class MXClientInformationServiceUnitTests: XCTestCase {
```
paleksandrs, line 20: The name of the test class indicates that you are testing `MXClientInformationService`, but I can't seem to find that is created or tested anywhere?

MatrixSDK/ClientInformation/MXClientInformationService.swift
ismailgulek: To avoid retain cycles between session and the service because the session holds a strong reference to the service.

paleksandrs: Oh, right. The session holds the reference 👍

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: Yeah it was testing it implicitly, updated.

MatrixSDK/ClientInformation/MXClientInformationService.swift
ismailgulek: Yep good idea.

MatrixSDK/ClientInformation/MXClientInformationService.swift
ismailgulek: Updated the method name to `updateData`. Removing is a kind of update so it should be fine.

MatrixSDK/ClientInformation/MXClientInformationService.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        session.setAccountData(updatedInfo, forType: type) {
+            MXLog.debug("[MXClientInformationService] refresh: updated successfully")
+        } failure: { error in
+            MXLog.debug("[MXClientInformationService] refresh: update failed: \(String(describing: error))")
+        }
+    }
+
+    internal func removeDataIfNeeded(on session: MXSession) {
```
paleksandrs, line 56: Should it be `private` method? Same for other `internal` methods in this class

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,165 @@
...
+
+        // no client info before
+        let clientInfo = session.accountData.accountData(forEventType: type)
+        XCTAssertNil(clientInfo)
+
+        service.updateData()
+
+        // must be set after updateData
+        let updatedInfo1 = session.accountData.accountData(forEventType: type)
+        XCTAssertNotNil(updatedInfo1)
```
paleksandrs, line 45: Should the values of `updatedInfo1` be tested as well, not just not nil?

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,165 @@
...
+
+        // try updating again
+        service.updateData()
+
+        // must not be changed
+        let updatedInfo2 = session.accountData.accountData(forEventType: type)
+        XCTAssertTrue(NSDictionary(dictionary: updatedInfo1!).isEqual(to: updatedInfo2))
```
paleksandrs, line 52: Is this test not passing if you remove this [guard](https://github.com/matrix-org/matrix-ios-sdk/pull/1588/files#diff-0f39a17e4f2198d23ebf5a26c9ba8f659ea7d87a8936b0329c5957895dd919b5R59) in the implementation? If I am not mistaken, even if the call is made, you would be comparing equal objects. Maybe it would be useful to create a session mock and inject it into `MXClientInformationService` and test that `setAccountData` is not called on it? In that case, you wold be able also to test the values passed. 
From test method name it is hard to understand what this test is about. I would suggest splitting this test into multiple tests. Something like:
- test_whenUpdateData_accountDataIsUpdatedWithCorrectValues
- test_whenUpdateDataSecondTime_setAccountDataIsNotCalled

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,165 @@
...
+    func testRemoveData() {
+        let mockDeviceId = "some_device_id"
+        let credentials = MXCredentials(homeServer: "", userId: "@userid:example.com", accessToken: "")
+        credentials.deviceId = mockDeviceId
+        guard let session = MXSession(matrixRestClient: MXRestClientStub(credentials: credentials)) else {
+            XCTFail("Failed to setup test conditions")
+            return
+        }
+
+        let service = MXClientInformationService(withSession: session)
```
paleksandrs, line 66: Could this code block be extracted into a method to avoid repetition (e.g. createSession)?

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        let clientInfo = service.createClientInformation()
```
paleksandrs, line 138: It seems you are testing internal implementation method, that shouldn't be done. If you want to test values, please see the comment about testing values

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        let credentials = MXCredentials(homeServer: "", userId: "@userid:example.com", accessToken: "")
+        credentials.deviceId = mockDeviceId
+        guard let session = MXSession(matrixRestClient: MXRestClientStub(credentials: credentials)) else {
+            XCTFail("Failed to setup test conditions")
+            return
+        }
+
+        let service = MXClientInformationService(withSession: session)
+
+        XCTAssertEqual(service.accountDataType(for:session), "\(kMXAccountDataTypeClientInformation).\(mockDeviceId)")
```
paleksandrs, line 161: Same here

paleksandrs (commented):
Sorry, some more comments. Maybe something will be useful.

MatrixSDK/ClientInformation/MXClientInformationService.swift
ismailgulek: Should be `internal` to be accessible from tests.

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: It's a bit hard to test the values because `name` and `version` properties should come from the bundle. Now it's like:
`name: xctest iOS, version: 13.4`. Maybe name is standard but you need to read Xcode version to fully test it.

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: I have a thought about it but didn't like it afterwards because we're using `mockDeviceId` and we would also need to check optionals again, so it's still boring.

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: Updated.

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: I'm not quite sure I'm getting your concern here. Why an internal method shouldn't be tested?

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
paleksandrs: Sorry, what do you mean?

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
paleksandrs: When unit testing you shouldn't care about the internal implementation of the class your are testing. You are mainly interested to see what's the unit output based on the inputs you give. Pubic interface is the one that should be tested. That allows safely to refactor all the code in those classes without worrying that by changing internal implementation you may brake something. By writing enough tests that cover public interface, you are covering internal implementation as well. That is the beauty of the unit testing :)

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
paleksandrs: Ok. If you want, you can create a protocol something like `ClientInformationProvidable` and inject it into `MXClientInformationService`. In the real code, implementation of that providable would use real Bundle, but in the unit test target, you could create mock. It is just an idea if you would like to do it. Up to you

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: Nevermind, updated.

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: Updated

MatrixSDKTests/MXClientInformationServiceUnitTests.swift
ismailgulek: That was serving a purpose here but ok, i made them private again.

paleksandrs (approved):

PR #1589, opened 2022-09-29
Manual device verification with Crypto V2

Anderas:
Resolved https://github.com/vector-im/element-ios/issues/6781

- implement functionality to manually verify a device (and upload its signatures)
- reflect trust status when a device is blocked
- return cross signing keys as secrets when available


aringenbach (approved):
LGTM

PR #1590, opened 2022-10-03
Added support for MSC3881

gileluard:
needed for [#6787](https://github.com/vector-im/element-ios/issues/6787)

### DIFF
--- a/MatrixSDK/Contrib/Swift/JSONModels/MXJSONModels.swift
+++ b/MatrixSDK/Contrib/Swift/JSONModels/MXJSONModels.swift
@@ -56,16 +56,13 @@ public enum MXLoginFlowType: Equatable, Hashable {
 public enum MXPusherKind: Equatable, Hashable {
     case http, none, custom(String)
     
-    public static func from(value: String?) -> MXPusherKind {
+    public init(value: String?) {
         guard let value = value else {
-            return .none
-        }
-
-        if value == "http" {
-            return .http
+            self = .none
+            return
         }
         
-        return .custom(value)
+        self = value == "http" ? .http : .custom(value)
     }
     
     public var objectValue: NSObject {
--- a/MatrixSDK/Contrib/Swift/JSONModels/MXJSONModels.swift
+++ b/MatrixSDK/Contrib/Swift/JSONModels/MXJSONModels.swift
@@ -56,6 +56,18 @@ public enum MXLoginFlowType: Equatable, Hashable {
 public enum MXPusherKind: Equatable, Hashable {
     case http, none, custom(String)
     
+    public static func from(value: String?) -> MXPusherKind {
+        guard let value = value else {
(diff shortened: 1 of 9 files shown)


MatrixSDK/MXRestClient.m
```diff
@@ -3200,7 +3228,7 @@ - (MXHTTPOperation*)roomSummaryWith:(NSString*)roomIdOrAlias
                             success:(void (^)(MXPublicRoom *room))success
                             failure:(void (^)(NSError *error))failure
 {
-    NSMutableString *path = [NSMutableString stringWithFormat:@"%@/im.nheko.summary/rooms/%@/summary", kMXAPIPrefixPathUnstable, roomIdOrAlias];
+    NSMutableString *path = [NSMutableString stringWithFormat:@"%@/im.nheko.summary/rooms/%@/summary", kMXAPIPrefixPathUnstable, [MXTools encodeURIComponent:roomIdOrAlias]];
```
pixlwave, line 3231: Looks sensible, but not sure I understand how this change relates to the MSC.

MatrixSDK/JSONModels/MXMatrixVersions.m
```diff
@@ -42,6 +42,8 @@
 
 //  Unstable features
 static NSString* const kJSONKeyMSC3440 = @"org.matrix.msc3440.stable";
+static NSString* const kJSONKeyMSC3881 = @"org.matrix.msc3881";
+static NSString* const kJSONKeyMSC3881Stable = @"org.matrix.msc3881.stable";
```
pixlwave, line 46: Given `kJSONKeyMSC3440` is stable, would it make more sense to do this like so?
```suggestion
static NSString* const kJSONKeyMSC3881Unstable = @"org.matrix.msc3881";
static NSString* const kJSONKeyMSC3881 = @"org.matrix.msc3881.stable";
```

MatrixSDK/Contrib/Swift/JSONModels/MXJSONModels.swift
```diff
@@ -56,6 +56,18 @@ public enum MXLoginFlowType: Equatable, Hashable {
 public enum MXPusherKind: Equatable, Hashable {
     case http, none, custom(String)
     
+    public static func from(value: String?) -> MXPusherKind {
```
pixlwave, line 59: Unimportant, but wondering if this could be an `init`?

pixlwave (approved):
LGTM 😎

MatrixSDK/MXRestClient.m
gileluard: Indeed. I've just added this code to fix some eventual issues. Do you want me to revert it?

gileluard: This could definitely happen to me :D

PR #1596, opened 2022-10-04
Add support for m.local_notification_settings.<device-id> in account_…

gileluard:
needed for [#6797](https://github.com/vector-im/element-ios/issues/6797)

As per [MSC3890](https://github.com/matrix-org/matrix-spec-proposals/pull/3890) we'll need to catch the new `m.local_notification_settings.<device-id>` event in the account data to know if the session support local notification and if it's silenced.


ismailgulek (approved):
I had a look to understand https://github.com/vector-im/element-ios/pull/6798 better and lgtm!

gileluard:
@ismailgulek Yeah ... I didn't think this PR lacks of context....

stefanceriu (approved):
I'm happy if Ismail is happy but looks good to me too 👍 😄

PR #1597, opened 2022-10-05
Add extra logging and always cancel eventStreamRequest before setting to nil

aringenbach:
Extra logging for https://github.com/vector-im/element-ios/issues/6614 issue

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -1097,6 +1097,7 @@ typedef void (^MXOnResumeDone)(void);
         // Cancel the current request managing the event stream
         [eventStreamRequest cancel];
         eventStreamRequest = nil;
+        MXLogDebug(@"[MXSession] pause: did cancel and remove eventStreamRequest");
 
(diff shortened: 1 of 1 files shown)


MatrixSDK/MXSession.m
```diff
@@ -1672,6 +1680,9 @@ - (void)handleServerSyncError:(NSError*)error forRequestWithServerTimeout:(NSUIn
     // Check whether the caller wants to handle error himself
     if (failure)
     {
+        MXLogErrorDetails(@"[MXSession] handleServerSyncError: sync fail is handled by caller, resetting eventStreamRequest might not be handled", @{
+            @"error": error ?: @"unknown"
+        });
```
aringenbach, line 1685: @ismailgulek WDYT about this bit ? If the error is a timeout and failure != nil, the eventStreamRequest might not be reset 🤔

ismailgulek: Yeah you right. It even makes sense to reset the request here regardless of `failure` block.

ismailgulek (approved):
LGTM!

PR #1598, opened 2022-10-05
e2ee should not hinder verification

BillCarsonFr:
### Pull Request Checklist

Fixes [#6519](https://github.com/vector-im/element-ios/issues/6519) 
Verification events sent in e2ee rooms should bypass the encrypt to verified devices only option. If not it renders verification impossible.

## Motivation and context
As per spec:

> When using in-room messages and the room has encryption enabled, clients should ensure that encryption does not hinder the verification. For example, if the verification messages are encrypted, clients must ensure that all the recipient’s unverified devices receive the keys necessary to decrypt the messages, even if they would normally not be given the keys to decrypt messages in the room. Alternatively, verification messages may be sent unencrypted, though this is not encouraged.

### DIFF
--- a/MatrixSDK/Crypto/Algorithms/MXEncrypting.h
+++ b/MatrixSDK/Crypto/Algorithms/MXEncrypting.h
@@ -60,7 +60,7 @@
 
  @return a MXHTTPOperation instance. May be nil if all required materials is already in place.
  */
-- (MXHTTPOperation*)ensureSessionForUsers:(NSArray<NSString*>*)users
+- (MXHTTPOperation*)ensureSessionForUsers:(NSArray<NSString*>*)users forceDistributeToUnverified: (BOOL) forceDistributeToUnverified
                                   success:(void (^)(NSObject *sessionInfo))success
                                   failure:(void (^)(NSError *error))failure;
 
--- a/MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmEncryption.m
+++ b/MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmEncryption.m
@@ -93,7 +93,8 @@
     queuedEncryption.failure = failure;
     [pendingEncryptions addObject:queuedEncryption];
 
-    return [self ensureSessionForUsers:users success:^(NSObject *sessionInfo) {
+    BOOL forceDistributeToUnverified = [self isVerificationEvent:eventType eventContent:eventContent];
+    return [self ensureSessionForUsers:users forceDistributeToUnverified:forceDistributeToUnverified success:^(NSObject *sessionInfo) {
 
         MXOutboundSessionInfo *session = (MXOutboundSessionInfo*)sessionInfo;
         [self processPendingEncryptionsInSession:session withError:nil];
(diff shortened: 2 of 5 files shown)


MatrixSDK/Crypto/Algorithms/MXEncrypting.h
```diff
@@ -60,7 +60,7 @@
 
  @return a MXHTTPOperation instance. May be nil if all required materials is already in place.
  */
-- (MXHTTPOperation*)ensureSessionForUsers:(NSArray<NSString*>*)users
+- (MXHTTPOperation*)ensureSessionForUsers:(NSArray<NSString*>*)users forceDistributeToUnverified: (BOOL) forceDistributeToUnverified
```
Anderas, line 63: Sadly we don't have obj-c style checker / formatter to check this automatically, but to aligh the style:
- no spaces between type and parameter, i.e. `forceDistributeToUnverified: (BOOL) forceDistributeToUnverified` => `forceDistributeToUnverified:(BOOL)forceDistributeToUnverified` (across the whole PR)
- if some parameters are on multiple lines (e.g. `success` and `failure`), then all should be, incl `forceDistributeToUnverified`, and aligned vertically by the double colon

MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmEncryption.m
```diff
@@ -166,6 +190,7 @@ - (BOOL)isSessionSharingHistory:(MXOutboundSessionInfo *)session
  @param failure A block object called when the operation fails.
 */
 - (MXHTTPOperation *)getDevicesInRoom:(NSArray<NSString*>*)users
```
Anderas, line 192: This method includes documentation above so would be good to include the new parameter. Also perhaps just a matter of personal preference, but the method `getDevicesInRoom` does not suggest anything about distributing keys so perhaps the new parameter would be clearer as `includeUnverifiedUsers` rather than `forceDistributeToUnverified`

MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmEncryption.m
```diff
@@ -198,7 +223,7 @@ - (MXHTTPOperation *)getDevicesInRoom:(NSArray<NSString*>*)users
                 }
 
                 if (deviceInfo.trustLevel.localVerificationStatus == MXDeviceBlocked
-                    || (!deviceInfo.trustLevel.isVerified && encryptToVerifiedDevicesOnly))
+                    || (!deviceInfo.trustLevel.isVerified && encryptToVerifiedDevicesOnly && !forceDistributeToUnverified))
```
Anderas, line 226: I think it would be better / cleaner to include this extra condition in the `BOOL encryptToVerifiedDevicesOnly = ` declaration.

MatrixSDK/Crypto/Algorithms/Megolm/MXMegolmEncryption.m
```diff
@@ -103,14 +104,37 @@ - (MXHTTPOperation*)encryptEventContent:(NSDictionary*)eventContent eventType:(M
     }];
 }
 
-- (MXHTTPOperation*)ensureSessionForUsers:(NSArray<NSString*>*)users
+- (BOOL) isVerificationEvent:(MXEventTypeString) eventType eventContent:(NSDictionary*)eventContent
```
Anderas, line 107: Optionally this whole method could be moved into `MXTools`, as `[MXTools isVerificationEvent]`, which will make it easier to unit test

PR #1599, opened 2022-10-05
Refactor verification manager, requests, transactions

Anderas:
Turn `MXKeyVerificationManager` into a protocol and rename all existing objective-c implementations to `MXLegacy...` variants. This is to make Crypto V2 verification api safer.


gileluard (approved):
lgtm

PR #1600, opened 2022-10-05
Incoming verification requests with Crypto V2

Anderas:
Implement incoming and direct-message verification requests which includes:
- creating direct room with user if one does not exist yet
- showing verification UI in the room timeline and synchronizing with the state of the verification
- handle to-device and dm events to update the UI via notifications (This functionality should eventually be moved to CryptoSDK to simplify client-side implementation)
- add state resolver that displays the outcome of verification once CryptoMachine no longer has any active requests/transactions


MatrixSDK/Crypto/Algorithms/Olm/MXOlmDecryption.m
```diff
@@ -228,6 +228,11 @@ - (void)onRoomKeyInfo:(MXRoomKeyInfo *)keyInfo
     // No impact for olm
 }
 
+- (void)onRoomKey:(MXRoomKeyResult *)key
```
Anderas, line 231: Unrelated change, missing implementation required by a protocol

ismailgulek (approved):
LGTM!

PR #1601, opened 2022-10-07
Expose rest client method for generating login tokens through MSC3882

stefanceriu:


MatrixSDK/MXRestClient.m
```diff
@@ -906,6 +906,30 @@ - (NSString*)loginFallback;
     return loginFallback;
 }
 
+- (MXHTTPOperation*)generateLoginTokenWithSuccess:(void (^)(MXLoginToken *))success
+                                          failure:(void (^)(NSError *error))failure
+{
+    MXWeakify(self);
+    return [httpClient requestWithMethod:@"POST"
+                                    path:[NSString stringWithFormat:@"%@/login/token", apiPathPrefix]
```
ismailgulek, line 914: The path is stable?

ismailgulek (approved):
LGTM with a note.

MatrixSDK/MXRestClient.m
stefanceriu: Ah, no it's not 😞 Good catch 👍

stefanceriu:
This should be correct now although our current test synapse instance doesn't support it

ismailgulek (approved):

PR #1602, opened 2022-10-11
Refactor QR transactions

Anderas:
[Another](https://github.com/matrix-org/matrix-ios-sdk/pull/1599) refactor PR that replaces a concrete implementation of QR verification transaction with a protocol so that two implementations can be used interchangeably


pixlwave (approved):
LGTM 👍

PR #1604, opened 2022-10-11
QR code verification with Crypto V2

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/6859


stefanceriu (approved):
lgtm! 👍

PR #1605, opened 2022-10-13
Unencrypted verification events in Crypto V2

Anderas:
The last few changes (for a while) in verification for Crypto V2:

- process unencrypted verification events manually (they obviously cannot be handled automatically when calling `decryptRoomEvent`)
- add support for SAS decimals
- restore pending verification requests after app relaunch


stefanceriu (approved):
lgtm 👍

PR #1606, opened 2022-10-13
Remove megolm decrypt cache build flag

Anderas:
After a brief stabilization period with the cache variant set to YES, we can now remove the flag altogether


gileluard (approved):
lgtm 👍

Issue #1609, opened 2022-10-18
Device Manager: Exposed method to update client information


paleksandrs:
Part of https://github.com/vector-im/element-ios/issues/6904

PR #1610, opened 2022-10-18
Device Manager: Exposed method to update client information

paleksandrs:
closes https://github.com/matrix-org/matrix-ios-sdk/issues/1609


ismailgulek (approved):
LGTM!

PR #1613, opened 2022-10-19
Set local trust with Crypto V2 and deprecate legacy verification method

Anderas:
A few final bits to implement CryptoV2:

- set local trust (incl (un)blacklisting users)
- remove deprecated and legacy verification method that does not begin with verification request (use `requestVerificationWithDevice` instead)
- determine room encryption state
- retry decryption of events after manually importing keys


MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
```diff
@@ -399,6 +369,25 @@ class MXKeyVerificationManagerV2: NSObject, MXKeyVerificationManager {
     
     // MARK: - Verification requests
     
+    func requestVerificationByToDevice(
+        withUserId userId: String,
```
stefanceriu, line 373: Is this `userId` supposed to be used below instead of the session one? 🤔

stefanceriu (approved):
Apart from one small comment lgtm 👍

MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
Anderas: Not sure I understand the question. The method serves to verify any user, incl self, but the ffi  interface has explicitely `requestSelfVerification` vs `requestVerification(user:...)`. So if the passed-in userId matches the session, we use self-verification.

stefanceriu: Ups sorry, my bad. I completely missread that being as an `if **let** userId == session?.myUserId`, please disregard.

PR #1614, opened 2022-10-20
Define MXCrypto and MXCrossSigning as protocols

Anderas:
Convert `MXCrypto` and `MXCrossSigning` into protocols and define `..Legacy..` implementations for both. These two types are the last two that were previously overriden by rust-based V2 implementations, but can now be relatively safely converted into protocols.

To make the conversion possible a number of changes were made:
- Expose protocol types instead of implementation where possible
- Legacy crypto functionality (MXLegacyKeyVerificationManager, MXKeyVerificationRequest, MXDecrypting ...) need to access the legacy implementation directly, as they depend on symbols defined in`MXCrypto_Private`.
- A number of integration tests depend on private legacy symbols, not compatible with V2. This means those tests cannot be currently run against V2 implementation and will need to be fixed in the future.

Note that this PR does minimum changes to the actual protocols to ease the review of this PR. Future PRs will further tweak the protocol requirements and add nullability support for `MXCrypto`


MatrixSDK/Crypto/CrossSigning/MXCrossSigning.h
```diff
@@ -179,6 +174,20 @@ typedef NS_ENUM(NSInteger, MXCrossSigningErrorCode)
                 onPrivateKeysReceived:(void (^)(void))onPrivateKeysReceived
                               failure:(void (^)(NSError *error))failure;
 
+/**
+ Does given secret, interpreted as a private key, match given public cross-signing key
```
pixlwave, line 178:
```suggestion
 Does the given secret, interpreted as a private key, match given public cross-signing key
```

MatrixSDKTests/Crypto/MXLegacyCrypto+LegacyCrossSigning.swift
```diff
@@ -0,0 +1,36 @@
...
+import Foundation
+
+@objc extension MXLegacyCrypto {
+    
+    // MXLegacyCrossSigning exposes a number of internal methods and properties in `MXCrossSigning_Private`
+    // which are used in integration tests to perform actions and assert outcomes.
+    // These methods and properties will not be available in MXCrossSigningV2 and therefore given
+    // integration tests cannot be run (or have to be re-written).
+    var legacyCrossSigning: MXLegacyCrossSigning? {
+        guard let crossSignign = crossSigning else {
```
pixlwave, line 26:
```suggestion
        guard let crossSigning = crossSigning else {
```

pixlwave (approved):
Looks good to me. The only question I have is whether you might want to rename some of the files to `…Legacy…` to match updated class names?

Anderas:
> Looks good to me. The only question I have is whether you might want to rename some of the files to `…Legacy…` to match updated class names?

I do plan to align all the naming (and potentially replace V2 with something more sensible), but I think adding `Legacy` only makes sense to those classes that have at least one implementation (eg. Crypto, KeyVerificationManager). For others (eg MegolmDecryption, OlmEncryption etc) this is unnecessary: there is currently no confusion and once fully migrated these filed will be simply removed.

PR #1616, opened 2022-10-24
Cross-sign self after restoring session with Crypto V2

Anderas:
Implement several missing methods to fully cross-sign current user and device after restoring session with security key.


stefanceriu (approved):
lgtm 👍

PR #1618, opened 2022-10-25
Curate MXCrypto protocol methods

Anderas:
A number of related changes to the `MXCrypto` protocol:

- cleanup `MXCrypto` protocol by moving a number of methods and properties into `MXLegacyCrypto`, because they are used only with the legacy implementation of the crypto module, and thus do not need to be implemented by `MXCryptoV2`
- consolidate similar methods `trustLevelSummary` into one
- deprecate legacy synchronous `decryptEvent` method
- track duration of encrypting an event with Crypto V2

Note that in the future PR the remaining unimplemented methods in CryptoV2 will be dealt with

### DIFF
--- a/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
+++ b/MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
@@ -314,6 +314,11 @@ NSString *const MXCrossSigningErrorDomain = @"org.matrix.sdk.crosssigning";
     });
 }
 
+- (MXCrossSigningInfo *)crossSigningKeysForUser:(NSString *)userId
+{
+    return [self.crypto.store crossSigningKeysForUser:userId];
+}
+
 - (void)requestPrivateKeysToDeviceIds:(nullable NSArray<NSString*>*)deviceIds
                               success:(void (^)(void))success
                 onPrivateKeysReceived:(void (^)(void))onPrivateKeysReceived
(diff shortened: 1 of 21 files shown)


MatrixSDK/Crypto/CrossSigning/MXCrossSigning.m
```diff
@@ -314,6 +314,11 @@ - (void)signUserWithUserId:(NSString*)userId
     });
 }
 
+- (MXCrossSigningInfo *)crossSigningKeysForUser:(NSString *)userId
+{
+    return [self.crypto.store crossSigningKeysForUser:userId];
```
Anderas, line 319: Implementation merely moved here from `MXCrypto`

MatrixSDK/Crypto/MXCrypto.h
```diff
@@ -102,48 +102,11 @@ extern NSString *const MXDeviceListDidUpdateUsersDevicesNotification;
  */
 @property (nonatomic, readonly) MXRecoveryService *recoveryService;
 
-/**
- The secret storage on homeserver manager.
- */
-@property (nonatomic, readonly) MXSecretStorage *secretStorage;
```
Anderas, line 108: All of these (apart from deprecated `decryptEvent` and duplciate `trustLevelSummary`) are just moved to `MXLegacyCrypto` without change

MatrixSDK/Utils/MXTaskQueue.swift
```diff
@@ -48,6 +48,7 @@ public actor MXTaskQueue {
             assertionFailure("Failing to get value of the correct type should not be possible")
             throw Error.valueUnavailable
         }
+        previousTask = nil
```
Anderas, line 51: Unrelated and untestable change that clears a `Task` after it has been completed. Without this change everythign works as normal, but we keep a supended Task in memory unnecessarily

gileluard (approved):
lgtm

PR #1620, opened 2022-10-26
Complete MXCryptoV2 implementation

Anderas:
A large final round of changes to the `MXCryptoV2` implementation to make it ready for alpha testing. The changes in this PR include:

- Extract additional methods out of `MXCrypto` protocol and into `MXLegacyCrypto` implementation, as they are not needed with V2 (the functionality is handled internally by `CryptoMachine`)
- Annotate all of `MXCrypto` with nullability to make swift code safer and more explicit. At times this means making more things nullable than necessary just to handle any rogue objective-c `nil`s
- Update `MatrixSDKCrypto` to the latest version which includes `EncryptionSetting`
- Completely reorganize `MXCryptoV2` to group related functionality (apologies for impact on code review)

PR #1622, opened 2022-10-31
CryptoV2 changes

Anderas:
A number of changes to [rust-based Crypto V2](https://github.com/matrix-org/matrix-ios-sdk/pull/1620) after extensive testing with polyjuice:

- expose `VerificationState` on decryption result to display trusted / untrusted sessions
- ensure CryptoV2 gets initialized on the main queue
- move event observers from key verification manager to crypto v2 to gain access to downloading user keys and processing outgoing requests
- always download keys for new senders before processing to-device events
- restore room keys from backup after recieving recovery key

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -66,7 +66,7 @@ Pod::Spec.new do |s|
   
   # Experimental / NOT production-ready Rust-based crypto library
   s.subspec 'CryptoSDK' do |ss|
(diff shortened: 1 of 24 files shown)


paleksandrs (approved):
Looks good to me. Do you need to add a changelog for this?

Anderas:
> Looks good to me. Do you need to add a changelog for this?

Its pretty much bugfixes for https://github.com/matrix-org/matrix-ios-sdk/pull/1620, I just did not want to do it in the same PR for readability, so I don't think changelog is necessary

PR #1626, opened 2022-11-03
Doc: Update the synapse installation section with poetry usage

manuroe:
The doc using virtual env was obsolete.


Anderas (approved):
Is it not better to simply point to Synapse guide on how to install / setup? Otherwise we will always be getting out of sync with it

PR #1627, opened 2022-11-03
Room event decryption in Crypto V2

Anderas:
Refactors are the spice of life ...

In this PR I am doing a number of changes to make room event decryption more reliable after recieving room key event or import room keys. These changes include:

- Create `MXRoomEventDecryption` which contains all of the logic of decypting keys, logging errors (and supress duplicate warnings for a known missing key), and re-decrypt after recieving a room key or import. This means that a bunch of code can be moved out of `MXCryptoV2`, which is already growing large enough
- Change the way `MXCryptoV2` gets initialized, and have the session wait until initial keys are uploaded, just like we do in `MXLegacyCrypto`
- Import and export keys in background thread
- Skip decryption during initial sync (as we cannot have the keys yet after a login)
- A few fixes in verification flow


MatrixSDK/Crypto/Algorithms/RoomEvent/MXRoomEventDecryption.swift
```diff
@@ -0,0 +1,218 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+#if DEBUG
```
alfogrillo, line 19: Can you clarify why this type needs to exist just for debug builds?

MatrixSDK/Crypto/CryptoMachine/MXEventDecryptionResult+DecryptedEvent.swift
```diff
@@ -37,7 +37,10 @@ extension MXEventDecryptionResult {
         senderCurve25519Key = event.senderCurve25519Key
         claimedEd25519Key = event.claimedEd25519Key
         forwardingCurve25519KeyChain = event.forwardingCurve25519Chain
-        isUntrusted = event.verificationState == VerificationState.untrusted
+        
+        // Untrusted state not yet fully implemented, will equal to:
+        // `isUntrusted = event.verificationState == VerificationState.untrusted`
+        isUntrusted = false
```
alfogrillo, line 43: Is it better leaving a warning here?

MatrixSDK/Crypto/Algorithms/RoomEvent/MXRoomEventDecryption.swift
Anderas: Yea the entire `MatrixSDKCrypto` pod is only available in `DEBUG` model (see [here](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/Podfile#L19)), which is mostly for safety whilst Crypto V2 is being developed. This way we can keep merging changes to develop without worrying about accidentally shipping it to production.

alfogrillo (approved):

PR #1631, opened 2022-11-10
Threads: removed "unread_thread_notifications" from sync filters

gileluard:
Needed for https://github.com/vector-im/element-ios/issues/7066


Anderas (approved):
Would be awesome to add some unit tests for this to avoid future regressions

ismailgulek (approved):
LGTM!

PR #1632, opened 2022-11-11
Log decryption errors separately with Crypto V2

Anderas:
Log separate errors based on the type of `DecryptionError` so that it is easier to distinguish and monitor unrelated failure reasons in sentry


stefanceriu (approved):
lgtm 👍

PR #1633, opened 2022-11-11
Add poll start/end read receipts (PSB-183)

alfogrillo:
This PR adds the sending of read receipts for the following event types:
- org.matrix.msc3381.poll.start
- org.matrix.msc3381.poll.end
- m.poll.start
- m.poll.end

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -288,7 +288,12 @@ typedef void (^MXOnResumeDone)(void);
                                       kMXEventTypeStringCallHangup,
                                       kMXEventTypeStringCallReject,
                                       kMXEventTypeStringCallNegotiate,
-                                      kMXEventTypeStringSticker
+                                      kMXEventTypeStringSticker,
+                                      kMXEventTypeStringPollStart,
+                                      kMXEventTypeStringPollEnd,
+                                      // unstable event types
+                                      kMXEventTypeStringPollStartMSC3381,
+                                      kMXEventTypeStringPollEndMSC3381
                                       ];
 
         _unreadEventTypes = @[kMXEventTypeStringRoomName,


MatrixSDK/MXSession.m
```diff
@@ -288,7 +288,12 @@ - (id)initWithMatrixRestClient:(MXRestClient*)mxRestClient
                                       kMXEventTypeStringCallHangup,
                                       kMXEventTypeStringCallReject,
                                       kMXEventTypeStringCallNegotiate,
-                                      kMXEventTypeStringSticker
+                                      kMXEventTypeStringSticker,
+                                      kMXEventTypeStringPollStart,
+                                      kMXEventTypeStringPollEnd,
+                                      // unstable event types
+                                      kMXEventTypeStringPollStartMSC3381,
```
paleksandrs, line 295: Maybe it would be a good idea to add "Unstable" to the constant name instead of using a comment? E.g. 
 `kMXEventTypeStringPollStartMSC3381Unstable` and `kMXEventTypeStringPollEndMSC3381Unstable`. I am usually against comments in general because you have to maintain them, better to explain yourself with the code.

alfogrillo: I don't think is a good idea: having constants ending for `MSCxywz` is already an hint about the event type being unstable.
The reason for the comment is to try to make people to group all the unstable event type together so that it would be easier to delete them when we'll need to do it.
I don't mind to delete the comment if you think isn't very helpful btw.

paleksandrs (approved):

MatrixSDK/MXSession.m
alfogrillo: I agree we can live without it.
Removed here -> https://github.com/matrix-org/matrix-ios-sdk/pull/1633/commits/2a6a1dc66c3a22e222c39cf3edc4a4d3eb925fb9

PR #1637, opened 2022-11-15
Import progress for room keys with Crypto V2

Anderas:
Split up room keys into batches of 1000 when importing, and ensure to decrypt + import within `autoreleasepool`. This ensures we do not run out of memory when importing large key backups.

Additionally expose `importProgress` property, analogous to existing `backupProgress`, that can be used by any UI.

Tested by importing 1,000,000 keys on a device, which does grow the memory footprint considerably, but still slower than without the `autoreleasepool` and batching. The initial spike seen below is the http request, roughly 800MB for all the keys.

Screenshot 2022-11-15 at 11 47 06

### DIFF
--- a/MatrixSDK/Crypto/KeyBackup/Engine/MXCryptoKeyBackupEngine.swift
+++ b/MatrixSDK/Crypto/KeyBackup/Engine/MXCryptoKeyBackupEngine.swift
@@ -21,11 +21,21 @@ import Foundation
 import MatrixSDKCrypto
 
 class MXCryptoKeyBackupEngine: NSObject, MXKeyBackupEngine {
+    // Batch size chosen arbitrarily, will be moved to CryptoSDK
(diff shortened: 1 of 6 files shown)


MatrixSDK/Crypto/KeyBackup/Engine/MXCryptoKeyBackupEngine.swift
```diff
@@ -362,6 +382,10 @@ class MXCryptoKeyBackupEngine: NSObject, MXKeyBackupEngine {
         return try backup.exportRoomKeys(passphrase: passphrase)
     }
     
+    func importProgress() -> Progress? {
+        return activeImportProgress
```
alfogrillo, line 386:
```suggestion
        activeImportProgress
```

alfogrillo (approved):
IGTM!

PR #1638, opened 2022-11-21
NSE to persist megolm sessions

toshanmugaraj:
### Pull Request Checklist

* [ x] Pull request is based on the develop branch
Signed-off-by: Shunmugaraj S <toshanmugaraj@gmail.com>

### DIFF
--- a/MatrixSDK/Background/MXBackgroundCryptoStore.m
+++ b/MatrixSDK/Background/MXBackgroundCryptoStore.m
@@ -230,6 +230,8 @@ NSString *const MXBackgroundCryptoStoreUserIdSuffix = @":bgCryptoStore";
 - (void)storeInboundGroupSessions:(NSArray<MXOlmInboundGroupSession *>*)sessions
 {
     [bgCryptoStore storeInboundGroupSessions:sessions];
+    [cryptoStore setReadOnly:NO];
+    [cryptoStore storeInboundGroupSessions:sessions];
 }
 
 - (void)performSessionOperationWithGroupSessionWithId:(NSString*)sessionId senderKey:(NSString*)senderKey block:(void (^)(MXOlmInboundGroupSession *inboundGroupSession))block


Anderas:
Hi @toshanmugaraj , thanks for the contribution. I think this is unfortunatelly not the correct solution to the problem, nevermind the fact that making crypto store read-write could introduce further race conditions.

You note in a [comment](https://github.com/vector-im/element-ios/issues/5426#issuecomment-1316816540) that there is "no `bgCryptoStore` to `cryptoStore` transfer." You are right that there is no direct copy of the keys, however when the app launches, it will restore a [cached response](https://github.com/matrix-org/matrix-ios-sdk/blob/2c925129b3aff4b2070ac1797308c2df5e7edb41/MatrixSDK/MXSession.m#L2158) from the bgd sync which includes room keys. This response will then be processed as regular sync incl [`handleRoomKeyEvent`](https://github.com/matrix-org/matrix-ios-sdk/blob/4dc8446d7c71bcc252a009907f169133b1be1a8f/MatrixSDK/Crypto/MXCrypto.m#L860) as if the key was inserted during regular sync. So the keys are not directly transfered from the bgd to crypto store, rather they are replayed from a cached sync response, and stored in the crypto store.

If you are indeed able to reproduce the issue (I was not, following your steps), I would encourage you to see where in the replaying of the device keys we fail to process events correctly, i.e. see if we do indeed try to restore with cached response, whether it contains the to-device event with room key etc. If you need any more help let me know.

toshanmugaraj:
> Hi @toshanmugaraj , thanks for the contribution. I think this is unfortunatelly not the correct solution to the problem, nevermind the fact that making crypto store read-write could introduce further race conditions.
> 
> You note in a [comment](https://github.com/vector-im/element-ios/issues/5426#issuecomment-1316816540) that there is "no `bgCryptoStore` to `cryptoStore` transfer." You are right that there is no direct copy of the sessions, however when the app launches, it will restore a [cached response](https://github.com/matrix-org/matrix-ios-sdk/blob/2c925129b3aff4b2070ac1797308c2df5e7edb41/MatrixSDK/MXSession.m#L2158) from the bgd sync which includes inbound sessions. This response will then be processed as regular sync incl [`handleRoomKeyEvent`](https://github.com/matrix-org/matrix-ios-sdk/blob/4dc8446d7c71bcc252a009907f169133b1be1a8f/MatrixSDK/Crypto/MXCrypto.m#L860) as if the inbound session / room key was inserted during regular sync. So the sessions are not directly transfered from the bgd to crypto store, rather they are replayed from a cached sync response, and stored in the crypto store.
> 
> If you are indeed able to reproduce the issue (I was not, following your steps), I would encourage you to see where in the replaying of the cached response we fail to process events correctly, i.e. see if we do indeed try to restore with cached response, whether it contains the to-device event with room key / sessions etc. If you need any more help let me know.

It is more complex now , when source of truth is in 3 places :-) `bgCryptoStore`, `cryptoStore` and the `cached response`.

toshanmugaraj:
Recent SDK update has fixed the issues. 💯

toshanmugaraj:
> > Recent SDK update has fixed the issues. 100
> 
> Hi,
> 
> Can you specify the SDK which has the fix for olm session issue? It is [v0.24.3](https://github.com/matrix-org/matrix-ios-sdk/releases/tag/v0.24.3)?
> 
> Thanks!

It should have been fixed before v0.24.3 not sure.  Especially after this:

https://github.com/matrix-org/matrix-ios-sdk/pull/1575/files#diff-6a95981f17eb9d0318142f842e292d678077846a0ec483a55b4c20c5654bb56b

PR #1639, opened 2022-11-21
Run all Crypto tasks with default priority

Anderas:
- Run all crypto `Task`s with default priority rather than assigning `medium`
- Make sure `CryptoV2` does not get initialized more than once (which can happen during failing initial syncs)
- Remove unnecessary processing of outgoing requests


pixlwave (approved):
LGTM

PR #1641, opened 2022-11-22
Fix backup performance with Crypto V2

Anderas:
The check to see whether we have any keys to backup relies on getting the actual count of total vs backed-up keys. For large accounts this rust method is very compute intensive and blocks the main thread. This has to be improved on the rust side, but for the time being we will return a cached count synchronously and then compute the actual count on a separate thread asynchronously. This will free up the main thread.

Additionally and unrelatedly I noticed that space graph computation during sync also blocks the main thread (once again for large accounts). This is due to an innocent looking `MXRoom.isDirect` method that actually has O(n) performance. Instead of being a state property, it is a computed property that iterates through all direct rooms to determine if itself is direct or not. Changing the performance of this method is a rabbit hole I don't want to get into. Luckily the solution is quite simple, given that `MXSpaceService` can use an existing `MXSession.directRooms` method to get all of the ids at once, instead of doing essentially O(n^2) work (for every room, check every other direct room).

### DIFF
--- a/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
+++ b/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
@@ -64,6 +64,12 @@ class MXCryptoMachine {
     private let syncQueue = MXTaskQueue()
     private var roomQueues = RoomQueues()
     
+    // Temporary properties to help with the performance of backup keys checks
+    // until the performance is improved in the rust-sdk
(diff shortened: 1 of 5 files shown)


Velin92 (approved):
LGTM

Anderas:
> LGTM but I see one integration test failing.

Yea our flaky integration tests 😞. Confirmed that failing tests pass locally

PR #1643, opened 2022-11-23
Calculate sync progress state

Anderas:
Add `MXSessionSyncProgress` which tracks and calculates the state of the sync in order to determine whether we are waiting for the server response (and after how many attempts) or what is the overal percentage of data processed. For large accounts we may need to do 2-3 sync attempts with the server and data processing could take minutes, thus knowing what is the percentage completed can help users see how much longer they have to wait.

The design of sync progress is such that individual phases may choose to calculate progress however they like (e.g. handling of sync response only looks at rooms, not groups), and this single number is then incorporated in the overal sync progress, which depends on the number of phases. The design also makes it easy to add new phases in the future with ease if new areas need to report their progress.

This feature is currently under a feature flag and disabled by default.

Additionally fix some warnings with nullability and `MatrixSDKCrypto` warnings.


MatrixSDK/MXSession.m
```diff
@@ -1416,6 +1441,11 @@ - (void)serverSyncWithServerTimeout:(NSUInteger)serverTimeout
                       clientTimeout:(NSUInteger)clientTimeout
                         setPresence:(NSString*)setPresence
 {
+    if (MXSDKOptions.sharedInstance.enableSyncProgress)
+    {
+        [self.syncProgress incrementSyncAttempt];
```
stefanceriu, line 1446: You can probably just call this directly without checking, here and in other places in this file. If it's `nil` nothing happens.

stefanceriu (approved):
Looks sweet! 👏

MatrixSDK/MXSession.m
Anderas: This is true, though I have burned too many fingers on the implicit obj-c nil behaviour and prefer to be explicit where possible (e.g. incorrect typecast not actually being nil, dictionaries crashing on nil etc)

stefanceriu: Fair enough 👍

PR #1644, opened 2022-11-23
Notifications with Crypto V2

Anderas:
Extract `OlmDevice` and `MXBackgroundCryptoStore` out of `MXBackgroundSyncService` and behind a new `MXBackgroundCrypto` protocol. This protocol is similar to `MXCrypto` but with much smaller set of functionality, basically just the ability to recieve new room keys and decrypt events. All code moved out of `MXBackgroundSyncService` into `MXLegacyBackgroundCrypto` is unchanged, apart from logs.

Furthermore add `MXBackgroundCryptoV2` which uses rust-based `CryptoMachine` under the hood and is also able to recieve keys and decrypt messages. One difference compared to the legacy crypto is that crypto v2 will only use one shared store between foreground and background processes meaning there is no need to synchronize these two stores.


MatrixSDK/Background/Crypto/MXLegacyBackgroundCrypto.swift
```diff
@@ -0,0 +1,202 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+class MXLegacyBackgroundCrypto: MXBackgroundCrypto {
```
Anderas, line 19: All code in this class is moved unchaged from `MXBackgroundSyncService`

ismailgulek (approved):
LGTM 👌

PR #1645, opened 2022-11-28
Fix invites count

alfogrillo:
### Description
This PR refactor the count logic for invites making it coherent with unread messages and mentions count.

Currently for mentions the same event is counted both in:
- `*count`
- `*HighlightedCount`

Conversely Invites are counted just in:
- `*HighlightedCount`

This break the invariant:
`*count >= *HighlightedCount` 
bringing inconsistencies in the client.

After this PR invites are counted **both as highlighted and in the general count as well**.

### Related PR
https://github.com/vector-im/element-ios/pull/7088


pixlwave (approved):
Nice catch, LGTM 👍

PR #1654, opened 2022-12-07
Update to latest Crypto V2 Verification API

Anderas:
A number of changes in crypto verification API (only for Crypto V2) originating in `rust-sdk-crypto`. Unfortunatelly this is quite a lot of changes and there is no easy way to break this up. In the future we should be deprecating API in upstream SDKs before removing them entirely, which will make upgrades easier. The changes include:

- `Sas` (but not `QrCode`) verification updates its state via a callback. This makes it easier to observe updates instead of needing to manually `processUpdates`. `QrCode` will be updated in similar regard in the future but for now uses the old approach
- Some methods previously exposed on `OlmMachine` are now instead exposed on `VerificationRequest` (eg. `startSasVerification`
- A lot of types in the `rust-sdk-crypto` have changed from structs to protocols, which broke a lot of stubs in unit tests. This PR fixes that wherever possible
- Add some extra logs
- Process cached sync response with crypto v2


aringenbach (approved):
Code LGTM

PR #1661, opened 2022-12-14
Update synapse installation doc

alfogrillo:
### Description
This PR updates the documentation for installing a local synapse instance on Mac OS machines.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,12 +1,12 @@
-.. image:: https://img.shields.io/cocoapods/v/MatrixSDK?style=flat-square 
+.. image:: https://img.shields.io/cocoapods/v/MatrixSDK?style=flat-square
    :target: https://github.com/matrix-org/matrix-ios-sdk/releases
 .. image:: https://img.shields.io/cocoapods/p/MatrixSDK?style=flat-square
    :target: README.rst
-.. image:: https://img.shields.io/github/workflow/status/matrix-org/matrix-ios-sdk/Lint%20CI/develop?style=flat-square 
+.. image:: https://img.shields.io/github/workflow/status/matrix-org/matrix-ios-sdk/Lint%20CI/develop?style=flat-square
    :target: https://github.com/matrix-org/matrix-ios-sdk/actions?query=branch%3Adevelop
-.. image:: https://codecov.io/gh/matrix-org/matrix-ios-sdk/branch/develop/graph/badge.svg?token=2c9mzJoVpu 
+.. image:: https://codecov.io/gh/matrix-org/matrix-ios-sdk/branch/develop/graph/badge.svg?token=2c9mzJoVpu
    :target: https://codecov.io/gh/matrix-org/matrix-ios-sdk
-.. image:: https://img.shields.io/badge/License-Apache%202.0-yellowgreen.svg?style=flat-square 
+.. image:: https://img.shields.io/badge/License-Apache%202.0-yellowgreen.svg?style=flat-square
    :target: https://opensource.org/licenses/Apache-2.0
 
 Matrix iOS SDK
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -472,20 +472,30 @@ Then, open ``MatrixSDK.xcworkspace``.
...
 
+Before you install synapse you may need few dependencies to be installed on Mac OS:
+
+- **Homebrew**: run ``/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)”``. More information can be found here https://brew.sh
+- **python 3**: downloading the latest stable version should be fine. Download the ``.pkg`` and install it from here https://www.python.org/downloads/
+- **pipx**: with python installed run ``pip3 install --user pipx``
+- **Rust**: run ``curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh``. more information can be found here https://www.rust-lang.org/tools/install
+- **icu4c**: Run ``brew install icu4c``
+- **Update env variables for icu4c**: if you use zsh run ``echo 'export PATH="/opt/homebrew/opt/icu4c/bin:$PATH"' >> ~/.zshrc``. Otherwise try to update ``.bash_profile`` in the same way. If you can’t find ``icu4c`` in the default homebew folder try to run ``which icu4c`` to find the correct path and update the command above.
+- **pg_config**: you can get it by running ``brew install postgresql``
```
Johennes, line 491: In https://github.com/matrix-org/synapse/pull/13480/files, we postulated that libpq would be enough. Had you tried that out?

Johennes: (original issue: https://github.com/matrix-org/synapse/issues/13464)

alfogrillo: Do you mean `libpq` instead of `postgresql`?

clokep:
Would it be possible to point to the Synapse docs and ensure those are correct instead of duplicating them here?

README.rst
Johennes: Yeah, exactly

Johennes: Hm, ok sorry I may have mixed this up then. It's odd though that postgres is not mentioned in the official install instructions?

https://matrix-org.github.io/synapse/latest/setup/installation.html

alfogrillo:
> Would it be possible to point to the Synapse docs and ensure those are correct instead of duplicating them here?

@clokep for what I saw the [official documentation](https://matrix-org.github.io/synapse/latest/setup/installation.html) explains how to install synapse using prebuilt packages or as a PyPI module.
The goal of this update is being able to run a demo from sources which is useful especially for devs.
Do you think this knowledge is useful outside this repo? If yes, where do you think it should go?

clokep:
> > Would it be possible to point to the Synapse docs and ensure those are correct instead of duplicating them here?
> 
> @clokep for what I saw the [official documentation](https://matrix-org.github.io/synapse/latest/setup/installation.html) explains how to install synapse using prebuilt packages or as a PyPI module. The goal of this update is being able to run a demo from sources which is useful especially for devs. Do you think this knowledge is useful outside this repo? If yes, where do you think it should go?

There's a section on what you need to install for macOS: https://matrix-org.github.io/synapse/latest/setup/installation.html#macos

There's also information in the developer guide of requirements: https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#2-what-do-i-need

That seems to already mention rust though.

alfogrillo:
> There's a section on what you need to install for macOS: https://matrix-org.github.io/synapse/latest/setup/installation.html#macos

This seems to be under PyPI. The scope is probably different.

> There's also information in the developer guide of requirements: https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#2-what-do-i-need
> 
> That seems to already mention rust though.

Yeah it's mentioned. 
So I guess most of dependencies are covered.
I think just **icu4c** and **pg_config** aren't.

The problem I guess is that there aren't platform specific sections for devs?
If so, it seems like there isn't a good spot where to merge this info in.

stefanceriu (approved):
No matter what you decide these steps worked on my side and look good to me 👍

PR #1663, opened 2022-12-14
Verification request and QR listener in Crypto V2

Anderas:
Integrate [verification signalling](https://github.com/matrix-org/matrix-rust-sdk/pull/1272) which exposes state stream for `VerificationRequest` and `QrCode` the same way as was previously done for [`SAS`](https://github.com/matrix-org/matrix-ios-sdk/pull/1654). This makes it possible to completely remove `updatePendingVerification` method which manually compares the state of verification objects.

There are a few things that this PR does not address yet and will be resolved in future PR:
- sprinkling @MainActor in too many places now meaning I'll have to extract common functionality into a dedicated `actor` type to make concurrency more rigorous
- calling `processOutgoingRequests` from multiple places also makes it less obvious when and how this should be called (after sync, after decrypting events, ...)


MatrixSDK/Crypto/Verification/Requests/MXKeyVerificationRequestV2.swift
```diff
@@ -29,82 +29,61 @@ class MXKeyVerificationRequestV2: NSObject, MXKeyVerificationRequest {
...
-            return nil
+    private(set) var state: MXKeyVerificationRequestState = MXKeyVerificationRequestStatePending {
+        didSet {
+            if state != oldValue {
+                log.debug("\(oldValue.description) -> \(state.description)")
+                
+                DispatchQueue.main.async {
+                    NotificationCenter.default.post(name: .MXKeyVerificationRequestDidChange, object: self)
+                }
+            }
```
alfogrillo, line 40:
```suggestion
            guard state != oldValue else { return }
            log.debug("\(oldValue.description) -> \(state.description)")
                
             DispatchQueue.main.async {
                 NotificationCenter.default.post(name: .MXKeyVerificationRequestDidChange, object: self)
              }
```

MatrixSDK/Crypto/Verification/Transactions/QRCode/MXQRCodeTransactionV2.swift
```diff
@@ -26,97 +26,72 @@ class MXQRCodeTransactionV2: NSObject, MXQRCodeTransaction {
         case cannotCancel
     }
     
-    var state: MXQRCodeTransactionState {
-        qrCode.state
+    private(set) var state: MXQRCodeTransactionState = .unknown {
+        didSet {
+            if state != oldValue {
```
alfogrillo, line 31: Same `guard` consideration

alfogrillo (approved):
LGTM
(just left a couple of minor comments)

PR #1664, opened 2022-12-15
Do not track cancellation errors

Anderas:
We log errors in a lot of places across the codebase via `MXLog.error(... error)` passing in any error that was returned from some api, incl rest clients that may return cancellation errors. These are not userful to track in analytics as they occur due to some user action. Instead of filtering them out in the many many places where we log errors, we will filter them out all at once in `MXAnalyticsDestination`

### DIFF
--- a/MatrixSDK/Utils/Logs/MXAnalyticsDestination.swift
+++ b/MatrixSDK/Utils/Logs/MXAnalyticsDestination.swift
@@ -30,10 +30,18 @@ class MXAnalyticsDestination: BaseDestination {
     
     override func send(_ level: SwiftyBeaver.Level, msg: String, thread: String, file: String, function: String, line: Int, context: Any? = nil) -> String? {
         let message = super.send(level, msg: msg, thread: thread, file: file, function: function, line: line, context: context)
-        #if !DEBUG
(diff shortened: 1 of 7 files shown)


MatrixSDK/Utils/Logs/MXAnalyticsDestination.swift
```diff
@@ -30,12 +30,20 @@ class MXAnalyticsDestination: BaseDestination {
     
     override func send(_ level: SwiftyBeaver.Level, msg: String, thread: String, file: String, function: String, line: Int, context: Any? = nil) -> String? {
         let message = super.send(level, msg: msg, thread: thread, file: file, function: function, line: line, context: context)
-        #if !DEBUG
```
Anderas, line 33: Moving the compiler check to the parent, otherwise cannot write unit tests for this class

pixlwave (approved):
LGTM, sensible change 👏

Maybe worth doing the same for `NSURLErrorNotConnectedToInternet`, but presumably this is driven by looking at the reports on Sentry.

Anderas:
> LGTM, sensible change 👏
> 
> Maybe worth doing the same for `NSURLErrorNotConnectedToInternet`, but presumably this is driven by looking at the reports on Sentry.

Yea exactly. And actually network connectivity errors may still be useful to track, because they are not caused / expected by the user, so typically result in some kind of visible error, unlike explicit cancellations.

PR #1673, opened 2022-12-23
Capture decryption errors (PSG-1023)

alfogrillo:
### Description

This PR adds a logic to consider events not decrypted inside a `PollAggregator`.

The final goal is to warn the user when a poll is affected by decryption errors since otherwise the votes count may be different across sessions of the same user or different users in the same room.

**Related PR**: https://github.com/vector-im/element-ios/pull/7206

### DIFF
--- a/MatrixSDK/MXSession.m
+++ b/MatrixSDK/MXSession.m
@@ -71,7 +71,6 @@ NSString *const kMXSessionDidUpdateGroupSummaryNotification = @"kMXSessionDidUpd
 NSString *const kMXSessionDidUpdateGroupRoomsNotification = @"kMXSessionDidUpdateGroupRoomsNotification";
 NSString *const kMXSessionDidUpdateGroupUsersNotification = @"kMXSessionDidUpdateGroupUsersNotification";
 NSString *const kMXSessionDidUpdatePublicisedGroupsForUsersNotification = @"kMXSessionDidUpdatePublicisedGroupsForUsersNotification";
-NSString *const kMXSessionDidFailToDecryptEventsNotification = @"kMXSessionDidFailToDecryptEventsNotification";
 
 NSString *const kMXSessionNotificationRoomIdKey = @"roomId";
 NSString *const kMXSessionNotificationGroupKey = @"group";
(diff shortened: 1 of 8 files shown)


MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -52,6 +52,7 @@ public class PollAggregator {
     
     private var events: [MXEvent] = []
     private var hasBeenEdited = false
+    private var relatedErroredEvents: Set<String> = []
```
Anderas, line 55: Optional: `Errored` does seem to be grammatical, though I find it a bit unusual and almost looks like a typo. Maybe `Failed` would be better? Alternatively if this only captures decryption errors, it could even be `relatedUndecryptableEvents`

MatrixSDK/MXSession.m
```diff
@@ -5000,6 +5002,12 @@ - (void)decryptEvents:(NSArray<MXEvent*> *)events
                 }
             }
             
+            if (failedEvents.count > 0) {
+                [NSNotificationCenter.defaultCenter postNotificationName:kMXSessionDidFailToDecryptEventsNotification
```
Anderas, line 5006: There is already class `DecryptionFailureTracker` that deals with decryption failures (and logs them to Posthog analytics). It has some additional checks (i.e. not report if user has not joined the room) and is triggered when rendering the timeline (whereas `MXSession decrypt` could be called in other circumstances as well. I think it would be better to group this functionality instead of reporting failures from multiple places. Could you have a look whether the notification could be posted from there?

MatrixSDK/Room/Polls/PollAggregator.swift
alfogrillo: Yeah, it capture just crypto errors.
`relatedUndecryptableEvents` looks better. ;-)

alfogrillo: Applied -> https://github.com/matrix-org/matrix-ios-sdk/pull/1673/commits/7a29bc6bc5f6972427c8f6c71af39c2fe051eba9

MatrixSDK/MXSession.m
Anderas: Hm I understand. I think the truth is that this has not been really consolidated properly before, so different parts of the SDK / app are doing their own tracking. Ideallly the tracking would be handled inside `MXCrypto`, but since we now have two versions, what you did makes sense, especially since the `MXSession` is already collecting `failedEvents` 👍

Anderas (approved):

PR #1674, opened 2023-01-04
Render m.poll.end event (PSG-196)

alfogrillo:
This PR:
- Adds a new init for the `PollAggregator` so that it will be able to aggregate a poll starting from a `m.poll.end` event
- Enable replies for a `m.poll.end` event


stefanceriu (approved):
Looks good to me 👍

PR #1675, opened 2023-01-06
Replace `MXTaskQueue` with `MXSerialTaskScheduler`

Anderas:
I have previously added `MXTaskQueue` (a serial queue for executing `Task`s) to `MXCryptoMachine` in order to control access to shared resources and prevent certain types of data races. Whilst this worked correctly, I later realized that a queue was actually the incorrect data structure to use as it allows the enquing of an unlimited number of future tasks. In reality we should only have at most one "next" task that will start after the current one completes. This will prevent performing a potentially large number of unnecessary operations, when the client schedules a task multiple times.

Since the `MXTaskQueue` is not being used anywhere else, I am deleting it entirely.


Anderas:
Closing as I haven't considered use case with varying parameters yet

Issue #1670, opened 2022-12-20
CryptoV2 is not tracking users on membership changes, and only querying keys for unknown users if forceDownload is true


BillCarsonFr:
MXCryptoV2 is not listening to room membership events of encrypted rooms in order to start tracking their devices (ask olmMachine to query keys).
That is problematic if computeE2ERoomSummaryTrust is FALSE, as the keys will be downloaded later when sending a message, and that could affect message sending time.

Also `MXCryptoV2.downloadKeys` looks like it's not behaving as expected?
If forceDownload is false, it will only return keys in cache.
It should in this case check what keys are missing from the provided user list, and only querying for them and return the newly downloaded one + the one that are already in cache

Issue #1668, opened 2022-12-19
MXCrossSigningV2 calls to keys/query on refreshState


BillCarsonFr:
Looks like each time the app enters foreground (or is launched), a keys/query for our own user is done 4 times in a row

PR #1676, opened 2023-01-09
Add keys query scheduler

Anderas:
Fixes https://github.com/matrix-org/matrix-ios-sdk/issues/1668 and partially resolves https://github.com/matrix-org/matrix-ios-sdk/issues/1670

Add `MXKeysQueryScheduler` used with crypto V2, which serves similar purpose as `MXDeviceList` serves in legacy crypto, albeit only focusing on optimizing `keys/query` requests.

Specifically it will:
- ensure that only one keys/query request is in-flight at any given time
- all further requests are aggregated, and if they contain any new users, will be executed right after the first one finishes

These optimizations should ensure that the application does not make unnecessary amout of queries, as we currently call the `downloadKeys` from far too many places. Note that I have [previously](https://github.com/matrix-org/matrix-ios-sdk/pull/1675) tried to solve the current/next task in more generic way, but was missing the additional need to aggregate further users. So for now I am adding a case specific `MXKeysQueryScheduler` which may later evolve to more generic `MXSerialTaskScheduler`

In general the application should never query keys directly, except for very few cases (starting new room, refreshing cross signing), but this is not unfortunatelly the case at the moment. To start moving in the right direction I split out `downloadKeysIfNecessary` and a deprecated `reloadKeys` instead of passing `forceDownload` parameter. This should make it clearer which API to use.


gileluard (approved):
lgtm 👍

PR #1681, opened 2023-01-12
Create Crypto V2 Migration Data

Anderas:
Create `MigrationData` object defined in rust-crypto-sdk, which is populated with existing data (account, sessions, cross signing ...), and used to migrate into the rust-based `OlmMachine`. Note that this PR only adds the creation and migration of this object, but no codepath will currently trigger the migration itself. This will be done in a future PR.

Main changes in this PR:
- add `MXCryptoMigrationStore` which will use legacy crypto store (`MXCryptoStore`) to construct rust-based `MigrationData`
- add `MXCryptoMigrationV2` which will actually perform the migration from legacy store to rust-based crypto
- change `MXOlmSession` and `MXCryptoStore` API to include `deviceKey` on the session itself rather than just stored internally in realm

To be done in future PR:
- More integration tests for cases like migrating cross-signing status, trusting own devices, restoring keys from backup etc. Some of these require a few bugfixes in the rust-crypto
- Actually perfom the migration on app start when necessary
- Optimizations for migrating large accounts (i.e. migrating keys in batches)

### DIFF
--- a/MatrixSDKTests/Crypto/CryptoMachine/MXKeysQuerySchedulerUnitTests.swift
+++ b/MatrixSDKTests/Crypto/CryptoMachine/MXKeysQuerySchedulerUnitTests.swift
@@ -71,37 +71,41 @@ class MXKeysQuerySchedulerUnitTests: XCTestCase {
         users: Set<String>,
         completion: @escaping (Response) -> Void,
         failure: ((Swift.Error) -> Void)? = nil
-    ) {
-        Task.detached {
-            do {
-                let result = try await self.scheduler.query(users: users)
-                completion(result)
-            } catch {
-                failure?(error)
+    ) async {
+        return await withCheckedContinuation { continuation in
+            Task.detached {
+                continuation.resume()
+                
+                do {
+                    let result = try await self.scheduler.query(users: users)
+                    completion(result)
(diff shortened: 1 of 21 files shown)


MatrixSDKTests/Crypto/CryptoMachine/MXKeysQuerySchedulerUnitTests.swift
```diff
@@ -71,71 +71,75 @@ class MXKeysQuerySchedulerUnitTests: XCTestCase {
...
         failure: ((Swift.Error) -> Void)? = nil
-    ) {
-        Task.detached {
-            do {
-                let result = try await self.scheduler.query(users: users)
-                completion(result)
-            } catch {
-                failure?(error)
+    ) async {
+        return await withCheckedContinuation { continuation in
```
Anderas, line 75: Unrelated change but sometimes these tests fail. I think this is because calling `Task.detached` several times in a row may not result in them being executed in the same order. To solve this we make the function `async` and completed as soon as the `Task.detached` is triggered (but not when the task completes its work). Its an await on enquing rather than completing the async work.

aringenbach (approved):
LGTM

PR #1685, opened 2023-01-18
Reply to ended polls events (PSG-1131)

alfogrillo:
### Description

This PR ensure that placeholders in reply events are always sent in the English version.
Matrix client receiving events containing this placeholder may chose to localize them or not.
It doesn't make sense localizing placeholders on the sender side since the receivers could prefer a different language.

Affected placeholders are:
- "In reply to"
- "Ended poll" (new one, it's used when replying to `m.poll.end` events)


pixlwave (approved):
LGTM, nice catch 👍

PR #1686, opened 2023-01-18
Rename MXSessionSyncProgress

Anderas:
Rename `MXSessionSyncProgress` to `MXSessionStartupProgress` as it more reflects its purpose. A future PR will be adding `migrationData` stage which does not really belong into the "sync" state but rather during "startup" stage where loading screen is shown to the user, before the session is ready / interactable.


alfogrillo (approved):
IGTM!

PR #1687, opened 2023-01-18
Batch migrate olm and megolm sessions

Anderas:
When migrating large accounts from libolm to rust (100K + room keys) the application consumes a lot of memory, first by fetching data from realm (all at once), then converting to rust objects, and then migrating to rust crypto store. For these large accounts we may even run out of memory before the migration completes.

To solve this add new methods to crypto store that can enumerate batches of sessions (with configurable batch size) that are wrapped in `autoreleasepool`. During migration we continuously enumerate olm and megolm sessions, and no matter how many there are, we can keep a mostly constant memory footprint.

Due to the number of work necessary, migration now becomes an asynchronous task that needs to report its own progress. This is eventually tied into existing [MXSessionStartupProgress](https://github.com/matrix-org/matrix-ios-sdk/pull/1686) and [displayed](https://github.com/vector-im/element-ios/pull/7286) on the loading page


Velin92 (approved):

MatrixSDK/Crypto/Migration/MXCryptoMigrationV2.swift
```diff
@@ -60,7 +64,69 @@ class MXCryptoMigrationV2: NSObject {
             progressListener: self
         )
         
+        log.debug("Migrating olm sessions in batches")
+        
+        // How much does migration of olm vs megolm sessions contribute to the overall progress
+        let olmToMegolmProgressRatio = 0.25
```
Velin92, line 70: We discussed this on element but yeah this could be a computed value of `olm / (olm + megolm`

Anderas: Fix applied [here](https://github.com/matrix-org/matrix-ios-sdk/pull/1692/files#diff-85e898e29304782019f3a309bd00f3f2631df8d1d0e83a306457c72c0101dae8R71)

PR #1688, opened 2023-01-20
Implement MSC3912: redaction with relations

nimau:
SDK support for [MSC3912](https://github.com/matrix-org/matrix-spec-proposals/pull/3912)

### DIFF
--- a/MatrixSDK/Data/MXRoom.h
+++ b/MatrixSDK/Data/MXRoom.h
@@ -977,7 +977,7 @@ FOUNDATION_EXPORT NSInteger const kMXRoomInvalidInviteSenderErrorCode;
                   withRelations:(NSArray<NSString *>*)relations
                          reason:(NSString*)reason
                         success:(void (^)(void))success
-                        failure:(void (^)(NSError *error))failure NS_REFINED_FOR_SWIFT;
+                        failure:(void (^)(NSError *error))failure;
 
 /**
  Report an event in this room.
--- a/MatrixSDK/Data/MXRoom.h
+++ b/MatrixSDK/Data/MXRoom.h
@@ -962,6 +962,24 @@ FOUNDATION_EXPORT NSInteger const kMXRoomInvalidInviteSenderErrorCode;
                         failure:(void (^)(NSError *error))failure NS_REFINED_FOR_SWIFT;
 
 /**
+ Redact an event and all related events in this room.
+ 
+ @param eventId the id of the redacted event.
+ @param relations the list of relation types (optional).
+ @param reason the redaction reason (optional).
+ 
+ @param success A block object called when the operation succeeds.
+ @param failure A block object called when the operation fails.
+
+ @return a MXHTTPOperation instance.
+ */
+- (MXHTTPOperation*)redactEvent:(NSString*)eventId
+                  withRelations:(NSArray<NSString *>*)relations
+                         reason:(NSString*)reason
+                        success:(void (^)(void))success
+                        failure:(void (^)(NSError *error))failure NS_REFINED_FOR_SWIFT;
+
+/**
  Report an event in this room.
 
  @param eventId the id of the event event.
--- a/MatrixSDK/Data/MXRoom.m
+++ b/MatrixSDK/Data/MXRoom.m
@@ -2080,9 +2080,12 @@ NSInteger const kMXRoomInvalidInviteSenderErrorCode = 9002;
 {
     BOOL isFeatureSupported = (mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations || mxSession.store.supportedMatrixVersions.supportsRedactionWithRelationsUnstable);
     BOOL isFeatureStable = mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations;
-    if (isFeatureSupported) {
+    if (isFeatureSupported)
+    {
         return [mxSession.matrixRestClient redactEvent:eventId withRelations:relations inRoom:self.roomId reason:reason txnId:nil featureIsStable:isFeatureStable success:success failure:failure];
-    } else {
+    }
+    else
+    {
         MXLogDebug(@"[MXRoom] redaction with relations is not supported (MSC3912)");
         return [self redactEvent:eventId reason:reason success:success failure:failure];
     }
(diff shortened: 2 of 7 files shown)


pixlwave (approved):
LGTM 😎

MatrixSDK/Data/MXRoom.m
```diff
@@ -2072,6 +2072,22 @@ - (MXHTTPOperation*)redactEvent:(NSString*)eventId
...
 
+- (MXHTTPOperation*)redactEvent:(NSString*)eventId
+                  withRelations:(NSArray<NSString *>*)relations
+                         reason:(NSString*)reason
+                        success:(void (^)(void))success
+                        failure:(void (^)(NSError *error))failure
+{
+    BOOL isFeatureSupported = (mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations || mxSession.store.supportedMatrixVersions.supportsRedactionWithRelationsUnstable);
+    BOOL isFeatureStable = mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations;
+    if (isFeatureSupported) {
```
gileluard, line 2083:
```suggestion
    if (isFeatureSupported)
    {
```

MatrixSDK/Data/MXRoom.m
```diff
@@ -2072,6 +2072,22 @@ - (MXHTTPOperation*)redactEvent:(NSString*)eventId
...
+        return [mxSession.matrixRestClient redactEvent:eventId withRelations:relations inRoom:self.roomId reason:reason txnId:nil featureIsStable:isFeatureStable success:success failure:failure];
+    } else {
```
gileluard, line 2085:
```suggestion
    }
    else
    {
```

MatrixSDK/MXRestClient.m
```diff
@@ -3047,6 +3047,53 @@ - (MXHTTPOperation*)redactEvent:(NSString*)eventId
...
+                      [MXTools encodeURIComponent:txnId]];
+    
+    // All query parameters are optional. Fill the request parameters on demand
+    NSMutableDictionary *parameters = [NSMutableDictionary dictionary];
+    if (reason)
+    {
+        parameters[@"reason"] = reason;
+    }
+    
+    if (relations && [relations count] > 0) {
```
gileluard, line 3078:
```suggestion
    if (relations && [relations count] > 0)
    {
```

MatrixSDK/Data/MXRoom.m
```diff
@@ -2072,6 +2072,22 @@ - (MXHTTPOperation*)redactEvent:(NSString*)eventId
...
+        MXLogDebug(@"[MXRoom] redaction with relations is not supported (MSC3912)");
+        return [self redactEvent:eventId reason:reason success:success failure:failure];
```
gileluard, line 2087: Shouldn't we mark this method as deprecated since it is called by `- (MXHTTPOperation*)redactEvent:(NSString*)eventId withRelations:(NSArray<NSString *>*)relations reason:(NSString*)reason success:(void (^)(void))success failure:(void (^)(NSError *error))failure` if the server doesn't support this new MSC?

gileluard (approved):
Some cosmetic changes but LGTM :)

giomfo (commented):
LGTM, only one comment

MatrixSDK/Data/MXRoom.m
giomfo: This is too early to mark it as deprecated. This would be the case eventually when MSC3912 will be stable

giomfo: After thinking we think you're right. We should not use the previous endpoint anymore.
So we will call the most [recent endpoint](https://spec.matrix.org/unstable/client-server-api/#put_matrixclientv3roomsroomidredacteventidtxnid) defined for the redaction by ignoring the relations

@nimau will update his PR

giomfo: Thx @gileluard , I checked this point with @nimau. Finally he will call here the new implemented redact function (based on the actual [redact endpoint](https://spec.matrix.org/latest/client-server-api/#put_matrixclientv3roomsroomidredacteventidtxnid)) by ignoring the relations

MatrixSDK/Data/MXRoom.m
```diff
@@ -2072,6 +2072,22 @@ - (MXHTTPOperation*)redactEvent:(NSString*)eventId
...
+- (MXHTTPOperation*)redactEvent:(NSString*)eventId
+                  withRelations:(NSArray<NSString *>*)relations
+                         reason:(NSString*)reason
+                        success:(void (^)(void))success
+                        failure:(void (^)(NSError *error))failure
+{
+    BOOL isFeatureSupported = (mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations || mxSession.store.supportedMatrixVersions.supportsRedactionWithRelationsUnstable);
+    BOOL isFeatureStable = mxSession.store.supportedMatrixVersions.supportsRedactionWithRelations;
+    if (isFeatureSupported) {
+        return [mxSession.matrixRestClient redactEvent:eventId withRelations:relations inRoom:self.roomId reason:reason txnId:nil featureIsStable:isFeatureStable success:success failure:failure];
```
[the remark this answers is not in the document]
nimau: Thx @pixlwave, I'll update it.

MatrixSDK/Data/MXRoom.h
```diff
@@ -961,6 +961,24 @@ FOUNDATION_EXPORT NSInteger const kMXRoomInvalidInviteSenderErrorCode;
...
+ @param success A block object called when the operation succeeds.
+ @param failure A block object called when the operation fails.
+
+ @return a MXHTTPOperation instance.
+ */
+- (MXHTTPOperation*)redactEvent:(NSString*)eventId
+                  withRelations:(NSArray<NSString *>*)relations
+                         reason:(NSString*)reason
+                        success:(void (^)(void))success
+                        failure:(void (^)(NSError *error))failure NS_REFINED_FOR_SWIFT;
```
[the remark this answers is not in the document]
nimau: @giomfo, @pixlwave I removed the NS_REFINED_FOR_SWIFT in the last commit as this method is not exposed in MXRoom.swift.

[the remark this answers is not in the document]
pixlwave: Oh my bad, I missed that. Implementing it in Swift would probably be best to my mind. Might be possible to add the extra parameter to the existing method with a default value so should be simple enough.

MatrixSDK/Data/MXRoom.m
nimau: @gileluard: I fixed it in the last commit. Now the fallback (if the MSC3912 is not supported) will be on the new endpoint but without passing any relation types.

gileluard (approved):
lgtm 👍

PR #1689, opened 2023-01-20
Extract room event encryption

Anderas:
Extract all encryption functionality (incl. setting encryption, sharing keys, encrypting content) out from `MXCryptoV2` and into `MXRoomEventEncryption`. This makes for a better separation of concerns as well as testability of the core functionality.

Some additional changes in this PR include moving objects around into new files (e.g. `MXCryptoV2Factory`)


stefanceriu (approved):
lgtm 👍

PR #1692, opened 2023-01-23
Migration improvements in Crypto V2

Anderas:
A set of unrelated improvements when testing crypto v2 migrations:
- ensure sync response is handled on the main thread for now (not entirely thread safe in rust atm)
- add integration tests for migration to ensure only existing users are migrated and only once
- do not force download keys when computing trust summary
- change the weight ratio of olm to megolm to be calculated dynamically from total amount of sessions
- ensure we report startup progress only during initial syncs


gileluard (approved):
lgtm

PR #1694, opened 2023-01-24
More migration fixes for Crypto V2

Anderas:
Some further changes when testing account migration from libolm to rust:
- remove `open` method from `MXCryptoStore` that is empty in all implementations of this protocol. This was probably used in the past but is clearly not doing anything, and removing it reduces objc callback nesting by one level
- add `hasSecretWithSecretId` on the `MXCryptoStore` protocol. For legacy store this is just a convenience API, but for crypto V2 it calls a different (and faster) rust API
- some changes in key verification management, where we only `processOutgoingRequests` and `downloadKeysIfNecessary` from within the manager, rather than `MXCryptoV2`, which makes for a better logic encapsuation

### DIFF
--- a/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
+++ b/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
@@ -530,7 +530,7 @@ extension MXCryptoMachine: MXCryptoVerifying {
         try machine.receiveUnencryptedVerificationEvent(event: string, roomId: roomId)
         
         // Out-of-sync check if there are any verification events to sent out as a result of
-        // the event just recieved
+        // the event just received
         for request in try machine.outgoingRequests() {
             if case .roomMessage = request {
                 try await handleRequest(request)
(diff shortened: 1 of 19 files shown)


MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -523,15 +522,20 @@ extension MXCryptoMachine: MXCryptoCrossSigning {
...
         }
-        do {
-            try machine.receiveUnencryptedVerificationEvent(event: string, roomId: roomId)
-        } catch {
-            log.error("Error receiving unencrypted event", context: error)
+        
+        try machine.receiveUnencryptedVerificationEvent(event: string, roomId: roomId)
+        
+        // Out-of-sync check if there are any verification events to sent out as a result of
+        // the event just recieved
```
aringenbach, line 533:
```suggestion
        // the event just received
```

MatrixSDK/Crypto/Verification/MXKeyVerificationManagerV2.swift
```diff
@@ -267,25 +267,30 @@ class MXKeyVerificationManagerV2: NSObject, MXKeyVerificationManager {
...
-            handleIncomingVerification(userId: event.sender, flowId: flowId)
-            return event.sender
+            await handleIncomingVerification(userId: event.sender, flowId: flowId)
+            newUserId = event.sender
         } else {
-            return nil
+            newUserId = nil
+        }
+
+        // If we recieved a verification event from a new user we do not yet track
```
aringenbach, line 290:
```suggestion
        // If we received a verification event from a new user we do not yet track
```

aringenbach (approved):
LGTM

PR #1696, opened 2023-01-25
Flescio/7253 add mar kas unread option for rooms

flescio:
Needed for https://github.com/vector-im/element-ios/issues/7253


Anderas:
As discussed elsewhere I would suggest two improvements:
- `MXMemoryStore` should not be instructing `MXFileStore` when to set (via a method only implemented by `MXFileStore`). Instead memory store merely mutates a dictionary, and lets file store save at opportune moments in the same way that other data is handled (filters, state ... )
- the double negative of `unmarkAsUnread` is somewhat unreadable/confusing, in one place the method even calls related `markAsRead`. I would suggest instead something like `resetUnreadMarkers` or `clearUnreadMarkers`

MatrixSDK/Data/MXRoom.h
```diff
@@ -1304,6 +1304,23 @@ Remove a tag applied on an event of the room
  */
 - (void)acknowledgeEvent:(MXEvent*)event andUpdateReadMarker:(BOOL)updateReadMarker;
 
+/**
+ It will set the entire room as unread.
+ This is only local since it's not possible to remove from the server the read events.
+ */
+-(void)setUnread;
```
Anderas, line 1311: Would be good to align with the `MXStore` api, either having `set/resetUnread` or `set/resetUnreadMarker` everywhere

MatrixSDK/Data/MXRoom.m
```diff
@@ -3348,8 +3348,26 @@ - (void)acknowledgeEvent:(MXEvent*)event andUpdateReadMarker:(BOOL)updateReadMar
     }
 }
 
+- (void)setUnread
+{
+    [mxSession.store setUnreadMarkerForRoom:self.roomId];
+    [mxSession.store commit];
```
Anderas, line 3354: Probably won't be an issue in practice (`store` is always `MXFileStore`), but `commit` is marked as `optional` on the protocol, so should be surrounded with `if ([mxSession.store respondsToSelector:@selector(commit)])` for safety as is done elsewhere in that file

MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
```diff
@@ -998,6 +1000,36 @@ - (RoomThreadedReceiptsStore*)getOrCreateRoomThreadedReceiptsStore:(NSString*)ro
     return threadedStore;
 }
 
+- (void)setUnreadMarkerForRoom:(nonnull NSString*)roomId;
```
Anderas, line 1003: If these 2 methods just call `super` there is no need to override them, and can be safely deleted.

MatrixSDK/MXSession.h
```diff
@@ -925,6 +925,11 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
                                                     success:(void (^)(BOOL canEnableE2E))success
                                                     failure:(void (^)(NSError *error))failure;
 
+/**
+ it will return if the room is marked unread by the user
+ */
+- (BOOL) isRoomMarkedAsUnread:(NSString*)roomId;
```
Anderas, line 931: Remove space `(BOOL) isRoom...` => `(BOOL)isRoom...` for consistency

MatrixSDKTests/MXRoomTests.m
```diff
@@ -879,6 +880,101 @@ - (void)testUpgrade
     }];
 }
 
+- (void)testMarkUnread
```
Anderas, line 883: Great to include tests for this 👍

Anderas (commented):
I left just a few comments regarding style, naming and consistency, but the overal change looks good to me

Anderas (approved):

PR #1697, opened 2023-01-26
Upload fallback keys with Crypto V2

Anderas:
Upload fallback keys via `keys/upload` whenever present in the body of `UploadKeysRequest`.

Additionally:
- move `MXKeysQueryScheduler` into `MXCryptoRequests` (rather than the other way around), which hides away the details of making `keys/query` requests
- rename `MXCryptoSDKLogger`
- additional log messages

### DIFF
--- a/MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
+++ b/MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
@@ -60,6 +60,8 @@ class MXBackgroundCryptoV2: MXBackgroundCrypto {
           - to-device events : \(syncResponse.toDevice?.events.count ?? 0)
           - devices changed  : \(syncResponse.deviceLists?.changed?.count ?? 0)
           - devices left     : \(syncResponse.deviceLists?.left?.count ?? 0)
+          - one time keys    : \(syncResponse.deviceOneTimeKeysCount?[kMXKeySignedCurve25519Type] ?? 0)
+          - fallback keys    : \(syncResponse.unusedFallbackKeys ?? [])
         """
         log.debug(details)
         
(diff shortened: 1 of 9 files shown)


MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
```diff
@@ -60,6 +60,8 @@ class MXBackgroundCryptoV2: MXBackgroundCrypto {
           - to-device events : \(syncResponse.toDevice?.events.count ?? 0)
           - devices changed  : \(syncResponse.deviceLists?.changed?.count ?? 0)
           - devices left     : \(syncResponse.deviceLists?.left?.count ?? 0)
+          - one time keys    : \(syncResponse.deviceOneTimeKeysCount?[kMXKeySignedCurve25519Type] ?? 0)
+          - fallback keys    : \(syncResponse.unusedFallbackKeys ?? [])
```
pixlwave, line 64: I'm sure its not, but double checking: Is this going to log keys that are sensitive?

pixlwave (approved):
Code looks good to me, although I don't have a huge amount of context around the changes :)

MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
poljar: Those keys are public, after all they get uploaded to the server.

Anderas: Yea, actually in this particular log we only print the count of keys (or type in the case of fallback keys), but I have also started logging the actual keys elsewhere so your question is valid (and it is safe as we only print hte public keys)

PR #1699, opened 2023-01-27
Set passphrase for the crypto store

Anderas:
Use existing `MXKeyProvider` mechanism to get a new key that is used as base64-encoded passphrase when migrating or creating `OlmMachine` in crypto V2. This means all data stored in the underlying sqlite db will be encrypted.

Additionally:
- listen to room membership changes and track new users
- listen to room encryption events and set encryption if necessary


alfogrillo (approved):
LGTM!
Just noticing that almost all the GH workflows seem broken. :/

Anderas:
> LGTM! Just noticing that almost all the GH workflows seem broken. :/

Thanks. Yea some flaky tests + one I could not fix due to how OlmKit delegates keys. Will fix this in a separate PR, and disabled that test for now (not in production yet)

PR #1701, opened 2023-01-30
Import legacy backup in batches with progress

Anderas:
Import room backups in batches and report progress continuously.

Note that even with this change memory growns linearly during import (albeit with a smaller slope), which means we are either leaking some memory or filling up in-memory cache somewhere in the crypto layer. Either way this should make imports slightly less memory hungry, and compute the progress regardless.

This PR also contains some small fixes for crypto v2 (logs + tracked users optimization)


Velin92 (approved):
LGTM

PR #1704, opened 2023-01-31
New CryptoMachine on each background operation

Anderas:
The background crypto process (used to decrypt notifications) relies on several rust-crypto-sdk methods that are not multiprocess safe. In particular, since the background's `MXCryptoMachine` may live in memory for days, whereas foreground's will be recreated on each app launch, we may end up in a situation where foreground has written changes to disk that the background machine does not pick up. For more details see https://github.com/matrix-org/matrix-rust-sdk/issues/1415

This will be fixed in the future in the rust-sdk via a `proc_lock` or similar multiprocess lock, but in a meanwhile a relatively quick solution is to recreate `MXCryptoMachine` on each operation (recieve sync, decrypt event), ensuring that each time we have relatively fresh data taken straight from disk.

We could addionally add `NSFileCoordinator` to provide full mutual exclusion between the processes, but this adds some amount of complexity that will be eventually moved out to rust anyway. The chance of foreground and background processing sync at the same time is minimized by explicit `session.state` checks.

### DIFF
--- a/MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
+++ b/MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
@@ -27,29 +27,13 @@ class MXBackgroundCryptoV2: MXBackgroundCrypto {
         case missingCredentials
     }
     
-    private let machine: MXCryptoMachine
+    private let credentials: MXCredentials
+    private let restClient: MXRestClient
     private let log = MXNamedLog(name: "MXBackgroundCryptoV2")
     
-    init(credentials: MXCredentials, restClient: MXRestClient) throws {
-        guard
-            let userId = credentials.userId,
-            let deviceId = credentials.deviceId
-        else {
-            throw Error.missingCredentials
-        }
-        
-        // `MXCryptoMachine` will load the same store as the main application meaning that background and foreground
-        // sync services have access to the same data / keys. Possible race conditions are handled internally.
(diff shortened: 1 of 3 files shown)


nimau:
Hi @Anderas, I have just one question about throwing the error. Since I'm not very familiar with the CryptoMachine and what the side effects of creating a new one each time might be, it might be a good idea to request a second reviewer.

Anderas:
> Hi @Anderas, I have just one question about throwing the error. Since I'm not very familiar with the CryptoMachine and what the side effects of creating a new one each time might be, it might be a good idea to request a second reviewer.

No worries, I added one more reviewer. The cost of creating a new machine each time is mostly performance, it will obviously take a bit longer, though given the rate of notification delivery is not rapid, this is relatively minor cost.

Even a small cost is not ideal though and we will eventually fix the underlying problem directly in rust-sdk (by ensuring there is no stale cache). This is basically just a quick solution whist crypto v2 is still a labs feature.

What is your specific question about throwing an error?

pixlwave (approved):
Sounds reasonable to me 👍

nimau:
> 



> What is your specific question about throwing an error?

I was just wondering why we don't throw the error if something bad happens during handleSync. But since that was already the case, I guess it's ok.

Anderas:
Something weird happened with crypto tests (even locally), not related to this PR. I'm investigating it separately so going to merge this PR anyway

PR #1705, opened 2023-02-01
Fix protocol conformance

Anderas:
A recently introduced [protocol change](https://github.com/matrix-org/matrix-ios-sdk/pull/1696/files#diff-b475ca0411f0b9bc774120fba6703e2efe8a6cf1d71c982c7baa8b450501def0R179-R198) was not correctly implemented on all classes, leading to some integration tests to fail.


flescio (approved):
Thanks

PR #1707, opened 2023-02-02
Fix JSONDictionary of keys query responses

Anderas:
When downloading user keys via `keys/query` we sometimes use `downloadByChunks` method that makes several HTTP requests and combines the result into a single `MXKeysQueryResponse`. The Crypto V2 relies on getting the value of this response via `JSONDictionary` property, however this property was previously just returning whatever initial value was used to create it, and is nil when combining multiple responses together.

To solve this we cannot rely on stored property but rather have to compute the `JSONDictionary` out of the current values of the response. Additionally there was a minor bug in `downloadByChunks` that would trigger this problem even if we are downloading exactly chunk-sized number of users

### DIFF
--- a/MatrixSDK/Categories/MXRestClient+Extensions.swift
+++ b/MatrixSDK/Categories/MXRestClient+Extensions.swift
@@ -35,7 +35,7 @@ public extension MXRestClient {
                              failure: @escaping (_ error: NSError?) -> Void) -> MXHTTPOperation {
         
         // Do not chunk if not needed
-        if users.count < chunkSize {
+        if users.count <= chunkSize {
             return self.downloadKeys(forUsers: users, token: token) { response in
                 switch response {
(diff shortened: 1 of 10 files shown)


MatrixSDK/Categories/MXRestClient+Extensions.swift
```diff
@@ -35,7 +35,7 @@ public extension MXRestClient {
                              failure: @escaping (_ error: NSError?) -> Void) -> MXHTTPOperation {
         
         // Do not chunk if not needed
-        if users.count < chunkSize {
+        if users.count <= chunkSize {
```
Anderas, line 38: Naughty naughty. Will write a test for this in a separate PR as it requires some integration test stubbing

flescio (approved):
LGTM

PR #1708, opened 2023-02-02
Enable Crypto SDK for production

Anderas:
This is a big moment !

- update `MatrixSDKCrypto` pod to 0.2.0 which is now production ready (though still considered beta)
- remove all DEBUG compiler flags as we are ready to ship to production
- enable `isCryptoSDKAvailable`, which makes crypto 2.0 / SDK visible as a labs flag

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
       ss.dependency 'OLMKit', '~> 3.2.5'
       ss.dependency 'Realm', '10.27.0'
       ss.dependency 'libbase58', '~> 0.1.4'
-      ss.dependency 'MatrixSDK/CryptoSDK'
+      ss.dependency 'MatrixSDKCrypto', '0.2.0', :configurations => ["DEBUG", "RELEASE"], :inhibit_warnings => true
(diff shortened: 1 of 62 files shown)


MatrixSDK.podspec
```diff
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
       ss.dependency 'OLMKit', '~> 3.2.5'
       ss.dependency 'Realm', '10.27.0'
       ss.dependency 'libbase58', '~> 0.1.4'
-      ss.dependency 'MatrixSDK/CryptoSDK'
+      ss.dependency 'MatrixSDKCrypto', '0.2.0', :configurations => ["DEBUG", "RELEASE"], :inhibit_warnings => true
```
Anderas, line 48: I dont know if this is a coocapods bug, but I have to keep the `configuration` section in (even though it includes all configurations) if I also have the `inhibit_warnings` in

pixlwave: Odd, but if that's how it has to be its not going to bother anyone :)

pixlwave (approved):
Hurray congrats Andy 🍾

Looks great to me!

PR #1710, opened 2023-02-03
Fixed some scenarios where an answered call keeps ringing.

nimau:
This PR fixes some scenarios where a call continues to ring, even after being answered (or declined) elsewhere.

Related issues:
[iOS: Call answered elsewhere continues to ring](https://gitlab.matrix.org/ps/aarenet/issue-tracker/-/issues/450)
[App continues to ring #2](https://element-io.atlassian.net/browse/PSC-203)

The main issue here is that the expiration timer may be fired after 1 minute for a call that is already ended (because it has been answered elsewhere). In this case, setting its state to ended can pause the MXSession and prevents the current ringing call to receive any further events.

Here is the list of changes made to fix this issue:
- Invalidate the expiration timer once the call has been answered or declined elsewhere
- Avoid updating the state of an ended call:


aringenbach (approved):
LGTM

PR #1712, opened 2023-02-08
Fix some crashes

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/7338

Fix a bunch of unrelated crashes as reported by Sentry and Xcode

### DIFF
--- a/MatrixSDK/Crypto/CryptoMachine/MXCryptoRequests.swift
+++ b/MatrixSDK/Crypto/CryptoMachine/MXCryptoRequests.swift
@@ -40,7 +40,7 @@ struct MXCryptoRequests {
     }
     
     func sendToDevice(request: ToDeviceRequest) async throws {
-        return try await performCallbackRequest {
+        try await performCallbackRequest {
             restClient.sendDirectToDevice(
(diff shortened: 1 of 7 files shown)


MatrixSDK/Crypto/CryptoMachine/MXCryptoRequests.swift
```diff
@@ -106,15 +106,22 @@ struct MXCryptoRequests {
         }
     }
     
+    @MainActor
+    // Calling methods on `MXRoom` has various state side effects so should be called on the main thread
     func roomMessage(request: RoomMessageRequest) async throws -> String? {
-        var event: MXEvent?
-        return try await performCallbackRequest {
+        return try await withCheckedThrowingContinuation { cont in
```
alfogrillo, line 112:
```suggestion
        try await withCheckedThrowingContinuation { continuation in
```
**Optional**
- remove return when it's not needed
- avoid abbreviations like `cont`

alfogrillo (approved):
LGTM! 👍🏻
I left just an optional comment on code style.

PR #1713, opened 2023-02-08
Add fallback text for m.poll.end events (PSG-1156)

alfogrillo:
### Description

This PR adds a fallback text for `m.poll.end` events.
These fallbacks are useful for receivers when the associated `m.poll.start` event is unavailable.

### DIFF
--- a/MatrixSDK/Data/MXSendReplyEventDefaultStringLocalizer.m
+++ b/MatrixSDK/Data/MXSendReplyEventDefaultStringLocalizer.m
@@ -26,7 +26,7 @@
 @property (nonatomic, strong) NSString *senderSentTheirLocation;
 @property (nonatomic, strong) NSString *senderSentTheirLiveLocation;
 @property (nonatomic, strong) NSString *messageToReplyToPrefix;
-@property (nonatomic, strong) NSString *replyToEndedPoll;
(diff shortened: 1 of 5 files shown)


MatrixSDK/Data/MXSendReplyEventDefaultStringLocalizer.m
```diff
@@ -44,7 +44,7 @@ - (instancetype)init
         _senderSentTheirLocation = @"has shared their location.";
         _senderSentTheirLiveLocation = @"Live location.";
         _messageToReplyToPrefix = @"In reply to";
-        _replyToEndedPoll = @"Ended poll";
+        _endedPollMessage = @"Ended poll";
```
Anderas, line 47: These strings do not live in the regular localized strings file?

Anderas (approved):

MatrixSDK/Data/MXSendReplyEventDefaultStringLocalizer.m
alfogrillo: Yes, generally clients can override them if they wish.
In this context they are meant to be **not** localized since we don't want to localize stuff from the sender side when sending an event.

PR #1714, opened 2023-02-09
🟢 Apply (polls) push rules client side for encrypted rooms

flescio:
push rules for an event was broken because it was reading the encrypted one.
Now if present the clear event, the proper rule id is matching properly.

It was also missing the pushRulesManager's handle account data in the updateBackgroundServiceStoresIfNeeded, that way it was not synchronised and a reboot of the app was needed in order to fetch the proper push rules

### DIFF
--- a/MatrixSDK/Background/MXBackgroundPushRulesManager.swift
+++ b/MatrixSDK/Background/MXBackgroundPushRulesManager.swift
@@ -101,9 +101,6 @@ import Foundation
     public func pushRule(matching event: MXEvent,
                          roomState: MXRoomState,
                          currentUserDisplayName: String?) -> MXPushRule? {
-        // getting the uncrypted event if present or fallback
-        let event = event.clear ?? event
-        
         //  return nil if current user's event
         if event.sender == credentials.userId {
             return nil
@@ -118,8 +115,8 @@ import Foundation
             .roomMemberCount: memberCountConditionChecker,
             .senderNotificationPermission: permissionConditionChecker
         ]
-        
-        let eventDictionary = event.jsonDictionary()
+        // getting the unencrypted event if present or fallback
+        let eventDictionary = (event.clear ?? event).jsonDictionary()
         let equivalentCondition = MXPushRuleCondition()
         
         for rule in flatRules.filter({ $0.enabled }) {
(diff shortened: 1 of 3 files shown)


MatrixSDK/Background/MXBackgroundPushRulesManager.swift
```diff
@@ -101,6 +101,9 @@ import Foundation
     public func pushRule(matching event: MXEvent,
                          roomState: MXRoomState,
                          currentUserDisplayName: String?) -> MXPushRule? {
+        // getting the uncrypted event if present or fallback
```
aringenbach, line 104:
```suggestion
        // getting the unencrypted event if present or fallback
```

aringenbach (approved):
LGTM

MatrixSDK/Background/MXBackgroundSyncService.swift
```diff
@@ -500,5 +500,11 @@ public enum MXBackgroundSyncServiceError: Error {
             MXLog.debug("[MXBackgroundSyncService] updateBackgroundServiceStoresIfNeeded: Reset MXBackgroundCryptoStore")
             crypto.reset()
         }
+        
+        if let accountData = syncResponseStoreManager.syncResponseStore.accountData {
+            pushRulesManager.handleAccountData(accountData)
+        } else if let accountData = store.userAccountData ?? nil {
+            pushRulesManager.handleAccountData(accountData)
+        }
```
alfogrillo, line 508: This is copy and paste from the init.
Please create a new method like this `func updatePushRules` and call it from both.

alfogrillo: Actually this doesn't look like the right spot to put this code.
The method where you are adding this code in is called `func updateBackgroundServiceStoresIfNeeded()` which isn't what we are doing here
.
Probably a better spot would be here:
```swift
    private func fetchAndProcessEvent(withEventId eventId: String, roomId: String) {
        MXLog.debug("[NotificationService] fetchAndProcessEvent")
                
        NotificationService.backgroundSyncService.event(withEventId: eventId, inRoom: roomId) { [weak self] (response) in
            switch response {
            case .success(let event):
                MXLog.debug("[NotificationService] fetchAndProcessEvent: Event fetched successfully")
                self?.checkPlaybackAndContinueProcessing(event, roomId: roomId)
            case .failure(let error):
                MXLog.error("[NotificationService] fetchAndProcessEvent: Failed fetching notification event", context: error)
                self?.fallbackToBestAttemptContent(forEventId: eventId)
            }
        }
    }
```
in the `.success` case just before `checkPlaybackAndContinueProcessing(event, roomId: roomId)`
Wdyt?

alfogrillo (approved):
LGTM!

PR #1715, opened 2023-02-09
Typealias MXResponse to Swift.Result

alfogrillo:
### Description

This PR typealias `MXReponse` to `Swift.Result`.

This have a couple of benefit:
- `MXReponse` gets the `Swift.Result` api for free
- We may use an `MXResponse` wherever a `Swift.Result` is required without additional boilerplate (for example [here](https://developer.apple.com/documentation/swift/checkedcontinuation/resume(with:)-3gh60)).


Velin92 (approved):
LGTM

PR #1716, opened 2023-02-13
Fix: don’t send a verification cancel request if the session is closed

nimau:
This PR fixes:
- a crash in DEBUG when the app tries to send a cancel verification request while the session is closed.
- 2 retain cycles preventing MXCrypto from being properly released

### DIFF
--- a/MatrixSDK/Crypto/KeyBackup/Engine/MXNativeKeyBackupEngine.m
+++ b/MatrixSDK/Crypto/KeyBackup/Engine/MXNativeKeyBackupEngine.m
@@ -101,7 +101,9 @@ static Class DefaultAlgorithmClass;
     self.crypto.store.backupVersion = keyBackupVersion.version;
     Class algorithmClass = AlgorithmClassesByName[keyBackupVersion.algorithm];
     //  store the desired backup algorithm
+    MXWeakify(self);
     self.keyBackupAlgorithm = [[algorithmClass alloc] initWithCrypto:self.crypto authData:authData keyGetterBlock:^NSData * _Nullable{
+        MXStrongifyAndReturnValueIfNil(self, nil);
         return self.privateKey;
     }];
     MXLogDebug(@"[MXNativeKeyBackupEngine] enableBackupWithVersion: Algorithm set to: %@", self.keyBackupAlgorithm);
--- a/MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
+++ b/MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
@@ -699,6 +699,17 @@ static NSArray<MXEventTypeString> *kMXKeyVerificationManagerVerificationEventTyp
                           success:(void(^)(void))success
                           failure:(void(^)(NSError *error))failure
 {
+    // If the session is closed, don't try to send a request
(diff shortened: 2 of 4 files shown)


MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
```diff
@@ -707,6 +707,15 @@ - (void)cancelVerificationRequest:(id<MXKeyVerificationRequest>)request
     {
         [self cancelTransaction:transaction code:cancelCode success:success failure:failure];
     }
+    // If the session is closed, don't try to send a request
+    else if (self.crypto.mxSession.state == MXSessionStateClosed)
```
Anderas, line 711: Does this actually fix the problem? What if the first `if` branch is evaluated first, we never get to check session being closed. Maybe better to put this at the very start of the `cancelVerificationRequest`?

I would also be curious to know why `cancelVerificationRequest` is called in the first place when session is already closed, and perhaps the `isClosed` check really belongs in the caller of the `cancelVerificationRequest`. This is because if "cancel" can cause troubles, than so could the other methods in this manager, such as `cancelTransaction` or `sendToOtherInRequest`. Do you have reliable steps to reproduce this crash?

MatrixSDK/Crypto/KeyBackup/Engine/MXNativeKeyBackupEngine.m
```diff
@@ -101,7 +101,9 @@ - (BOOL)enableBackupWithKeyBackupVersion:(MXKeyBackupVersion *)keyBackupVersion
     self.crypto.store.backupVersion = keyBackupVersion.version;
     Class algorithmClass = AlgorithmClassesByName[keyBackupVersion.algorithm];
     //  store the desired backup algorithm
+    MXWeakify(self);
```
Anderas, line 104: Nice find on the retain cycles

MatrixSDK/Crypto/Verification/MXKeyVerificationManager.m
Anderas: The thing is that when the session is closed (say via the `close`), this will also close `session.crypto`, which will in turn close `crypto.keyVerificationManager`, even though by nillifying it. In other words when a session is closed no other objects should continue to function, and in the case of `MXKeyVerificationManager` it should cancel its timeout once closed. The reason this does not work at the moment is that even if `crypto` nillifies the manager, some other object keeps it in memory (not necessarily as a memory leak).

Whilst your solution does solve this particular problem, I fear this is merely one of many symptoms of the manager continuing its work after a session is closed. I would instead propose this solution:
- add new `close` method in the `MXLegacyKeyVerificationManager` header file
- implement the `close` method by clearing any timeouts that we have scheduled
- call the method just before `self->_keyVerificationManager = nil` is called from `MXLegacyCrypto close`

nimau: @Anderas, Ok, thanks. The verification manager was retained by its timers. So I removed my previous "fix", change the way we create the timers to prevent them to retain the manager, and finally, properly invalidate the timer on dealloc. It should be better and safer now.

Anderas (approved):
Thats great, thanks for the change

PR #1718, opened 2023-02-13
Encrypt rooms' last messages

alfogrillo:
### Description

This PR adds the AES encryption for rooms' last messages before they get stored into Core Data.

Fixes: https://github.com/vector-im/element-ios/issues/7358

| Before | After | 
| --- | ---|
| b|a|

### DIFF
--- a/MatrixSDK/Data/MXRoomLastMessage.h
+++ b/MatrixSDK/Data/MXRoomLastMessage.h
@@ -69,7 +69,14 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 
 - (instancetype)initWithEvent:(MXEvent *)event;
 
-- (nullable NSData*)encryptedAttributedString;
+/**
+ Returns an archived (possibly encrypted) version of MXRoomLastMessage data.
+ These include:
+ - `text`
+ - `attributedText`
+ - `others`
+ */
+- (nullable NSData*)sensitiveData;
 
 #pragma mark - CoreData Model
 
--- a/MatrixSDK/Data/MXRoomLastMessage.h
+++ b/MatrixSDK/Data/MXRoomLastMessage.h
@@ -70,7 +70,7 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 - (instancetype)initWithEvent:(MXEvent *)event;
 
 /**
- Returns an archived (possibly encrypted) version of MXRoomLastMessage sensible data.
+ Returns an archived (possibly encrypted) version of MXRoomLastMessage sensitive data.
  These include:
  - `text`
  - `attributedText`
--- a/MatrixSDK/Data/MXRoomLastMessage.h
+++ b/MatrixSDK/Data/MXRoomLastMessage.h
@@ -69,6 +69,8 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 
 - (instancetype)initWithEvent:(MXEvent *)event;
 
+- (nullable NSData*)encryptedAttributedString;
+
 #pragma mark - CoreData Model
 
 - (instancetype)initWithManagedObject:(MXRoomLastMessageMO *)model;
--- a/MatrixSDK/Data/MXRoomLastMessage.h
+++ b/MatrixSDK/Data/MXRoomLastMessage.h
@@ -70,7 +70,7 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 - (instancetype)initWithEvent:(MXEvent *)event;
 
 /**
- Returns an archived (possibly encrypted) version of MXRoomLastMessage data.
+ Returns an archived (possibly encrypted) version of MXRoomLastMessage sensible data.
  These include:
  - `text`
  - `attributedText`
--- a/MatrixSDK/Data/MXRoomLastMessage.m
+++ b/MatrixSDK/Data/MXRoomLastMessage.m
@@ -174,20 +174,7 @@ NSString *const kCodingKeyOthers = @"others";
     [coder encodeObject:_sender forKey:kCodingKeySender];
     
     // Build last message sensitive data
-    NSMutableDictionary *lastMessageDictionary = [NSMutableDictionary dictionary];
-    if (_text)
-    {
-        lastMessageDictionary[kCodingKeyText] = _text;
-    }
-    if (_attributedText)
-    {
-        lastMessageDictionary[kCodingKeyAttributedText] = _attributedText;
-    }
-    if (_others)
-    {
-        lastMessageDictionary[kCodingKeyOthers] = _others;
-    }
-    
+    NSDictionary *lastMessageDictionary = [self sensitiveData];
     // And encrypt it if necessary
     if (_isEncrypted)
     {
@@ -203,6 +190,26 @@ NSString *const kCodingKeyOthers = @"others";
     {
         [coder encodeObject:lastMessageDictionary forKey:kCodingKeyData];
     }
+}
+
+-(NSDictionary*)sensitiveData
+{
+    NSMutableDictionary *lastMessageDictionary = [NSMutableDictionary dictionary];
+    
+    if (_text)
+    {
+        lastMessageDictionary[kCodingKeyText] = _text;
+    }
+    if (_attributedText)
+    {
+        lastMessageDictionary[kCodingKeyAttributedText] = _attributedText;
+    }
+    if (_others)
+    {
+        lastMessageDictionary[kCodingKeyOthers] = _others;
+    }
+    
+    return  lastMessageDictionary;
 }
 
 #pragma mark - Data encryption
--- a/MatrixSDK/Data/MXRoomLastMessage.m
+++ b/MatrixSDK/Data/MXRoomLastMessage.m
@@ -61,10 +61,10 @@ NSString *const kCodingKeyOthers = @"others";
 
 - (nullable NSData*)encryptedAttributedString
 {
-    if (_attributedText)
+    if (self.attributedText)
     {
-        NSData* archieved = [NSKeyedArchiver archivedDataWithRootObject:_attributedText];
-        return [self encrypt:archieved];
+        NSData* archived = [NSKeyedArchiver archivedDataWithRootObject:self.attributedText requiringSecureCoding:NO error:nil];
+        return [self encrypt:archived];
     }
     else
     {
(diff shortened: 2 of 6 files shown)


MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -58,6 +59,19 @@ - (instancetype)initWithEvent:(MXEvent *)event
     return self;
 }
 
+- (nullable NSData*)encryptedAttributedString
+{
+    if (_attributedText)
```
Anderas, line 64: If not within a `init` method it is best to use synthesized accessors instad of ivars, i.e. `self.attributedText`

MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -58,6 +59,19 @@ - (instancetype)initWithEvent:(MXEvent *)event
...
+    {
+        NSData* archieved = [NSKeyedArchiver archivedDataWithRootObject:_attributedText];
```
Anderas, line 66: Typo `archieved` => `archived`.

Also `archivedDataWithRootObject` is now deprecated, so best to use the recommended variant with explicit error handling.

MatrixSDK/Data/RoomSummaryStore/CoreData/Models/MXRoomLastMessageMO.swift
```diff
@@ -47,19 +47,15 @@ public class MXRoomLastMessageMO: NSManagedObject {
...
-        }
+        s_text = lastMessage.isEncrypted ? nil : lastMessage.text
         
-        if let others = lastMessage.others {
-            s_others = NSKeyedArchiver.archivedData(withRootObject: others)
+        if lastMessage.isEncrypted {
+            s_attributedText = lastMessage.encryptedAttributedString()
         } else {
-            s_others = nil
+            s_attributedText = lastMessage.attributedText.map(NSKeyedArchiver.archivedData(withRootObject:))
```
Anderas, line 56: I think its a bit confusing that archiving of encrypted data happens in `MXRoomLastMessage` but unencrypted in `MXRoomLastMessageMO` ... it creates some assumptions about responsibilities that are not obvious from the API.

An alternative approach would be to simply expose `archivedAttributedString` getter on `MXRoomLastMessage` which would be used by `MXRoomLastMessageMO` directly. That way the latter will not need to perform any additional encryption checks

MatrixSDK/Data/MXRoomLastMessage.h
```diff
@@ -69,6 +69,8 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 
 - (instancetype)initWithEvent:(MXEvent *)event;
 
+- (nullable NSData*)encryptedAttributedString;
```
Anderas, line 72: As suggested below, I would rename this `archivedAttributedText` or `attributedTextData` and let it handle both encrypted and unencrypted variant, otherwise we are handling the archiving in two separate objects, depending on the encryption status.

Also I'd convert this into `property (nonatomic, readonly, nullable) NSData *archivedAttributedText` and move just below `attributedText` to make the API more consistent and clearer

MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -87,11 +101,29 @@ - (instancetype)initWithManagedObject:(MXRoomLastMessageMO *)model
...
+        }
+        
+        if (attributedTextData)
+        {
+            _attributedText = [NSKeyedUnarchiver unarchiveObjectWithData:attributedTextData];
+        }
+        
+        if (_attributedText == nil && model.s_text)
+        {
+            _attributedText = [[NSAttributedString alloc] initWithString:model.s_text];
```
Anderas, line 122: Does this not change the existing behaviour? Previously if `_attributedText` was unset, it would not be created from `model.s_text`.  It is not obvious to me whether `text` and `attributedText` are always variants of the same data or not. And if we always end up using `attributedText` whether it would be simpler to remove `text` entirely

MatrixSDK/Data/MXRoomLastMessage.m
alfogrillo: Applied -> https://github.com/matrix-org/matrix-ios-sdk/pull/1718/commits/1e1c24323531ced30bc9d7bfab5e3b1b67abb822

MatrixSDK/Data/MXRoomLastMessage.h
```diff
@@ -69,6 +69,15 @@ FOUNDATION_EXPORT NSString *const MXRoomLastMessageDataType;
 
 - (instancetype)initWithEvent:(MXEvent *)event;
 
+/**
+ Returns an archived (possibly encrypted) version of MXRoomLastMessage sensible data.
```
pixlwave, line 73:
```suggestion
 Returns an archived (possibly encrypted) version of MXRoomLastMessage sensitive data.
```

MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -58,6 +59,22 @@ - (instancetype)initWithEvent:(MXEvent *)event
     return self;
 }
 
+- (nullable NSData*)sensitiveData;
+{
+    NSData* archived = [NSKeyedArchiver archivedDataWithRootObject:[self sensitiveDataDictionary]
+                                             requiringSecureCoding:NO
+                                                             error:nil];
```
pixlwave, line 66: Is it worth logging any potential archiving errors that occur?

MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -87,16 +104,39 @@ - (instancetype)initWithManagedObject:(MXRoomLastMessageMO *)model
...
+        
+        if (archivedSensitiveData)
+        {
+            NSDictionary* sensitiveDataDictionary = [NSKeyedUnarchiver unarchiveObjectWithData:archivedSensitiveData];
+            
+            _text = sensitiveDataDictionary[kCodingKeyText];
+            _attributedText = sensitiveDataDictionary[kCodingKeyAttributedText];
+            _others = sensitiveDataDictionary[kCodingKeyOthers];
+        }
+        else // fallback logic for old database versions
```
pixlwave, line 126: You can completely avoid having to deal with old versions by incrementing `kMXFileVersion` in `MXFileStore`. It'll wipe the data and force a fresh sync when the user upgrades. Probably better than running the old way in parallel and avoids having to think about migrations etc.

pixlwave (commented):
I think that changing the file store version would probably add enough changes that I'll leave finishing the review until then.

Also, I'm not super keen on the `sensitiveData` naming tbh - I think being more explicit and saying `encryptedData` when its encrypted and simply `lastMessageData`/`Dictionary` when its unencrypted would be clearer here.

pixlwave:
I've just twigged about the naming as `sensitiveData` is only encrypted for rooms with encryption enabled. Probably just `lastMessageData` or `messageData` to be clearer, otherwise ignore that comment 🙃

MatrixSDK/Data/MXRoomLastMessage.m
alfogrillo: Technically I didn't change the `MXFileStore` so I'm not sure this is clean.
Btw does it mean people will need to do a full sync potentially taking a lot of time (for account big as Matthew's one)?

MatrixSDK/Data/RoomSummaryStore/CoreData/MXCoreDataRoomSummaryStore.xcdatamodeld/MXRoomSummaryCoreDataStore2.xcdatamodel/contents
```diff
@@ -62,10 +63,4 @@
         <attribute name="s_usersCount" attributeType="Integer 32" defaultValueString="0" usesScalarValueType="YES"/>
         <relationship name="s_ofRoom" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="MXRoomSummary" inverseName="s_trust" inverseEntity="MXRoomSummary"/>
     </entity>
-    <elements>
-        <element name="MXRoomLastMessage" positionX="117" positionY="90" width="128" height="149"/>
-        <element name="MXRoomMembersCount" positionX="297.1484375" positionY="-66.14453125" width="128" height="89"/>
-        <element name="MXRoomSummary" positionX="-63" positionY="-18" width="128" height="509"/>
-        <element name="MXUsersTrustLevelSummary" positionX="236.4921875" positionY="519.4296875" width="128" height="104"/>
-    </elements>
```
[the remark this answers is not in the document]
alfogrillo: I think this has been removed automatically by Xcode since they removed the [Core Data graph view](https://stackoverflow.com/questions/72824184/where-does-xcodes-core-data-graph-goes)! 😭

MatrixSDK/Data/MXRoomLastMessage.m
alfogrillo: Archiving a dictionary is quite safe, but added it anyway!
https://github.com/matrix-org/matrix-ios-sdk/pull/1718/commits/e6a0aee351bf4d30363767b5c7c81cdebfdbfa09

MatrixSDK/Data/RoomSummaryStore/CoreData/MXCoreDataRoomSummaryStore.xcdatamodeld/MXRoomSummaryCoreDataStore2.xcdatamodel/contents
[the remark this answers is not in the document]
pixlwave: Nooooo, why did they do that?! 😢

pixlwave (approved):
LGTM, chatted about the file store versions offline.

MatrixSDK/Data/MXRoomLastMessage.m
```diff
@@ -58,6 +59,27 @@ - (instancetype)initWithEvent:(MXEvent *)event
...
 
+- (nullable NSData*)sensitiveData;
+{
+    NSError* error;
+    NSData* archived = [NSKeyedArchiver archivedDataWithRootObject:[self sensitiveDataDictionary]
+                                             requiringSecureCoding:NO
+                                                             error:&error];
+    
+    if (error) {
+        MXLogDebug(@"[MXRoomLastMessage] did fail to archive sensitiveDataDictionary. Error: %@", error.description);
```
Anderas, line 70: For future reference: errors are best logged via `MXLogErrorDetails` which will also send them to Sentry (if enabled), or even `MXLogFailureDetails` if we do not expect this error to ever occur

Anderas (commented):
This looks so much clearer now 👍

PR #1719, opened 2023-02-15
Control CryptoSDK via feature flag

Anderas:
Create `MXCryptoV2Feature` protocol, implemented in the main app, which controls whether rust-based crypto sdk is enabled or not.

This allows more nuanced approach to enabling the feature (incl. at runtime), for instance by combining manual opt-in and remote feature flags.


nimau (approved):
LGTM

PR #1721, opened 2023-02-16
Fix an issue where the media download would not start until the end of the scroll

nimau:
This PR fixes an issue with the back pagination where images would not download until the end of the scroll.


stefanceriu (approved):
👍

PR #1724, opened 2023-02-24
Add more crypto logs

Anderas:
Based on rageshake investigations, curb existing logs, but adding more detail to some, and removing other


pixlwave (approved):
LGTM

PR #1726, opened 2023-02-27
Use correct next users with keys query

Anderas:
Fix a minor bug where starting the next pending `keys/query` would use the wrong list of users used to deduplicate requests. The consequence of this bug is that we may be executing slightly more requests than necessary


MatrixSDK/Crypto/CryptoMachine/MXKeysQueryScheduler.swift
```diff
@@ -94,15 +92,18 @@ public actor MXKeysQueryScheduler<Response> {
 
     private func executeQuery(users: Set<String>) async throws -> Response {
         defer {
-            if let next = nextTask {
-                log("... query completed, starting next pending query.")
-                currentQuery = .init(users: users, task: next)
+            if let nextTask = nextTask {
+                log("... query for \(users) completed, starting next pending query.")
+                currentQuery = .init(users: nextUsers, task: nextTask)
```
Velin92, line 97: maybe is a personal thing, but I don't like inferred constructor, could we use the constructor directly?

Velin92 (approved):
LGTM just one comment

PR #1727, opened 2023-02-27
Direct room with a third party

nimau:
This PR fixes part of [element-ios #6612](https://github.com/vector-im/element-ios/issues/6612)

When creating a direct chat with a 3rd party, use their email as the direct room id.
When updating the room summary, use the third party display name, if available, rather than empty room.

### DIFF
--- a/MatrixSDK/Data/MXRoomSummaryUpdater.m
+++ b/MatrixSDK/Data/MXRoomSummaryUpdater.m
@@ -564,8 +564,14 @@
     switch (memberNames.count)
     {
         case 0:
+        {
             displayname = _roomNameStringLocalizer.emptyRoom;
-            if (roomState.thirdPartyInvites.firstObject.displayname != nil)
+            NSString *directUserId = [session roomWithRoomId: roomState.roomId].directUserId;
+            if (directUserId != nil && [MXTools isEmailAddress:directUserId])
+            {
+                displayname = directUserId;
+            }
+            else if (roomState.thirdPartyInvites.firstObject.displayname != nil)
             {
                 displayname = roomState.thirdPartyInvites.firstObject.displayname;
             }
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
looks good to me 👍

MatrixSDK/Data/MXRoomSummaryUpdater.m
```diff
@@ -516,8 +516,7 @@ - (BOOL)updateSummaryDisplayname:(MXRoomSummary *)summary session:(MXSession *)s
             }
         }
 
-        if (memberCount > 1
-            && (!displayName || [displayName isEqualToString:_roomNameStringLocalizer.emptyRoom]))
+        if (!displayName || [displayName isEqualToString:_roomNameStringLocalizer.emptyRoom])
```
giomfo, line 519: I think we should keep here a sanity check on memberCount
`memberCount > 0` 
because of the computation based on this count in `fixUnexpectedEmptyRoomDisplayname`

giomfo (approved):
SGTM, only one remark

MatrixSDK/Data/MXRoomSummaryUpdater.m
nimau: @giomfo in the particular case, `memberCount` is initialized to 0 and updated only if there is a `serverRoomSummary` (which is not the case) or if `roomState.membersCount.members > 1` (which is also not the case).
So the value of `memberCount` will be 0, which is not a problem for the `fixUnexpectedEmptyRoomDisplayname` method because there are no other members in the room, so `memberNames.count` will be 0 and the fix will apply.

giomfo: @nimau I'm ok about the targeted use case. So remove this sanity check here is ok for me
But don't you think we would add some sanity check in `fixUnexpectedEmptyRoomDisplayname` where we reduce by 1 or 2 the provided `memberCount`?

PR #1728, opened 2023-03-02
Fix room list last message when the key comes late

manuroe:
Fix https://github.com/vector-im/element-ios/issues/6848.

**TODO:** add dependency on another incoming PR.

This bug fix handles the case when the key arrives after an app restart.

### DIFF
--- a/MatrixSDKTests/MXRoomListDataManagerTests.swift
+++ b/MatrixSDKTests/MXRoomListDataManagerTests.swift
@@ -337,6 +337,71 @@ class MXRoomListDataManagerTests: XCTestCase {
         }
     }
     
+    /// Test: Expect a last message to recover after being a UTD
+    /// - Create a scenario where Bob is getting a UTD from Alice
+    /// - Restart Bob session
+    /// - Make alice send the blocked to_device event (this simulates a late room key
+    /// - Expect Bob to be able to decrypt the last message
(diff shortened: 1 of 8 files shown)


MatrixSDKTests/MXRoomListDataManagerTests.swift
```diff
@@ -290,6 +447,78 @@ class MXRoomListDataManagerTests: XCTestCase {
...
+            
+            let fetcher = manager.fetcher(withOptions: self.basicFetchOptions)
+            fetcher.paginate()
+            completion(aliceSession, bobSession, fetcher, expectation)
+        }
+    }
+    
+    /// Create a scenario with an UTD (Unable To Decrypt) for the last message
+    /// - Parameter closure called when scenario has been created. Among all parameters, it provides the to_device message that has been blocked, creating the UTD.
+    private func createFetcherWithBobAndAliceWithUTD(_ completion: @escaping (MXSession, MXSession, MXRoomListDataFetcher, String, String, MXToDevicePayload, XCTestExpectation) -> Void) {
```
stefanceriu, line 471: Very nice 👌

stefanceriu (approved):
Not much to comment, the tests ar great and even though the retry mechanism isn't the most self-describing the comments make it clear

PR #1731, opened 2023-03-03
Fix crash when roomSummary is nil

Velin92:
This PR fixes the following crash: https://github.com/vector-im/element-ios/issues/7403


MatrixSDK/Data/RoomList/Common/MXBreadcrumbsRoomListDataFetcher.swift
```diff
@@ -91,7 +91,7 @@ internal class MXBreadcrumbsRoomListDataFetcher: NSObject, MXRoomListDataFetcher
         
         if let query = fetchOptions.filterOptions.query?.lowercased(), !query.isEmpty {
             recentRoomIds = recentRoomIds.filter({ roomId in
-                guard let displayname = session?.roomSummary(withRoomId: roomId).displayname else {
+                guard let displayname = session?.roomSummary(withRoomId: roomId)?.displayname else {
```
stefanceriu, line 94: Pet peve, but can we rename this to `displayName` please?

MatrixSDK/MXSession.h
```diff
@@ -1087,7 +1087,7 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
 
  @return the MXRoomSummary instance.
  */
-- (MXRoomSummary *)roomSummaryWithRoomId:(NSString*)roomId;
+- (nullable MXRoomSummary *)roomSummaryWithRoomId:(NSString*)roomId;
```
stefanceriu, line 1090: I'm pretty sure this nullability specifier will trigger a ton of the warnings for the rest of the methods in this file. Have a look at `MX_ASSUME_MISSING_NULLABILITY_BEGIN`

stefanceriu (approved):
Looks good to me except for the inline comments

MatrixSDK/MXSession.h
Velin92: Oh we don't have it in MXSession.h

PR #1733, opened 2023-03-08
Always update tracked users when sharing keys

Anderas:
Resolves https://github.com/matrix-org/element-web-rageshakes/issues/19974

There are some rare cases when room members are not correctly tracked in `OlmMachine` (investigating separately), which is typically triggered by `m.room.membership` events. To solve the consequence (rather than the root cause), ensure that we always update tracked users when sharing keys, even if this does not trigger keys query for the current event, but will do so before the next one. In other words if we do indeed have users that are not yet tracked, they will be unable to decrypt at most the current message, but should recieve the key for the next message.


nimau (approved):
LGTM

PR #1734, opened 2023-03-09
Improved session startup progress

Anderas:
Relates to https://github.com/vector-im/element-ios/issues/7417

Refactor `MXSessionStartupProgress` to align with current use case requirements, where we do not display sync attempts or copy for individual stages but rather calculate an overall progress.


aringenbach (approved):
LGTM

PR #1735, opened 2023-03-10
Fully deprecate MXCryptoStore

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/7419

Related rust-sdk change: https://github.com/matrix-org/matrix-rust-sdk/pull/1460

Until now the rust-based Crypto V2 still required access to the legacy Realm crypto database to store the following data:
- room algorithms
- whether to block untrusted devices per room
- whether to block untrusted devices globally

The rust-sdk can now store this data as well (after the above PR is merged) so we can fully deprecate the `MXCryptoStore` in the ios sdk.

Because there are already clients in production that have migrated from legacy to rust crypto we need to consider several use cases for migration of settings:
- new login => no realm store, so nothing to migrate
- user in legacy crypto => we can migrate settings via the existing migrate method
- user in rust crypto => we must only migrate room settings, but none of the other data (otherwise we could override changes in the account)

This is achieved by adding another `MXCryptoVersion` case and migrating in steps depending on the current version.

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
       ss.dependency 'OLMKit', '~> 3.2.5'
       ss.dependency 'Realm', '10.27.0'
       ss.dependency 'libbase58', '~> 0.1.4'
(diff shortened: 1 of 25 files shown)


Anderas:
Note that the PR will not build as i've not yet published new rust crypto framework (waiting to merge the associated PR)

flescio (approved):
LGTM!

PR #1738, opened 2023-03-13
Make CallKit maximumCallGroups configurable

manuroe:
This PR supersedes #1085  to cover call transfer about other use cases too.


pixlwave (approved):
LGTM 👍

PR #1739, opened 2023-03-14
MXCallManager: Make call transfer requests sequential

manuroe:
Make sure we made requests one by one to avoid to create weird scenarii.

### DIFF
--- a/MatrixSDK/VoIP/MXCallManager.m
+++ b/MatrixSDK/VoIP/MXCallManager.m
@@ -882,36 +882,37 @@ NSTimeInterval const kMXCallDirectRoomJoinTimeout = 30;
                         //  generate a new call id
                         NSString *newCallId = [[NSUUID UUID] UUIDString];
                         
-                        dispatch_group_t dispatchGroupReplaces = dispatch_group_create();
+                        dispatch_group_t dispatchGroup = dispatch_group_create();
+                        dispatch_group_enter(dispatchGroup);
                         
                         if (callWithTarget)
                         {
                             //  send replaces event to target
-                            dispatch_group_enter(dispatchGroupReplaces);
                             [callWithTarget transferToRoom:nil
                                                       user:transferee
                                                 createCall:nil
                                                  awaitCall:newCallId
                                                    success:^(NSString * _Nonnull eventId) {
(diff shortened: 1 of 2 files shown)


MatrixSDK/VoIP/MXCallManager.m
```diff
@@ -882,36 +882,37 @@ - (void)transferCall:(MXCall *)callWithTransferee
...
                         {
                             //  send replaces event to target
-                            dispatch_group_enter(dispatchGroupReplaces);
                             [callWithTarget transferToRoom:nil
                                                       user:transferee
                                                 createCall:nil
                                                  awaitCall:newCallId
                                                    success:^(NSString * _Nonnull eventId) {
-                                dispatch_group_leave(dispatchGroupReplaces);
+                                dispatch_group_leave(dispatchGroup);
```
alfogrillo, line 896: I think you also need to leave the dispatch group in case of `failure`

MatrixSDK/VoIP/MXCallManager.m
```diff
@@ -947,49 +948,50 @@ - (void)transferCall:(MXCall *)callWithTransferee
...
                                 
                                 //  send replaces event to target
-                                dispatch_group_enter(dispatchGroupReplaces);
                                 [callWithTarget transferToRoom:transferRoom.roomId
                                                           user:transferee
                                                     createCall:nil
                                                      awaitCall:newCallId
                                                        success:^(NSString * _Nonnull eventId) {
-                                    dispatch_group_leave(dispatchGroupReplaces);
+                                    dispatch_group_leave(dispatchGroup);
```
alfogrillo, line 964: Same problem here

MatrixSDK/VoIP/MXCallManager.m
alfogrillo: > Does it create an issue (like a retain cycle) if we let the dispatch_group open

This is what I meant. 
A dispatch group with a work count equal to "one" will stick somewhere forever I guess.
But I can't find easily a confirmation of this (DispatchGroup objects don't work as normal ARC objects).
Btw I don't see it as a huge problem, I'm going to approve it anyway.

alfogrillo (approved):

PR #1740, opened 2023-03-16
Simplify user verification state

Anderas:
Resolves https://github.com/matrix-org/internal-config/issues/1325

The SDK has previously relied on `MXUserTrustLevel` to determine whether any particular user is verified or not. It did so by distinguishing between "local" and "cross-signed" trust, where a user is considered verified if either one of those fields is true. This is however unnecessarily complex model, not used upcoming rust-crypto, and does not really make sense, given that "user verification" should only be possible via cross-signing, not by setting manual values.

To improve this, remove `MXUserTrustLevel` entirely and replace it with a simple `isVerified` boolean. This makes the API everywhere much simpler to use. As a downside, we have to deal with some amount of migration from the previous model (we use NSCoding to persist these values). To ensure correct migration, we will now store "version" property in MXCrossSigningInfo, which is used to decode the values correctly

Additionally replace `NSProgress` with `MXTrustSummary` when computing `trustSummary` for more than one user or device. Using `NSProgress` has worked okay, but it makes for some strange and clunky API (e.g. `.fractionCompleted >= 0` instead of `allAreTrusted`).


Velin92 (approved):
LGTM

PR #1743, opened 2023-03-21
Add event decryption decoration

Anderas:
Previously the legacy SDK would expose `untrusted` property to mark events that were decrypted using an untrusted key (e.g. one downloaded from asymmetrical backup). On the app side, this untrusted property would be combined with other encryption properties, namely the trust of the sender and their device. Only then the combine result gets turned into a visual decoration (e.g. red or grey shield).

To improve this, and align with rust-sdk, expose the decoration color directly from the crypto SDK, alongside an explanation message if necessary. This moves some of the trust computation down to the crypto module and leaves the app with fewer decisions to make.

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
       ss.dependency 'OLMKit', '~> 3.2.5'
       ss.dependency 'Realm', '10.27.0'
       ss.dependency 'libbase58', '~> 0.1.4'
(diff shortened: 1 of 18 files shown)


changelog.d/pr-1743.change
pixlwave, line 1: Should we have a `.api` changelog too that documents the removal of the `untrusted` property and its replacements for other users of the SDK?

pixlwave (approved):
LGTM 😎

changelog.d/pr-1743.change
Anderas: Yea that makes sense, renamed. thx

PR #1751, opened 2023-03-27
Upgrade verification if necessary

Anderas:
Set `needsVerificationUpgrade` property when migrating from legacy to rust crypto, which will be used to display altered UX for "verify your session" modal. This is because sessions considered as verified in the legacy crypto should not always be considered as verified in rust crypto, as the latter imposes stricter rules on session verification. In case the rust *does* consider the session as verified the property is reset.

Note that this is not meant to be 100% reliable way to detect verification state change. For instance if legacy crypto was already unverified, migrating to rust would still display the "upgrade verification" dialoge. This is acceptable, given that the "upgrade verification" prompt will only be shown until either successful verification or ignoring of the alert.

Additionally keep the existing realm store around after migration instead of deleting it. It will be removed in a future PR.


BillCarsonFr (approved):
LGTM

pixlwave (approved):
LGTM 👍

flescio (approved):
LGTM!

PR #1754, opened 2023-03-30
Upgrade to Jitsi pod to 7.0.1-lite

pixlwave:
I picked the Lite version as it excludes a dependency on Giphy which appears to be a 22MB binary for iOS and I'm not sure we really need that.

Closes https://github.com/vector-im/customer-retainer/issues/28 although discussed with @benparsons that this on its own isn't enough to enable screensharing. Extra info should be in Jira.


alfogrillo (approved):
LGTM

PR #1756, opened 2023-04-03
Resolve QR warnings

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/7470

We log several errors to Sentry regarding QR codes which should not really be errors, namely the inability to generate QR code data when the other side cannot scan (but we can scan theirs).

Improve this by creating a dedicated enum to capture these two variants as opposed to an optional that is more ambiguous


stefanceriu (approved):
Nice! 👏

PR #1763, opened 2023-04-11
Delete data for mismatched accounts

Anderas:
Resolves https://github.com/vector-im/element-ios/issues/7488

When we recieve a "Mismatched account" error (meaning we have existing data for the current user but under a different device) we need to delete the previous data first before we can try to create the machine.

Note that this behaviour should not be happening as we delete data on logouts, but there are clearly edge-cases where this still occurs. In those cases we will log the error.

### DIFF
--- a/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
+++ b/MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
@@ -36,6 +36,8 @@ class MXCryptoMachine {
     }
     
     private static let kdfRounds: Int32 = 500_000
+    // Error type will be moved to rust sdk
+    private static let MismatchedAccountError = "the account in the store doesn't match the account in the constructor"
     
     enum Error: Swift.Error {
         case invalidEvent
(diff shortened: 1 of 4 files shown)


MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -36,6 +36,8 @@ class MXCryptoMachine {
     }
     
     private static let kdfRounds: Int32 = 500_000
+    // Error type will be moved to rust sdk
```
aringenbach, line 39: Maybe add this as a FIXME/TODO for visibility ?

aringenbach (approved):
LGTM !

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
Anderas: I am not a fan of adding [TODOs to code](https://www.osedea.com/en/blog/should-you-put-todos-in-the-source-code), but it is tracked in a draft issue on a sprint board so to be resolved asap

PR #1765, opened 2023-04-12
Fetch device info by sender key

Anderas:
When fetching device info, ensure that we retrieve it via its sender key, which may not be set on an event yet, but is a part of the decryption result.

Additionally upgrade crypto sdk version


flescio (approved):
LGTM!

pixlwave (approved):
LGTM 👍

PR #1768, opened 2023-04-14
Ensure device signed after restoring cross-signing keys

Anderas:
`MXRecoveryService` responsible for restoring services after the import of cross-signing keys (e.g. when passphrase is entered) should verify the current user and device and upload relevant signature. However an early exit only checking current user trust and not the device trust will prevent this from happening ("canCrossSign" does not imply "device is signed").

To resolve this, remove the early exit in `MXRecoveryService` and always call `setUserVerification` and `crossSignDeviceWithDeviceId`. To avoid duplicate verifications of an already trusted device, add an early exit into `crossSignDeviceWithDeviceId` instead.


alfogrillo (approved):
LGTM!

PR #1770, opened 2023-04-15
Enable Crypto SDK by default

Anderas:
After enabling rust crypto sdk for all users via [feature flag](https://github.com/vector-im/element-ios/pull/7485) we can now enable it in matrix sdk as well by default, without yet removing any of the legacy code.

This change has quite a significant impact on the existing test infrastructure, which has been running against legacy crypto all along. Majority of the integration tests assert against now-unavailable API (e.g. inspecting realm store) and thus cannot be run against crypto sdk. We have however plenty of new unit tests, as well as the external crypto sdk having its own tests.

 The following changes are made in this PR:
- remove `MXCryptoV2Feature`, default `enableCryptoSDK` to `YES`
- remove `CryptoTests` plan entirely and only depend on `AllWorkingTests`, there is no need to have two plans for integration tests
- heavily currate `AllWorkingTests` to only run tests that are actually passing. Unfortunatelly there are many flaky integration tests not related to crypto at all (e.g. MXRestClient) and it is better to run a few actually passing tests rather than many that often fail
- create a [task](https://github.com/vector-im/crypto-internal/issues/71) to start enabling crypto integration tests where possible
- add `LegacyRealmStore` and a few pre-created `realm` database files that can be used to test crypto migration even when its not possible to initialize legacy crypto anymore.

Note that this PR is a pre-requisite to start deleting some of the now unnecessary legacy crypto code.


manuroe (approved):
LGTM
Thanks for making integration tests CI happy again!

PR #1774, opened 2023-04-18
Setup cross-signing even without auth session

Anderas:
Calling `[MXRestClient authSessionForRequestWithMethod]` is supposed to return an auth session which is then passed to crypto machine to setup cross-singing keys. It is however valid for the auth session to be empty, which is already accounted for in `MXLegacyCrossSigning`, but not in `MXCrossSigningV2`. This PR fixes that issue by behaving in the same way as legacy cross-signing, namely returning empty auth session instead of throwing an error.


pixlwave (approved):
👍

PR #1776, opened 2023-04-20
Fix: Refreshing the poll when receiving pollEnd can break the chronological order in the store.

nimau:
This PR fixes [element-ios issue #7497](https://github.com/vector-im/element-ios/issues/7497)

When a .pollEnd event is received, if the .pollStart event is not in the store, the SDK retrieves all events related to the poll and saves them in the store, breaking its chronological order.

Furthermore, even with this refresh, the timeline does not seem to be updated after the events are injected into the store and the behaviour of the timeline seems better without it.


alfogrillo:
This potentially breaks the rendering of a `m.poll.end` event.
When a user open the timeline, if there are gaps in the local store and the related `m.poll.start` event isn't present in the local store, the `m.poll.end` event won't be rendered at all. 😕
Probably isn't super common but it's possible.
fyi @Johennes

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -149,14 +162,33 @@ public class PollAggregator {
...
             }
             
+            guard let event = response.originalEvent else {
+                // we didn't found the start event
+                self.delegate?.pollAggregator(self, didFailWithError: PollAggregatorError.invalidPollStartEvent)
+                return
+            }
+            
+            // if we don't already have a pollStartedEvent, update it
+            if pollStartedEvent == nil {
```
alfogrillo, line 180: What happens if we remove this if?

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -137,8 +135,23 @@ public class PollAggregator {
                                  events: events,
                                  currentUserIdentifier: session.myUserId,
                                  hasBeenEdited: hasBeenEdited)
+    }
+    
+    
+    /// Creates a temporary poll without having the start event yet
+    private func buildPollWithoutStartEvent() {
```
alfogrillo, line 142: It would be nice if we would be able to switch the `Poll` to an optional one as described [here](https://github.com/vector-im/element-ios/pull/7518#issuecomment-1523124000).
What do you think?

MatrixSDK/Room/Polls/PollAggregator.swift
nimau: Everything works fine. I removed it (it was no longer necessary, as the poll is now optional)

MatrixSDK/Room/Polls/PollAggregator.swift
nimau: Done. (see the related element-ios PR).

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -153,7 +149,23 @@ public class PollAggregator {
...
+                return
+            }
+            
+            if let pollStartEventContent = MXEventContentPollStart(fromJSON: event.content), pollStartEventContent.answerOptions.count >= Constants.minAnswerOptionCount {
+                pollStartedEvent = event
+                self.pollStartEventContent = pollStartEventContent
+                hasBeenEdited = (event.unsignedData.relations?.replace != nil)
+            } else {
+                // we were not able to decrypt the start event content
+                self.delegate?.pollAggregator(self, didFailWithError: PollAggregatorError.invalidPollStartEvent)
```
alfogrillo, line 168: I see a bit of overlap with the body of `buildPollStartContent`.
I think we can avoid duplication like this.

1. Write a reusable function like this

```swift
    private func tryUpdatePollStartedEvent(with event: MXEvent?) {
        guard
            let event = event,
            let eventContent = MXEventContentPollStart(fromJSON: event.content),
            eventContent.answerOptions.count >= Constants.minAnswerOptionCount
        else {
            delegate?.pollAggregator(self, didFailWithError: PollAggregatorError.invalidPollStartEvent)
            return
        }
        
        pollStartedEvent = event
        pollStartEventContent = eventContent
        hasBeenEdited = (event.unsignedData.relations?.replace != nil)
    }
```

2. Call it here and in `buildPollStartContent`
3. Finally I think we can also delete `throws` from `buildPollStartContent`.

Wdyt?

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -137,10 +135,8 @@ public class PollAggregator {
                                  events: events,
                                  currentUserIdentifier: session.myUserId,
                                  hasBeenEdited: hasBeenEdited)
-        
-        reloadPollData()
     }
-    
+
```
alfogrillo, line 139:
```suggestion

```

MatrixSDK/Room/Polls/PollAggregator.swift
```diff
@@ -100,9 +100,11 @@ public class PollAggregator {
         
         NotificationCenter.default.addObserver(self, selector: #selector(handleRoomDataFlush), name: .mxRoomDidFlushData, object: self.room)
         setupEditListener()
-        try buildPollStartContent()
+        try? buildPollStartContent()
+        
+        reloadPollData()
     }
-    
+
```
alfogrillo, line 107:
```suggestion

```

alfogrillo:
There also build errors in the CI. Probably something isn't right building with Xcode 14.2.

nimau:
@alfogrillo I've updated the PR to reflect your comments (and I've also fixed the CI and UnitTests issues). Let me know if it works for you.

alfogrillo (approved):
LGTM! 🎉

PR #1777, opened 2023-04-21
Update breadcrumb list when leaving room

JanNikGra:
### Previous behaviour:
After leaving a room, the breadcrumb list is not updated. A user can still click on the breadcrumb item and a blank screen with no information will be displayed. Even if the user removes a room using the breadcrumb item action menu, the room will not be removed from the breadcrumb list.

### Improved behaviour:
After leaving a room, the breadcrumb list will be updated in all active sessions.

### Pull Request Checklist

Signed-off-by: Jan Niklas Grabowski janniklas.grabowski.ext@bwi.de


nimau (changes requested):
Hi @JanNikGra,
That seems good to me, just one point:
Can we call setAccountData only if we have found the roomId in recentRoomIds?

nimau (approved):
LGTM.

PR #1784, opened 2023-05-18
Fix threads empty state: Allow empty array for chunk in MXAggregationPaginatedResponse

langleyd:
Fixes https://github.com/vector-im/element-ios/issues/7551

### DIFF
--- a/MatrixSDK/JSONModels/Aggregations/MXAggregationPaginatedResponse.m
+++ b/MatrixSDK/JSONModels/Aggregations/MXAggregationPaginatedResponse.m
@@ -35,12 +35,18 @@
     MXAggregationPaginatedResponse *paginatedResponse;
 
     NSArray<MXEvent*> *chunk;
-    MXJSONModelSetMXJSONModelArray(chunk, MXEvent.class, JSONDictionary[@"chunk"])
-
+    NSArray *chunkJson = JSONDictionary[@"chunk"];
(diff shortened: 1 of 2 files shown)


MatrixSDK/JSONModels/Aggregations/MXAggregationPaginatedResponse.m
```diff
@@ -35,12 +35,18 @@ + (instancetype)modelFromJSON:(NSDictionary *)JSONDictionary
     MXAggregationPaginatedResponse *paginatedResponse;
 
     NSArray<MXEvent*> *chunk;
-    MXJSONModelSetMXJSONModelArray(chunk, MXEvent.class, JSONDictionary[@"chunk"])
-
+    NSArray *chunkJson = JSONDictionary[@"chunk"];
+    MXJSONModelSetMXJSONModelArray(chunk, MXEvent.class, chunkJson)
+    // For some reason modelsFromJSON returns nil if you pass it an empty array.
```
langleyd, line 40: Really a more sensible behaviour for `modelsFromJSON` would be to return an empty array if you pass it an empty array. But probably too risky a change, considering this repo is mostly deprecated, so just special-cased it here.

stefanceriu (approved):
Looks sensible to me. I would normally say we need to check all usages but you're right, not worth the risk in this case.

PR #1789, opened 2023-05-30
Updating Crypto API to v3

### LINKED ISSUE #1788: Use v3 API for crypto endpoints
Currently we are using /_matrix/client/r0/keys/signatures/upload instead of the newer and correct /_matrix/client/v3/keys/signatures/upload route

Original issue:
https://github.com/vector-im/element-ios/issues/7557

Velin92:
https://github.com/vector-im/element-ios/issues/7557

### DIFF
--- a/MatrixSDK/MXEnumConstants.m
+++ b/MatrixSDK/MXEnumConstants.m
@@ -22,7 +22,7 @@
  Matrix content respository path
  */
 NSString *const kMXContentUriScheme  = @"mxc://";
-NSString *const kMXContentPrefixPath = @"_matrix/media/r0";
+NSString *const kMXContentPrefixPath = @"_matrix/media/v3";
 
 /**
  Prefix used in path of antivirus server API requests.
--- a/MatrixSDK/MXRestClient.m
+++ b/MatrixSDK/MXRestClient.m
@@ -4817,7 +4817,7 @@ andUnauthenticatedHandler: (MXRestClientUnauthenticatedHandler)unauthenticatedHa
                        success:(void (^)(MXKeysUploadResponse *keysUploadResponse))success
                        failure:(void (^)(NSError *error))failure
 {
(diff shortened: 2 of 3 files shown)


MatrixSDK/MXEnumConstants.m
```diff
@@ -22,7 +22,7 @@
  Matrix content respository path
  */
 NSString *const kMXContentUriScheme  = @"mxc://";
-NSString *const kMXContentPrefixPath = @"_matrix/media/r0";
+NSString *const kMXContentPrefixPath = @"_matrix/media/v3";
```
BillCarsonFr, line 25: It's not related to crypto?

BillCarsonFr (changes requested):
Looks like we should check `/versions` to know what version of the spec is supported by the homeserver then use the correct end point. 
Some details here https://github.com/matrix-org/matrix-spec-proposals/blob/old_master/proposals/2844-global-versioning.md ? not sure if safe to move all to v3

Velin92:
> Looks like we should check `/versions` to know what version of the spec is supported by the homeserver then use the correct end point. Some details here https://github.com/matrix-org/matrix-spec-proposals/blob/old_master/proposals/2844-global-versioning.md ? not sure if safe to move all to v3

Okay I understand, this PR was actually suggested by a member of the community, but maybe we should actually discard it and schedule a separate piece of work on this MSC that you linked.
What do you think? @manuroe

Velin92:
> The PR was originally suggested because the Conduit homeserver implementation had moved to the v3 endpoints, rendering all crypto-related functionality unusable on the Element client: [famedly/conduit#292](https://gitlab.com/famedly/conduit/-/issues/292).
> 
> FWIW, the changes I proposed were based on what FluffyChat did: [famedly/fluffychat#986](https://gitlab.com/famedly/fluffychat/-/issues/986). Of course, it’d be perfectly reasonable for Element to go with a different solution.
> 
> Whatever solution ends up getting adopted, I hope it makes Element clients work well with Conduit soon.

Hello thanks again for your help, we probably need to decide first from a product perspective if it's okay to leave behind any server that would not support the v3 endpoints or if we should instead have a way to support all of them based on the version check.

PR #1790, opened 2023-05-30
JItsi SDK update

Velin92:
to version 8.1.2-lite


pixlwave (approved):
LGTM, although it should definitely have a changelog to say we updated the Jitsi pod 👍

Velin92:
> LGTM, although it should definitely have a changelog to say we updated the Jitsi pod 👍

right! Just pushed it.

PR #1799, opened 2023-06-21
Ignore push rules with unknown condition kinds

Johennes:
Fixes: vector-im/element-ios#7601


pixlwave (approved):
Looks sane to me, and works testing it on my account.

PR #1801, opened 2023-07-03
Fix fallback parsing logic for legacy location events

alfogrillo:
This PR fixes a problem causing the fallback parsing logic not working properly for "legacy" location events like this:

```
{
   "type": "m.room.message",
   "content": {
       "body: "Location was shared at geo:42.3431550455947,13.3939963489709 as of 2023-07-03 11:15:01 +0000",
       "geo_uri": "geo:42.3431550455947,13.3939963489709",
       "msgtype": "m.location"
    }
}
```
More context [here](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/location/proposals/3488-location.md)

| **Before** | **After** |
| --- | --- |
| b|a|

### DIFF
--- a/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
+++ b/MatrixSDK/JSONModels/Event/Content/MXEventContentLocation.m
@@ -111,7 +111,8 @@
         }
         
         finalLocationDictionary = @{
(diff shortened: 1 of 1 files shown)


aringenbach (approved):
LGTM

PR #1805, opened 2023-07-20
Matrix sdk crypto update

stefanceriu:
This PR bumps the Matrix Crypto SDK version to 0.3.11 and fixes newly introduced problems:
* breaking api changes following uniffi udl->macro replacement
* toDevice sync response event processing after JSON format change + improved rust side API to avoid it in the future
* broken secret recovery following USK and SSK enum associtated value order change

Successful integration test run here https://github.com/matrix-org/matrix-ios-sdk/actions/runs/5609719049/jobs/10263800231


pixlwave (approved):
LGTM

PR #1807, opened 2023-07-28
Device dehydration v2

stefanceriu:
- add support for device dehydration v2 i.e. `org.matrix.msc3814`
- implement rest client methods and models
- implement flows for creating and storing the dehydrated device pickle in in 4s as well as recovering it to use for device rehydration
- implement methods for creating, retrieving, deleting dehydrated devices as well as consuming the messages stored on the backend in between sessions
- delete previous implementation

NB: this requires a new Crypto SDK releases, current implementation based on https://github.com/matrix-org/matrix-rust-sdk/tree/poljar/device-dehydration


MatrixSDK/Crypto/MXCryptoV2.swift
```diff
@@ -143,6 +143,10 @@ class MXCryptoV2: NSObject, MXCrypto {
             delegate: crossSign
         )
         
+        dehydrationService = DehydrationService(restClient: restClient,
+                                                  secretStorage: secretStorage,
+                                                  dehydratedDevices: machine.dehydratedDevices())
```
pixlwave, line 148:
```suggestion
        dehydrationService = DehydrationService(restClient: restClient,
                                                secretStorage: secretStorage,
                                                dehydratedDevices: machine.dehydratedDevices())
```

MatrixSDK/MXRestClient.h
pixlwave, line 1: Could we maybe keep/recreate the documentation in this file?

MatrixSDK/Crypto/Dehydration/DehydrationService.swift
```diff
@@ -0,0 +1,214 @@
...
+    case failedRetrievingSecret(Error)
+    case failedRetrievingPrivateKey(Error)
+    case invalidSecretStorageDefaultKeyId
+    case failedRetrievingToDeviceEvents(Error)
+    case failedDeletingDehydratedDevice(Error)
+}
+
+@objcMembers
+public class DehydrationService: NSObject {
+    let deviceDisplayName = "Backup Device"
```
pixlwave, line 35: Maybe this should be configurable through the SDK settings, for e.g. localisation.

MatrixSDK/Crypto/Dehydration/DehydrationService.swift
```diff
@@ -0,0 +1,214 @@
...
+            let rehydrationResult = await rehydrateDevice(pickleKeyData: [UInt8](pickleKeyData))
+            switch rehydrationResult {
+            case .success((let deviceId, let rehydratedDevice)):
+                // Fetch and process the to device events available on the dehydrated device
+                try await processToDeviceEvents(rehydratedDevice: rehydratedDevice, deviceId: deviceId)
+                
+                // And attempt to delete the dehydrated device but ignore failures
+                try? await deleteDehydratedDevice(deviceId: deviceId)
+            case .failure(let error):
+                // If not dehydrated devices are available just continue and create a new one
```
pixlwave, line 80:
```suggestion
                // If no dehydrated devices are available just continue and create a new one
```

MatrixSDK/Crypto/Dehydration/DehydrationService.swift
```diff
@@ -0,0 +1,214 @@
...
+        await withCheckedContinuation { continuation in
+            self.restClient.retrieveDehydratedDevice { dehydratedDevice in
+                MXLog.info("Successfully retrieved dehydrated device with id: \(dehydratedDevice.deviceId)")
+                                
+                guard let deviceDataJSON = MXTools.serialiseJSONObject(dehydratedDevice.deviceData) else {
+                    continuation.resume(returning: .failure(DehydrationServiceError.invalidRehydratedDeviceData))
+                    return
+                }
+                
+                let rehydratedDevice = self.dehydratedDevices.rehydrate(pickleKey: [UInt8](pickleKeyData), deviceId: dehydratedDevice.deviceId, deviceData: deviceDataJSON)
```
pixlwave, line 163: Maybe should have a `[weak self]` in the closure?

pixlwave (approved):
LGTM from the code perspective (I'm obviously not familiar with the crypto stuff)

manuroe (commented):
Nice work!

We should probably remove the old tests `MXDehydrationTests.m` file. I wonder how it can still compile.
Do we want to have tests?

@poljar, are we ok to remove the previous device dehydration implementation? Are we sure it is not use somewhere?

poljar:
> @poljar, are we ok to remove the previous device dehydration implementation? Are we sure it is not use somewhere?

I'm afraid I have no idea.

MatrixSDK/Crypto/Dehydration/DehydrationService.swift
BillCarsonFr: Is this customer facing? If so we need to think more on how to name it.

BillCarsonFr (changes requested):
I might have missed it, but I don't see the sharing part of 4S for the new secret?
`m.secret.request`/`m.secret.send`?
If you login a new session then do an interactive self verification (emoji), you should request the dehydrated secret and then start the rehydratation

poljar:
> I might have missed it, but I don't see the sharing part of 4S for the new secret?
m.secret.request/m.secret.send?

That's not part of this, even if it were it wouldn't implemented here. The only thing the client would need to do is persist the pickle key in the crypto store so Rust has access to it.

But for this initial version of device dehydration, resumption of the rehydration process and periodic rotation of the dehydrated device didn't make the cut.

MatrixSDK/Background/Crypto/MXBackgroundCryptoV2.swift
```diff
@@ -53,7 +53,9 @@ class MXBackgroundCryptoV2: MXBackgroundCrypto {
                 toDevice: syncResponse.toDevice,
                 deviceLists: syncResponse.deviceLists,
                 deviceOneTimeKeysCounts: syncResponse.deviceOneTimeKeysCount ?? [:],
-                unusedFallbackKeys: syncResponse.unusedFallbackKeys
+                unusedFallbackKeys: syncResponse.unusedFallbackKeys,
+                nextBatchToken: syncResponse.nextBatch
+
```
pixlwave, line 58:
```suggestion
```

fastlane/Fastfile
```diff
@@ -18,7 +18,7 @@ platform :ios do
 
   before_all do
     # Ensure used Xcode version
-    xcversion(version: "~> 13.2")
+    xcversion(version: "~> 14.3")
```
pixlwave, line 21: Should this be 14.2 to keep the linting action working?

pixlwave (approved):

fastlane/Fastfile
stefanceriu: Worth a shot

Issue #1828, opened 2023-10-26
Remove the libbase58 pod dependency


ARiezn:
Currently, there is a problem with building Matrix-dependent SDKs using Xcode versions 14.3 and above. For some scenarios without increasing the deployment target for iOS in the libbase58 podspec this becomes a blocker - it causes a build error, not simply a warning:

```
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It may be necessary to update the libbase58 podspec to specify iOS 11 as the minimum deployment target. The original libbase58.podspec.json file should contain the value `9.0` or even better `11.0` (the minimal version for the latest Xcode versions) in the `platforms` `ios` section: [libbase58.podspec.json:19](https://github.com/CocoaPods/Specs/blob/master/Specs/3/4/3/libbase58/0.1.4/libbase58.podspec.json#L19). Without this change, the target is set to version 8.0 which causes the above described ld error. I've tried to contact @mohakshah who seem to be the maintainer of this pod to push the change (but without success, he does not seem to be active online during the last time).

For the top-level apps using libbase58 as a dependency, there is a workaround. It is possible to override the `build_settings['IPHONEOS_DEPLOYMENT_TARGET']` in the `post_install` actions in the `Podfile`. For example, this is recommended in the discussion https://github.com/matrix-org/matrix-ios-sdk/issues/804#issuecomment-656611365  in issue #804

Unfortunately, this above workaround will not work when building an SDK. `pod lib lint` will fail with an error. So this is a serious blocker that makes building completely impossible in this scenario.

There is another way of fixing the problem proposed in the PR - https://github.com/matrix-org/matrix-ios-sdk/pull/1828 . It completely removes the `libbase58` pod dependency and replaces it with native Base58 encoding-decoding implementation. This will also fix related issue #804.

ceburo:
Same problem here. We need the matrix-sdk in a library and we are completely blocked. 
All our Macs are on Sonoma and XCode15 (impossible to install XCode14.x).

This as to be fixed ASAP.

In state, this library is not usable. Development blocked = Production blocked.

pixlwave:
@ceburo You can get Xcode 14 to work on Sonoma if you run it from the command line:

```
[...]

PR #1829, opened 2023-10-26
Remove the libbase58 pod dependency

ARiezn:
### Summary
Completely remove `libbase58` dependency and replace it with ObjC implementation to manage base58 for keys backup.

### Details
See issue https://github.com/matrix-org/matrix-ios-sdk/issues/1828 for details

### Pull Request Checklist

* [X] Pull request is based on the develop branch
* [X] Pull request contains a changelog file in ./changelog.d. See https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog


ARiezn:
Signed-off-by: Artem Rieznikov <artem.rieznikov@teladochealth.com>

pixlwave:
Hey @ARiezn, thanks for this PR. Unfortunately we won't be able to accept it at this time. Until https://github.com/CocoaPods/CocoaPods/issues/11839 is fixed, we're developing this pod against Xcode 14.2.

Side note: as we're starting to transition to a Rust-backed world with https://github.com/matrix-org/matrix-rust-components-swift/ large changes to this library are unlikely to be accepted.

ARiezn:
@pixlwave , thank you for your attention to my PR. I want to highlight that the change proposed in this PR does not actually require to switch from using Xcode 14.2 for building. It will not break compatibility with Xcode 14.2. So continuation of using the same version of XCode should not be a problem at all. It will only make it possible to build using Xcode 14.3 or above (since the libbase58 dependency is removed). Is it possible to reconsider the decision to close the PR? Let me know if there are some other requirements or necessary changes that I may take into account.

pixlwave:
Hi, sorry I'm afraid not at this time @ARiezn.

PR #1832, opened 2023-11-20
Emit `InitialSyncDone` notification to allow client to perform auto-a…

NicolasBuquet:
Emit InitialSyncDone notification to allow client to perform auto-activation of cross-signing as in Android client

### Pull Request Checklist

* [X] Pull request is based on the develop branch
* [X] Pull request contains a changelog file in ./changelog.d. See https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog
* [X] Pull request includes a [sign off](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#sign-off)

### DIFF
--- a/MatrixSDK/MXSession.h
+++ b/MatrixSDK/MXSession.h
@@ -302,6 +302,15 @@ FOUNDATION_EXPORT NSString *const kMXSessionDidUpdateGroupUsersNotification;
  */
 FOUNDATION_EXPORT NSString *const kMXSessionDidUpdatePublicisedGroupsForUsersNotification;
 
+/**
+ Posted when MXSession has finished its initial sync.
+ No userInfo passed.
+ 
+`kMXSessionInitialSyncDone` notification will allow client to perform auto-activation
+of cross-signing as in Android client (see `observeInitialSync` calling `maybeVerifyOrBootstrapCrossSigning`)
+ */
+FOUNDATION_EXPORT NSString *const kMXSessionInitialSyncDone;
+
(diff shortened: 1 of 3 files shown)


NicolasBuquet:
@pixlwave Hello Doug.

Can you have a look at this PR. We need it on Tchap.

pixlwave:
Hi Nicolas, I'm a bit confused about the approach of this change. Tchap is a for of EI, so how come you can't use SessionVerificationListener which does the right thing?

NicolasBuquet:
Thank you for your attention @pixlwave.

I was not aware of SessionVerificationListener.

I will have a look at it and check if :  
- I can hook in it at the right moment and in the right state
- I can auto-activate cross-signing without user interaction (the famous grace period) from here

Thank you.

NicolasBuquet:
@pixlwave 

Update to the previous comment. 

AuthenticationCoordinator is used by OnboardingCoordinator in Element.

But OnboardingCoordinator is not used in the custom Onboarding process in Tchap.

So, the simple method is to mimic what is done on the Android side, listen to an `initialSyncDone` notification.

NicolasBuquet:
@pixlwave 

Update to the previous comment. 

AuthenticationCoordinator is used by OnboardingCoordinator in Element.

But OnboardingCoordinator is not used in the custom Onboarding process in Tchap.

So, the simple method is to mimic what is done on the Android side, listen to an `initialSyncDone` notification.

PR #1835, opened 2023-11-28
Update cocoapods and other gems.

pixlwave:
There's an issue with Cocoapods 0.13 and ActiveSupport 17.1 so this updates to 0.14 which has a fix for the issue.


nimau (approved):
LGTM 👍

PR #1854, opened 2024-05-24
Invalidate crypto store cache when entering foreground

BillCarsonFr:
### Pull Request Checklist

depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3462

Partial fix for the NSE process causing olm session wedging.
See element-x related issue https://github.com/matrix-org/matrix-rust-sdk/issues/3110.

The crypto store is keeping a cache of olm sessions for a given device. When using multiprocess this cache will get outdated and causing session wedging.
If the main app is inactive and the NSE do process a sync new olm sessions might be created (or mutated, like when replying to a key request). If the main process is then resumed it will use it's outdated cache instead of reloading from store.

The improvement here is that we will invalidate the cache when the main app enters foreground to limit the possible occurences of outdated caches.

### DIFF
--- a/MatrixSDK.podspec
+++ b/MatrixSDK.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
       ss.dependency 'OLMKit', '~> 3.2.5'
       ss.dependency 'Realm', '10.27.0'
       ss.dependency 'libbase58', '~> 0.1.4'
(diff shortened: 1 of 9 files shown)


stefanceriu (approved):
Nice, looks good to me. Hope the tests agree 🤞

PR #1858, opened 2024-06-05
Fix | Share room keys with dehydrated devices with rust stack

BillCarsonFr:
### Pull Request Checklist

Fixes https://github.com/element-hq/element-ios/issues/7795

The new [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814)https://github.com/matrix-org/matrix-spec-proposals/pull/3814 is introducing a new format for dehydrated devices (to get messages even if all your sessions are logged out). 
As per the new format, the DeviceKey payload from a keys/query call now contains a new boolean field `dehydrated`, that is used when computing the signature.
The problem is that the SDK converts the http response to a JSON Model `MXKeysQueryResponse` and this model was not aware of that field. As a consequence when passing back the response to rust via `markRequestAsSent` using `model.jsonString()`, the `dehydrated` field was dropped, thus the rust layer was refusing the device because the signature was invalid.

First approach would be to add the `dehydrated` field to the model, but this is not very future proof as any new future field would cause the same problem.
The rust crypto stack also just pass this payload and is not using it (the legacy stack neeeded the typed model).

Unfortunatly I cannot just modify the existing model to just use a generic Dictionnary because some parts of the sdk for the legacy crypto still needed it.
So I created a new `MXKeysQueryResponseRaw` almost similar to the existing `MXKeysQueryResponse`, but that is not trying to convert the device info to a typed model. This is now used by the rust layer to query and pass back to the olm machine.

### DIFF
--- a/MatrixSDK/JSONModels/MXJSONModels.h
+++ b/MatrixSDK/JSONModels/MXJSONModels.h
@@ -1125,6 +1125,25 @@ FOUNDATION_EXPORT NSString *const kMXPushRuleScopeStringGlobal;
 
 @end
 
+@interface MXKeysQueryResponseRaw : MXJSONModel
+
+    /**
+     The device keys per devices per users.
+     */
+    @property (nonatomic) NSDictionary<NSString *, id> *deviceKeys;
+
+    /**
+     Cross-signing keys per users.
+     */
+    @property (nonatomic) NSDictionary<NSString*, MXCrossSigningInfo*> *crossSigningKeys;
+
+    /**
+     The failures sorted by homeservers.
+    */
+    @property (nonatomic) NSDictionary *failures;
+
+@end
+
 /**
  `MXKeysClaimResponse` represents the response to /keys/claim request made by
  [MXRestClient claimOneTimeKeysForUsersDevices].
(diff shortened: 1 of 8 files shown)


MatrixSDK/JSONModels/MXJSONModels.h
```diff
@@ -1125,6 +1125,25 @@ FOUNDATION_EXPORT NSString *const kMXPushRuleScopeStringGlobal;
 
 @end
 
+@interface MXKeysQueryResponseRaw : MXJSONModel
+
+    /**
+     The device keys per devices per users.
+     */
+    @property (nonatomic) NSDictionary<NSString *, id> *deviceKeys;
```
BillCarsonFr, line 1133: In the existing MXKeysQueryResponse this was a `MXUsersDevicesMap<MXDeviceInfo*> *deviceKeys`

MatrixSDK/JSONModels/MXJSONModels.h
```diff
@@ -1125,6 +1125,25 @@ FOUNDATION_EXPORT NSString *const kMXPushRuleScopeStringGlobal;
...
+
+    /**
+     Cross-signing keys per users.
+     */
+    @property (nonatomic) NSDictionary<NSString*, MXCrossSigningInfo*> *crossSigningKeys;
```
uhoreg, line 1138: How easy would it be to do the same thing to the cross-signing keys, because we'll have the same problem here.  If it's easy, may as well do it while we're touching the code here.

uhoreg (approved):
I don't understand the code, but looks plausible.

PR #1871, opened 2024-07-05
Media Authentication support MSC 3916

Velin92:
This PR adds support for authenticated media APIs as of MSC 3916

### DIFF
--- a/MatrixSDK/MXRestClient.h
+++ b/MatrixSDK/MXRestClient.h
@@ -230,6 +230,12 @@ extern NSString *const kMXCredentialsNewRefreshTokenDataKey;
 @property (nonatomic) NSString *contentPathPrefix;
 
 /**
+ The Matrix content repository prefix to use for authenticated access.
+ By default, it is defined by the constant kMXContentPrefixPath.
+ */
+@property (nonatomic) NSString *authenticatedContentPathPrefix;
+
+/**
  The current trusted certificate (if any).
  */
 @property (nonatomic, readonly) NSData* allowedCertificate;
--- a/MatrixSDK/MXRestClient.h
+++ b/MatrixSDK/MXRestClient.h
@@ -251,6 +251,10 @@ extern NSString *const kMXCredentialsNewRefreshTokenDataKey;
  */
 @property (nonatomic, copy) NSSet <NSString *> *acceptableContentTypes;
 
+/**
+ Supported server versions of the matrix server, only for internal use of the SDK, use the stored version on the app side.
+ */
+@property (nonatomic, readonly, nullable) MXMatrixVersions *supportedVersions;
 
 /**
  Create an instance based on homeserver url.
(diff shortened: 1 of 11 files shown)


MatrixSDK/MXRestClient.h
```diff
@@ -229,6 +229,12 @@ extern NSString *const kMXCredentialsNewRefreshTokenDataKey;
  */
 @property (nonatomic) NSString *contentPathPrefix;
 
+/**
+ The Matrix content repository prefix to use for authenticated access.
+ By default, it is defined by the constant kMXContentPrefixPath.
```
pixlwave, line 234:
```suggestion
 By default, it is defined by the constant kMXAuthenticatedContentPrefixPath.
```

MatrixSDK/Utils/Media/MXMediaLoader.h
```diff
@@ -129,6 +129,11 @@ extern NSString *const kMXMediaUploadIdPrefix;
     NSTimer* progressCheckTimer;
 }
 
+/**
+ The access token.
+ */
+@property (nonatomic, readonly) NSString* accessToken;
+
```
pixlwave, line 136: I think we could make this one private if you wanted.

MatrixSDK/MXRestClient.h
```diff
@@ -245,6 +251,10 @@ extern NSString *const kMXCredentialsNewRefreshTokenDataKey;
  */
 @property (nonatomic, copy) NSSet <NSString *> *acceptableContentTypes;
 
+/**
+ Supported server versions of the matrix server, only for internal use of the SDK, use the stored version on the app side.
+ */
+@property (nonatomic, readonly, nullable) MXMatrixVersions *supportedVersions;
```
pixlwave, line 257: Ah, I just had an idea. Instead of making this public and "protecting" it with a comment. Why don't we make this `BOOL isUsingAuthenticatedMedia` defaulting to `NO`…

MatrixSDK/MXRestClient.m
```diff
@@ -529,6 +531,7 @@ - (MXHTTPOperation*)supportedMatrixVersions:(void (^)(MXMatrixVersions *matrixVe
                                          [self dispatchProcessing:^{
                                              MXJSONModelSetMXJSONModel(matrixVersions, MXMatrixVersions, JSONResponse);
                                          } andCompletion:^{
+                                             self.supportedVersions = matrixVersions;
```
pixlwave, line 534: … and then in here set it to `matrixVersions.supportsAuthenticatedMedia`…

MatrixSDK/Utils/Media/MXMediaManager.m
```diff
@@ -376,7 +376,11 @@ - (NSString*)urlOfContent:(NSString*)mxContentURI
         }
         else
         {
-            mxMediaPrefix = [NSString stringWithFormat:@"%@/%@/download/", _homeserverURL, kMXContentPrefixPath];
+            mxMediaPrefix = [NSString stringWithFormat:@"%@/%@/download/", _restClient.homeserver, kMXContentPrefixPath];
+            if (_restClient.supportedVersions && [_restClient.supportedVersions supportsAuthenticatedMedia])
```
pixlwave, line 380: … and then it's only a single check here as it won't be `nil`.

pixlwave (approved):
LGTM, couple of small thoughts I've just had added inline.

Also, we still need a changelog file in the old App/SDK 🫠

PR #1879, opened 2024-09-12
Replaced OlmKit functions for content scanner with CryptoSDK

Velin92:


MatrixSDK/ContentScan/PKEncryptionHelper.swift
```diff
@@ -0,0 +1,50 @@
...
+import Foundation
+import MatrixSDKCrypto
+
+@objc public class PKEncryptionHelper: NSObject {
+    @objc public static func getPKMessage(_ message: String, using publicKey: String) -> PKMessageWrapper? {
+        do {
+            let pkEncryption = try PkEncryption.fromBase64(key: publicKey)
+            return PKMessageWrapper(pkMessage: pkEncryption.encrypt(plaintext: message))
+        } catch {
+            MXLog.error("[PKEncryptionHelper] failed to create pkEncryption", context: error)
```
pixlwave, line 26: Is this error safe to log/submit to Sentry?

pixlwave (approved):
LGTM other than the `errorOnverifiedUserProblem`, `clearCryptoCache` and `.created` changes which I have no context around what they should be doing

MatrixSDK/ContentScan/PKEncryptionHelper.swift
Velin92: @poljar not sure, let's ask Damir

MatrixSDK/Crypto/CryptoMachine/MXCryptoMachine.swift
```diff
@@ -117,7 +117,8 @@ class MXCryptoMachine {
     }
     
     func invalidateCache() async {
-        await machine.clearCryptoCache()
+        // Not sure what is the new function for this
+//        await machine.clearCryptoCache()
```
Velin92, line 121: this function seems to not exist anymore in this crypto sdk version, what is the replacement?
@poljar

poljar: The cache doesn't exist anymore, at least not inside the store. Rotating the Olm machine would clear all caches now.

MatrixSDK/ContentScan/PKEncryptionHelper.swift
poljar: Yes, it just tells you that the public key is invalid and why.

poljar: The encrypt call could be outside of the try, since it can't fail once you have a correctly decoded public key.

MatrixSDK/Crypto/Verification/Transactions/SAS/MXSASTransactionV2.swift
```diff
@@ -172,6 +172,8 @@ extension MXSASTransactionV2: SasListener {
                 humanReadable: cancelInfo.reason
             )
             self.state = cancelInfo.cancelledByUs == true ? MXSASTransactionStateCancelledByMe : MXSASTransactionStateCancelled
+        case .created:
```
poljar, line 175: This state has been split out from the `.started` state, we now have different states for when we start the verification process vs when the other side starts it.

It seems that you previously used the `isIncoming` boolean to decide which state you're going into. I think that you need to move `self.state = MXSASTransactionStateOutgoingWaitForPartnerToAccept` over here and remove the branching depending on the `isIncoming` boolean.

MatrixSDK/Crypto/Algorithms/RoomEvent/MXRoomEventEncryption.swift
```diff
@@ -184,7 +184,8 @@ struct MXRoomEventEncryption: MXRoomEventEncrypting {
             rotationPeriodMsgs: UInt64(Self.keyRotationPeriodMsgs),
             // If not set, history visibility defaults to `joined` as the most restrictive setting
             historyVisibility: state.historyVisibility?.visibility ?? .joined,
-            onlyAllowTrustedDevices: onlyTrustedDevices(in: roomId)
+            onlyAllowTrustedDevices: onlyTrustedDevices(in: roomId),
+            errorOnVerifiedUserProblem: false
```
poljar, line 188: Yeah, this is for invisible crypto and the old apps AFAIK won't convert to it, so defaulting to `fasle` is correct here.

MatrixSDK/Crypto/Verification/Transactions/SAS/MXSASTransactionV2.swift
poljar: Yes.

PR #1897, opened 2025-01-14
Remove trailing slash on identity server

NicolasBuquet:
Old Account data can have a trailing slash at the end of their Identity Server.
This can lead to unrecognized URL on the backend (like on 'invite to room') because the URL is then constructed with a double slash in its path.
This lead to error 500 for these calls.
So, fix this trailing slash as soon as we receive it.

### Pull Request Checklist


NicolasBuquet:
Bad files commited

PR #1902, opened 2025-03-11
Remove lib linting CI

Velin92:
Since we don't release anymore the library on cocoapods, and there is [an issue with Cocoapods](https://github.com/CocoaPods/CocoaPods/issues/11839) that only allows linting to work with up to Xcode 14.2 with the libbase58 library, even if the project can be built with newer versions... then we decided to just get rid of this workflow entirely.

We could run the linting and just the linting with an older version, but this feels like a waste of resources given the fact that as mentioned above, we don't publish on cocapods anymore


pixlwave (approved):
🔥🔥🔥

manuroe (approved):
It is fine by me.
Thanks for the cleanup.

Issue #1810, opened 2023-08-23
Crash when saving read receipts


stefanceriu:
```
Thread 22 name:
Thread 22 Crashed:
0   libobjc.A.dylib               	0x000000018a202e5c objc_retain_x0 + 16
1   Foundation                    	0x000000018b2d72d4 -[NSDictionary(NSDictionary) encodeWithCoder:] + 488 (NSDictionary_Foundation.m:46)
2   Foundation                    	0x000000018b2d3b44 _encodeObject + 1188 (NSKeyedArchiver.m:1234)
3   Foundation                    	0x000000018b31a1ec -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 468 (NSKeyedArchiver.m:1279)
4   Foundation                    	0x000000018b2d732c -[NSDictionary(NSDictionary) encodeWithCoder:] + 576 (NSDictionary_Foundation.m:47)
5   Foundation                    	0x000000018b2d3b44 _encodeObject + 1188 (NSKeyedArchiver.m:1234)
6   Foundation                    	0x000000018b837474 +[NSKeyedArchiver archiveRootObject:toFile:] + 216 (NSKeyedArchiver.m:750)
7   MatrixSDK                     	0x0000000102bf9600 __27-[MXFileStore saveReceipts]_block_invoke + 460 (MXFileStore.m:2212)
8   libdispatch.dylib             	0x00000001984ac4b4 _dispatch_call_block_and_release + 32 (init.c:1518)
9   libdispatch.dylib             	0x00000001984adfdc _dispatch_client_callout + 20 (object.m:560)
10  libdispatch.dylib             	0x00000001984b5694 _dispatch_lane_serial_drain + 672 (inline_internal.h:2632)
11  libdispatch.dylib             	0x00000001984b61e0 _dispatch_lane_invoke + 384 (queue.c:3940)
12  libdispatch.dylib             	0x00000001984c0e10 _dispatch_workloop_worker_thread + 652 (queue.c:6846)
13  libsystem_pthread.dylib       	0x00000001ddf33df8 _pthread_wqthread + 288 (pthread.c:2618)
14  libsystem_pthread.dylib       	0x00000001ddf33b98 start_wqthread + 8
```

lucasfeijo:
This still happens to me on the latest release

```
          Crashed: MXFileStoreDispatchQueue
0  libobjc.A.dylib                0x7930 objc_retain + 16
[...]

PR #1904, opened 2025-05-28
Fixing Race Condition on multiple commits to saveReceipts

brandonmaul:
This fix solves a relatively rare race condition that happens when multiple commits are made to MXFileStore in quick succession to readReceipts. It manifests in an error such as in #1810

To solve this, we copy the entire `RoomThreadedReceiptsStore` for a room, and it's sub-values. We then use that "copy" to actually write to the store, and passing it along to the `NSKeyedArchiver`.

If there is a better approach to solving this please let us know, we're more than happy to evaluate it and we understand that this may not be the most efficient way of solving this issue.

Signed-off-by: Brandon Maul <brandon.maul@powerhrg.com>

### DIFF
--- a/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
+++ b/MatrixSDK/Data/Store/MXFileStore/MXFileStore.m
@@ -2192,41 +2192,48 @@ static NSUInteger preloadOptions;
             // Save rooms where there was changes
             for (NSString *roomId in roomsToCommit)
             {
-                RoomThreadedReceiptsStore *receiptsStore =  self->roomThreadedReceiptsStores[roomId];
-                if (receiptsStore)
+                RoomThreadedReceiptsStore *store = self->roomThreadedReceiptsStores[roomId];
+                if (!store) continue;
+                
+                NSDictionary<NSString*, RoomReceiptsStore*> *snapshot;
+                @synchronized (store) {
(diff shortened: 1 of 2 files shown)


brandonmaul:
@pixlwave @stefanceriu @richvdh Asking for a review

brandonmaul:
For more context, here is roughly how you can reproduce the bug (that this change fixes)

```
let mxSession: MXSession = ...
let creds: MXCredentials = ...
guard mxSession.store is MXFileStore else {
  print("⚠️ Session's store is not MXFileStore.")
  return
}

var receipts: [MXReceiptData] = []
for i in 0...1000 {
  let receipt = MXReceiptData()
  receipt.userId      = creds.userId! + String(describing: i)
  receipt.threadId    = kMXEventTimelineMain
  receipt.eventId     = "$crashRepro\(UUID().uuidString)"
  receipt.ts          = UInt64(Date().timeIntervalSince1970 * 1000)
  
  receipts.append(receipt)
}

for i in 0...1000 {
  let room = "!reproRoom:\(creds.homeServer!)" + String(describing: i)
  for j in 0...1000 {
    let receipt = receipts[j]
    mxSession.store.storeReceipt(receipt, inRoom: room)
    mxSession.store.commit?()
  }
}
```

AlexandreHauber:
Following 👀!

PR #1908, opened 2025-07-22
Support for MSC 4289

Velin92:
Room creator will now have the highest possible power level, which won't be determined by `m.room.power_levels`.
Also a room may have additional creators.

Also the `powerLevelOfUserWithUserID` function from MXPowerLevels has been removed in favour of the newer version in `MXRoomState`.

### DIFF
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -634,7 +634,7 @@
 {
     if ([userId isEqualToString: [self creatorUserId]] || [[self additionalCreators] containsObject: userId])
     {
-        return NSIntegerMax;
+        return maxPowerLevel ? maxPowerLevel + 1 : NSIntegerMax;
     }
     return [powerLevels powerLevelOfUserWithUserID:userId];
 }
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -656,7 +656,18 @@
             return maxPowerLevel ? maxPowerLevel + 1 : NSIntegerMax;
         }
     }
-    return [powerLevels powerLevelOfUserWithUserID:userId];
+    
+    // By default, use usersDefault
+    NSInteger userPowerLevel = powerLevels.usersDefault;
+
+    NSNumber *powerLevel;
+    MXJSONModelSetNumber(powerLevel, powerLevels.users[userId]);
+    if (powerLevel)
+    {
+        userPowerLevel = [powerLevel integerValue];
+    }
+
+    return userPowerLevel;
 }
 
 - (BOOL)isMSC4289Supported {
--- a/MatrixSDK/Data/MXRoomState.m
+++ b/MatrixSDK/Data/MXRoomState.m
@@ -328,6 +328,23 @@
     return avatar;
 }
 
+- (NSString *)roomVersion
+{
+    NSString *roomVersion;
+    
+    // Check it from the state events
+    MXEvent *event = [stateEvents objectForKey:kMXEventTypeStringRoomCreate].lastObject;
+    NSDictionary<NSString *, id> *eventContent = [self contentOfEvent:event];
+    
+    if (event && eventContent)
+    {
+        MXJSONModelSetString(roomVersion, eventContent[@"room_version"]);
+        roomVersion = [roomVersion copy];
+
+    }
+    return roomVersion;
+}
+
 - (NSString *)creatorUserId
 {
     NSString *creatorUserId;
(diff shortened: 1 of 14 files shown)


MatrixSDK/Data/MXRoomState.m
```diff
@@ -601,13 +633,52 @@ - (float)memberNormalizedPowerLevel:(NSString*)userId
...
+}
+
+- (BOOL)isMSC4289Supported {
+    NSArray<NSString*> *supportedRoomVersions = @[@"org.matrix.hydra.11",@"12"];
+    if ([self roomVersion])
+    {
+        return [supportedRoomVersions containsObject:[self roomVersion]];
+    }
+    return NO;
+}
```
Velin92, line 680: Not sure if I should centralise or move this check into another place of is fine where it is.

Also not sure if checking an array of values is the best way, but I was told that room versions should not be treated as integers, and should not be compared using arithmetic operators, but supported versions should be hardcoded.

However considering the app is also sunsetting, maybe in this case we may make an exception and use and check also for the room version to be > 11 ?

pixlwave: > Not sure if I should centralise or move this check into another place of is fine where it is.

Maybe we could make `self.roomVersion` an `MXRoomVersion` and put the method on it? At least that way whenever there is a new room version it would be far more likely that it was found and updated?

> Also not sure if checking an array of values is the best way, but I was told that room versions should not be treated as integers, and should not be compared using arithmetic operators, but supported versions should be hardcoded.
> 
> However considering the app is also sunsetting, maybe in this case we may make an exception and use and check also for the room version to be > 11 ?

It's almost tempting to invert the check and say `![1, 2, 3, … 10, 11].contains(version)` but those unstable identifiers basically break this logic either way as `com.company.old-feat.2` and `com.company.feat.15` both need to do the right thing here 😕. Really don't know what the answer is here, other than "what does Rust/Web do?"

MatrixSDK/Data/MXRoomState.m
```diff
@@ -601,13 +633,52 @@ - (float)memberNormalizedPowerLevel:(NSString*)userId
...
+    MXJSONModelSetNumber(powerLevel, powerLevels.users[userId]);
+    if (powerLevel)
+    {
+        userPowerLevel = [powerLevel integerValue];
+    }
+
+    return userPowerLevel;
+}
+
+- (BOOL)isMSC4289Supported {
```
pixlwave, line 673: I'm wondering if this needs to be public (wherever it ends up) so that you can check if a user has PL150 for the new room version before upgrading a room.

pixlwave (approved):
LGTM (at least from what I understand about the MSC, I haven't read it fully yet).

Although note that the unit tests are failing to compile.

MatrixSDK/Data/MXRoomState.m
Velin92: for now web is going to do the same as us, not sure about Rust

MatrixSDK/Data/MXRoomState.m
Velin92: there are other PRs to be added, if it will be required I'll make it public

MatrixSDK/Data/MXRoomState.m
```diff
@@ -334,15 +334,30 @@ - (NSString *)creatorUserId
     
     // Check it from the state events
     MXEvent *event = [stateEvents objectForKey:kMXEventTypeStringRoomCreate].lastObject;
+    NSString* sender = [event sender];
     
+    if (event && sender)
+    {
+        creatorUserId = [sender copy];
```
stefanceriu, line 358: This used to fetch the creatorID from 
```
MXJSONModelSetString(creatorUserId, eventContent[@"creator"]);
creatorUserId = [creatorUserId copy];
```

are you sure that's not needed anymore?

MatrixSDK/Data/MXRoomState.m
```diff
@@ -601,13 +616,29 @@ - (float)memberNormalizedPowerLevel:(NSString*)userId
     // Ignore banned and left (kicked) members
     if (member.membership != MXMembershipLeave && member.membership != MXMembershipBan)
     {
-        float userPowerLevelFloat = [powerLevels powerLevelOfUserWithUserID:userId];
-        powerLevel = maxPowerLevel ? userPowerLevelFloat / maxPowerLevel : 1;
```
stefanceriu, line 605: probably could've used `MIN(userPowerLevelFloat / maxPowerLevel, 1)` here

stefanceriu (approved):
Not particularly familiar with the code but this looks sensible to me 👍

MatrixSDK/Data/MXRoomState.m
Velin92: `creator` has been removed since roomVersion 11 and the sender should be used in its place
https://spec.matrix.org/v1.11/rooms/v11/#remove-the-creator-property-of-mroomcreate-events

This is backwards compatible with all room versions.

MatrixSDK/Data/MXRoomState.m
Velin92: `maxPowerLevel` can not be `nil` in objc, it will be 0, so if `maxPowerLevel` is 0 we will divide by 0... `if (maxPoweLevel)` will still be `false` when `maxPowerLevel` is 0

stefanceriu: Ah, I see 👍

MatrixSDK/Data/MXRoomState.m
stefanceriu: Cool cool 👍

PR #1909, opened 2025-07-23
`tombStoneEventId` can now be optional

Velin92:
Even if not really used in the app, the initialisation of `MXPredecessorInfo` required `tombStoneEventId` to be `nonnull`, since room version 12 now such value can be optional. This PR just makes model able to be initialised if the value is null.


stefanceriu (approved):
🤩

PR #1917, opened 2025-12-15
Support for stable version of MSC3824 OAuth 2.0 aware

hughns:
Partial fix for https://github.com/element-hq/element-ios/issues/7925.

- [x] Support for stable OAuth 2.0 API discovery with fallback to unstable MSC2965 .well-known
  - Uses the  `/_matrix/client/v1/auth_metadata` endpoint if available.
- [x] Where server supports auth_metadata, then support for stable action parameter from MSC4191
  - Uses `account_management_actions_supported` discovery to choose the stable or unstable MSC4191 `action` parameters passed to `account_management_uri`.
- [x] Support for stable `oauth_aware_preferred` flag on `m.login.sso` auth flow from MSC3824
- Breaking change: `WKWellKnown.authentication` is now internal. `MXSession.hasOAuth2APIEnabled`, `MXSession.accountManagementURI` and `MXSession.getLogoutDeviceURLFromID()` should be used instead.

### Pull Request Checklist

Signed-off-by: Hugh Nimmo-Smith <hughns@element.io>

### DIFF
--- a/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.h
+++ b/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.h
@@ -32,7 +32,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface MXAuthMetadata : MXJSONModel<NSCoding>
 
 @property (nonatomic, readonly) NSString *issuer;
-@property (nonatomic, readonly, nullable) NSString *accountManagementUri;
+@property (nonatomic, readonly, nullable) NSString *accountManagementURI;
 @property (nonatomic, readonly, nullable) NSArray<NSString*> *accountManagementActionsSupported;
 
 -(NSURL * _Nullable) getLogoutDeviceURLFromID: (NSString * ) deviceID;
--- a/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.h
+++ b/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.h
@@ -1,0 +1,42 @@
+//
+// Copyright 2025 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "MXJSONModel.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Subset of OAuth 2.0 API Server metadata: https://spec.matrix.org/v1.16/client-server-api/#server-metadata-discovery
+ {
+    "issuer": "https://example.com/",
+    "account_management_uri": "https://example.com/account",
+    "account_management_actions_supported": ["org.matrix.profile", "org.matrix.device_delete"]
+ }
+ */
+
+@interface MXAuthMetadata : MXJSONModel<NSCoding>
+
+@property (nonatomic, readonly) NSString *issuer;
+@property (nonatomic, readonly, nullable) NSString *accountManagementUri;
+@property (nonatomic, readonly, nullable) NSArray<NSString*> *accountManagementActionsSupported;
+
+-(NSURL * _Nullable) getLogoutDeviceURLFromID: (NSString * ) deviceID;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.m
+++ b/MatrixSDK/JSONModels/Authentication/MXAuthMetadata.m
@@ -16,14 +16,14 @@
 
 #import "MXAuthMetadata.h"
 
-static NSString *const kMXIssuer = @"issuer";
-static NSString *const kMXAccountManagementUri = @"account_management_uri";
-static NSString *const kMXAccountManagementActionsSupported = @"account_management_actions_supported";
+static NSString *const kMXIssuerJSONKey = @"issuer";
+static NSString *const kMXAccountManagementUriJSONKey = @"account_management_uri";
+static NSString *const kMXAccountManagementActionsSupportedJSONKey = @"account_management_actions_supported";
 
 @interface MXAuthMetadata ()
 
 @property (nonatomic, readwrite) NSString *issuer;
-@property (nonatomic, readwrite, nullable) NSString *accountManagementUri;
+@property (nonatomic, readwrite, nullable) NSString *accountManagementURI;
 @property (nonatomic, readwrite, nullable) NSArray<NSString*> *accountManagementActionsSupported;
 
 @end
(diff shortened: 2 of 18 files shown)


MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
pixlwave, line 1: Ohhhh the key changed. I assumed this was all replaced by the CS API without even looking into it 🙈

MatrixSDK/MXSession.h
```diff
@@ -1675,4 +1675,38 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
...
+/**
+ Refresh self.authMetadata.
+
+ @param success A block object called when the operation succeeds.
+ @param failure A block object called when the operation fails.
+
+ @return a MXHTTPOperation instance.
+ */
+- (MXHTTPOperation*)refreshAuthMetadata:(void (^)(MXAuthMetadata *authMetadata))success
+                                       failure:(void (^)(NSError *error))failure;
```
pixlwave: Same in the `.m` file too.

```suggestion
- (MXHTTPOperation*)refreshAuthMetadata:(void (^)(MXAuthMetadata *authMetadata))success
                                failure:(void (^)(NSError *error))failure;
```

MatrixSDK/JSONModels/Authentication/MXAuthMetadata.h
```diff
@@ -0,0 +1,42 @@
...
+    "issuer": "https://example.com/",
+    "account_management_uri": "https://example.com/account",
+    "account_management_actions_supported": ["org.matrix.profile", "org.matrix.device_delete"]
+ }
+ */
+
+@interface MXAuthMetadata : MXJSONModel<NSCoding>
+
+@property (nonatomic, readonly) NSString *issuer;
+@property (nonatomic, readonly, nullable) NSString *accountManagementUri;
```
pixlwave: Major nitpick here:

```suggestion
@property (nonatomic, readonly, nullable) NSString *accountManagementURI;
```

MatrixSDK/JSONModels/Authentication/MXAuthMetadata.m
```diff
@@ -0,0 +1,104 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import "MXAuthMetadata.h"
+
+static NSString *const kMXIssuer = @"issuer";
+static NSString *const kMXAccountManagementUri = @"account_management_uri";
+static NSString *const kMXAccountManagementActionsSupported = @"account_management_actions_supported";
```
pixlwave: Not super important seeing as they're private, but generally these would be slightly more descriptive, something along the lines of

`kMXAuthMetadataIssuer`/`kMXAuthMetadataTypeIssuer`/`kMXIssuerJSONKey`

MatrixSDK/MXSession.h
```diff
@@ -1675,4 +1675,38 @@ typedef void (^MXOnBackgroundSyncFail)(NSError *error);
...
+/**
+ True if the homeserver has the OAuth 2.0 API enabled.
+ */
+@property (nonatomic, readonly) bool hasOAuth2APIEnabled;
+
+
+/**
+ The homeserver OAuth 2.0 account management URI, if any.
+ */
+@property (nonatomic, readonly, nullable) NSString *accountManagementUri;
```
pixlwave: Same here:

```suggestion
@property (nonatomic, readonly, nullable) NSString *accountManagementURI;
```

MatrixSDK/MXSession.m
```diff
@@ -5196,4 +5196,68 @@ - (NSString *)preferredSyncPresenceString
...
+    return nil;
+}
+
+
+- (MXAuthMetadata *)authMetadata
+{
+    return self.store.authMetadata;
+}
+
+- (MXHTTPOperation *)refreshAuthMetadata:(void (^)(MXAuthMetadata *))success
```
pixlwave, line 5242: I might be missing it, but I don't see this being called either in this PR, or in https://github.com/element-hq/element-ios/pull/7991.

I would assume we should drop it in here:

https://github.com/matrix-org/matrix-ios-sdk/blob/b0f967c9c884e1a24d037f516f40a5ffb6c29586/MatrixSDK/MXSession.m#L1046

pixlwave (commented):
Very nice, one comment about refreshing the metadata, otherwise just minor comments.

pixlwave:
Also 2 more thoughts
- Given this is backwards compatible, I think we should probably make `homeserverWellknown.authentication` internal now, just so that SDK users have to switch to the new API.
- This one also needs a changelog.

MatrixSDK/MXSession.m
hughns: Yes, that is what I meant to do.

hughns: Now added.

MatrixSDK/JSONModels/Login/MXLoginSSOFlow.m
hughns: Unfortunately not, you still need to check the flag on the flow.

MatrixSDK/JSONModels/AutoDiscovery/MXWellKnown.m
```diff
@@ -87,7 +88,6 @@ - (id)initWithCoder:(NSCoder *)aDecoder
         _identityServer = [aDecoder decodeObjectForKey:kMXIdentityServerKey];
         _integrations = [aDecoder decodeObjectForKey:kMXIntegrationsKey];
         _tileServer = [aDecoder decodeObjectForKey:kMXTileServerKey];
-        _authentication = [aDecoder decodeObjectForKey:kMXAuthenticationKey];
```
pixlwave, line 90: I think we need to keep this in, otherwise `hasOAuth2APIEnabled` etc may return the wrong value for servers without the CS-api when loading the well-known from the store without subsequently updating from the network.

MatrixSDK/JSONModels/AutoDiscovery/MXWellKnown.m
```diff
@@ -99,7 +99,6 @@ - (void)encodeWithCoder:(NSCoder *)aCoder
     [aCoder encodeObject:_identityServer forKey:kMXIdentityServerKey];
     [aCoder encodeObject:_integrations forKey:kMXIntegrationsKey];
     [aCoder encodeObject:_tileServer forKey:kMXTileServerKey];
-    [aCoder encodeObject:_authentication forKey:kMXAuthenticationKey];
```
pixlwave, line 102: And same here for storing it.

pixlwave (commented):
All looks good to me, just one thing I noticed regarding the cached well-known.

I haven't opened the project in Xcode to find out if you had to remove them as a result of using `MXWellKnown_Private.h`, but from my (very rusty) objc memory, I can't immediately see why the 2 lines below couldn't remain in the place.

MatrixSDK/JSONModels/AutoDiscovery/MXWellKnown.m
pixlwave: Ah, I see now, this one needs to use `self.authentication` as the ivar isn't synthesised for the private category. I've pushed a fix.

pixlwave: Ok well in practice the app crashes in `modelFromJSON` anyway… I've added the property definition in the `.m` file's class extension too (meaning `_authentication` also exists again) and now all seems good.

pixlwave (approved):

PR #1924, opened 2026-03-23
Update JitsiMeetSDKLite from v8 to v12.

pixlwave:
I've tested it with Web and everything works fine. The `LowComplexity` constants and `RTCEAGLVideoView` have been removed hence the updates, otherwise nothing jumped out at me in the release notes other than the fact it appears to now include a [Hermes](https://github.com/facebook/hermes/blob/main/README.md) XCFramework 🙄.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,4 +4,4 @@ gem "fastlane"
 gem "cocoapods", '~>1.16.2'
 gem "xcode-install"
 gem "slather"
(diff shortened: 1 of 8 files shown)


stefanceriu (approved):
Works for me, 🚀 it!

pixlwave:
Same old integration tests failure.

PR #1928, opened 2026-05-26
Update the Crypto SDK.

pixlwave:
Brings in the latest updates from 0.17.0.

I've also switched to using cocoapods specs directly (like Element Classic does) as the CDN is ridiculously broken and couldn't find the updated version 🤷‍♂️


pixlwave:
Integration test failures are the same ones we see on all the PRs.

stefanceriu (approved):
🚀

